Issue #13, opened 2021-11-10
Nautobot Example fails due to Pagination


chadell:
### Environment
* Python version:  3.9
* Nautobot version:  1.1.4
* nautobot-ssot version:  1.01

Triggering the diffsycn from Remote Nautobot (from demo.nautobot.com) to local Nautobot should execute successfully.

### Observed Behavior

Because pagination is disabled to load objects from Remote Nautobot (https://github.com/nautobot/nautobot-plugin-ssot/blob/develop/nautobot_ssot/jobs/examples.py#L293) it loads only 50 Regions and 50 Sites, so some of the Sites are referencing to Regions that do not exist, failing the execution.

### Steps to Reproduce
1. Plugins -> Dashboard
2. Data Sources -> Example Data Source
3.  Sync Now without Dry Run

chadell:
if it could help, I've already solved it in an exercise I'm working on, using simple pagination instead of `pynautobot` (I wanted to skip adding extra libraries). 
If it works for you I could update the code here too.

chadell:
I'm doing some work on performance testing and I will  use the example as reference, so I will fix it (in case you want to assign it to me)

chadell:
Fixed in #14

PR #14, opened 2021-11-22
Extend Nautobot example with Prefix model fix pagination issue 

chadell:
+ Fix issue #13 adding pagination
+ Adds a new `diffsyncmodel` for `Prefix`
+ Adds support for Celery


nautobot_ssot/jobs/examples.py
```diff
@@ -441,6 +557,13 @@ def lookup_object(self, model_name, unique_id):
                 return Site.objects.get(name=unique_id)
             except Site.DoesNotExist:
                 pass
+        elif model_name == "prefix":
+            try:
+                return Prefix.objects.get(
+                    prefix=unique_id.split("__")[0], tenant__slug=unique_id.split("__")[1] or None
```
chadell, line 563: @glennmatthews , is there a better way to get multiple keys from the `unique_id`?

glennmatthews: Not really -- ultimately it's up to your job what the `unique_id` contains, after all -- but it wouldn't be a bad idea to add a small helper function to wrap the `unique_id.split("__")` call.

development/nautobot_config.py
```diff
@@ -363,3 +363,6 @@ def is_truthy(arg):
     EXTRA_INSTALLED_APPS.append("debug_toolbar")
 if "debug_toolbar.middleware.DebugToolbarMiddleware" not in settings.MIDDLEWARE:
     settings.MIDDLEWARE.insert(0, "debug_toolbar.middleware.DebugToolbarMiddleware")
+
+# Increasing limit of Celery run time, because some dump examples are intense!
+NAUTOBOT_CELERY_TASK_SOFT_TIME_LIMIT = 600
```
glennmatthews, line 368: 1) `NAUTOBOT_CELERY_TASK_SOFT_TIME_LIMIT` is an environment variable - the settings variable you need here would be simply `CELERY_TASK_SOFT_TIME_LIMIT`
2) Also need to set the hard limit `CELERY_TASK_TIME_LIMIT` correspondingly.
3) We should document the need for this in the README so that any end users of this plugin don't get tripped up by it.

nautobot_ssot/jobs/examples.py
```diff
@@ -252,6 +276,63 @@ def delete(self):
...
+            defaults={"name": ids["tenant_slug"]},
+        )
+        prefix.tenant = tenant_obj
+        if "status_slug" in attrs:
+            status_obj, _ = Status.objects.get_or_create(
+                slug=attrs["status_slug"],
+                defaults={"name": attrs["status_slug"]},
+            )
+            status_obj.content_types.set([ContentType.objects.get_for_model(Prefix)])
+            status_obj.save()
```
glennmatthews, line 343:
```suggestion
            status_obj.content_types.add(ContentType.objects.get_for_model(Prefix))
```

nautobot_ssot/jobs/examples.py
```diff
@@ -252,6 +276,63 @@ def delete(self):
...
+            ids (dict): Initial values for this model's _identifiers
+            attrs (dict): Initial values for this model's _attributes
+        """
+        prefix = Prefix(prefix=ids["prefix"], description=attrs["description"])
+        tenant_obj, _ = Tenant.objects.get_or_create(
+            slug=ids["tenant_slug"],
+            defaults={"name": ids["tenant_slug"]},
+        )
+        prefix.tenant = tenant_obj
+        if "status_slug" in attrs:
```
glennmatthews, line 337: This check shouldn't be necessary in `create()` - unless I'm mistaken `status_slug` will always be provided.

nautobot_ssot/jobs/examples.py
```diff
@@ -252,6 +276,63 @@ def delete(self):
...
+        for attr_name in ("description",):
+            if attr_name in attrs:
+                setattr(prefix, attr_name, attrs[attr_name])
+
+        if "status_slug" in attrs:
+            status_obj, _ = Status.objects.get_or_create(
+                defaults={"name": attrs["status_slug"]},
+            )
+            status_obj.content_types.set([ContentType.objects.get_for_model(Prefix)])
+            status_obj.save()
```
glennmatthews, line 364:
```suggestion
            status_obj.content_types.add(ContentType.objects.get_for_model(Prefix))
```

nautobot_ssot/jobs/examples.py
```diff
@@ -252,6 +276,63 @@ def delete(self):
...
+            prefix.status = status_obj
+        prefix.validated_save()
+        return super().update(attrs)
+
+    def delete(self):
+        """Delete an existing Prefix record from local Nautobot."""
+        prefix = Prefix.objects.get(prefix=self.prefix)
```
glennmatthews, line 371: Shouldn't this also use `tenant_slug=self.tenant_slug`?

nautobot_ssot/jobs/examples.py
```diff
@@ -318,6 +405,22 @@ def load(self):
             self.add(site)
             self.job.log_debug(message=f"Loaded {site} from remote Nautobot instance")
 
+        prefix_data = requests.get(f"{self.url}/api/ipam/prefixes/", headers=self.headers, params={"limit": 0}).json()
+        prefixes = prefix_data["results"]
+        while prefix_data["next"]:
+            prefix_data = requests.get(prefix_data["next"], headers=self.headers, params={"limit": 0}).json()
+            prefixes.extend(prefix_data["results"])
```
glennmatthews, line 452: Maybe write a helper function for this 3x repeated pattern?

nautobot_ssot/jobs/examples.py
```diff
@@ -318,6 +405,22 @@ def load(self):
...
+        for prefix_entry in prefixes:
+            prefix = self.prefix(
+                prefix=prefix_entry["prefix"],
+                description=prefix_entry["description"],
+                status_slug=prefix_entry["status"]["value"] if prefix_entry["status"] else "active",
```
glennmatthews, line 457: Is there ever a case where `prefix_entry["status"]` isn't present? That would seem like a bug to me.

nautobot_ssot/jobs/examples.py
```diff
@@ -377,6 +481,17 @@ def load(self):
             self.add(site_model)
             self.job.log_debug(message=f"Loaded {site_model} from local Nautobot instance")
 
+        for prefix in Prefix.objects.all():
+            prefix_model = self.prefix(
+                prefix=str(prefix.prefix),
+                description=prefix.description,
+                status_slug=prefix.status.slug if prefix.status else "active",
```
glennmatthews, line 528: Same here - I wouldn't think that `prefix.status` should ever be `None`.

development/nautobot_config.py
chadell: sorry, should not be committed

nautobot_ssot/jobs/examples.py
chadell: don't think so, I guess it's present always because it's a mandatory field, but I followed the same pattern we had before for site and region. Should I update the others too?

glennmatthews: Hm, yeah, I guess status shouldn't ever be None for Sites either. Region `parent` and Site `region` can be None, so it makes sense to keep the current logic for them.

chadell:
@glennmatthews I took the time to migrate the development environment to use Celery in 843cb84. 
If you don't want to keep it dismiss this commit

development/docker-compose.base.yml
```diff
@@ -25,8 +25,27 @@ services:
...
       - "nautobot"
+    <<: *nautobot-base
+  celery_worker:
     healthcheck:
-      disable: true
+      interval: "5s"
+      timeout: "5s"
+      start_period: "5s"
+      retries: 3
+      test: ["CMD", "nautobot-server", "health_check"]
```
glennmatthews, line 43: see https://github.com/nautobot/nautobot/issues/1106 and https://github.com/nautobot/nautobot/pull/1108 - it appears we shouldn't be using `nautobot-server health_check` for the celery worker.

development/docker-compose.base.yml
```diff
@@ -25,8 +25,27 @@ services:
...
+    depends_on:
+      - "nautobot"
+      - "redis"
+    entrypoint:
+      - "sh"
+      - "-c"  # this is to evaluate the $NAUTOBOT_LOG_LEVEL from the env
+      - "nautobot-server celery worker -B -l $$NAUTOBOT_LOG_LEVEL"  # $$ because of docker-compose
```
glennmatthews, line 50: No need for `-B` here.

development/nautobot_config.py
```diff
@@ -359,7 +321,7 @@ def is_truthy(arg):
...
 DEBUG_TOOLBAR_CONFIG = {"SHOW_TOOLBAR_CALLBACK": lambda _request: DEBUG and not TESTING}
 
-if "debug_toolbar" not in EXTRA_INSTALLED_APPS:
-    EXTRA_INSTALLED_APPS.append("debug_toolbar")
-if "debug_toolbar.middleware.DebugToolbarMiddleware" not in settings.MIDDLEWARE:
-    settings.MIDDLEWARE.insert(0, "debug_toolbar.middleware.DebugToolbarMiddleware")
+if "debug_toolbar" not in INSTALLED_APPS:  # noqa: F405
+    INSTALLED_APPS.append("debug_toolbar")  # noqa: F405
+if "debug_toolbar.middleware.DebugToolbarMiddleware" not in MIDDLEWARE:  # noqa: F405
+    MIDDLEWARE.insert(0, "debug_toolbar.middleware.DebugToolbarMiddleware")  # noqa: F405
```
glennmatthews, line 327: It might be good to wrap this block in an `if DEBUG` so that we don't install the toolbar if running with DEBUG=False?

nautobot_ssot/jobs/examples.py
```diff
@@ -59,6 +61,25 @@ class SiteModel(DiffSyncModel):
...
+
+    # Metadata about this model
+    _modelname = "prefix"
+    _identifiers = ("prefix", "tenant_slug")
+    # To keep this example simple, we don't include **all** attributes of a Prefix here. But you could!
+    _attributes = ("description", "status_slug")
+
+    # Data type declarations for all identifiers and attributes
+    prefix: str
+    tenant_slug: str
```
glennmatthews, line 75: tenant_slug can be None, right?

```suggestion
    tenant_slug: Optional[str]
```

nautobot_ssot/jobs/examples.py
```diff
@@ -252,6 +316,60 @@ def delete(self):
...
+            diffsync (NautobotLocal): DiffSync adapter owning this Prefix
+            ids (dict): Initial values for this model's _identifiers
+            attrs (dict): Initial values for this model's _attributes
+        """
+        prefix = Prefix(prefix=ids["prefix"], description=attrs["description"])
+        tenant_obj, _ = Tenant.objects.get_or_create(
+            slug=ids["tenant_slug"],
+            defaults={"name": ids["tenant_slug"]},
+        )
+        prefix.tenant = tenant_obj
```
glennmatthews, line 336: Don't we need to handle the case where tenant_slug is None?

nautobot_ssot/jobs/examples.py
```diff
@@ -252,6 +316,60 @@ def delete(self):
...
+                defaults={"name": attrs["status_slug"]},
+            )
+            status_obj.content_types.add(ContentType.objects.get_for_model(Prefix))
+            prefix.status = status_obj
+        prefix.validated_save()
+        return super().update(attrs)
+
+    def delete(self):
+        """Delete an existing Prefix record from local Nautobot."""
+        prefix = Prefix.objects.get(prefix=self.prefix, tenant_slug=self.tenant_slug)
```
glennmatthews, line 368:
```suggestion
        prefix = Prefix.objects.get(prefix=self.prefix, tenant__slug=self.tenant_slug)
```

nautobot_ssot/jobs/examples.py
```diff
@@ -369,14 +504,25 @@ def load(self):
...
             )
             self.add(site_model)
             self.job.log_debug(message=f"Loaded {site_model} from local Nautobot instance")
 
+        for prefix in Prefix.objects.all():
+            prefix_model = self.prefix(
+                prefix=str(prefix.prefix),
+                description=prefix.description,
+                status_slug=prefix.status.slug,
+                tenant_slug=prefix.tenant.slug if prefix.tenant else "",
```
glennmatthews, line 520:
```suggestion
                tenant_slug=prefix.tenant.slug if prefix.tenant else None,
```

pyproject.toml
```diff
@@ -28,8 +28,11 @@ packages = [
 
 [tool.poetry.dependencies]
 python = "^3.6"
-nautobot = "^1.0.3"
+nautobot = "^1.1.4"
 diffsync = "^1.3.0"
+# Added pinned typed-add to overcome issues with black version
+typed-ast = "^1.5.0"
+markdown = "<=3.3.4"
```
glennmatthews, line 35: Why these additions? If typed-ast is a Black dependency, it should be a dev-dependency not an main dependency. Markdown is a Nautobot dependency so this plugin absolutely should not be specifying its own version dependency on it.

pyproject.toml
```diff
@@ -28,8 +28,11 @@ packages = [
 
 [tool.poetry.dependencies]
 python = "^3.6"
-nautobot = "^1.0.3"
+nautobot = "^1.1.4"
```
glennmatthews, line 31: Are we dropping support for nautobot 1.0.x? Why?

nautobot_ssot/jobs/examples.py
chadell: because I was using the prefix and tenant_slug as identifiers it can't be `None` but `""`

nautobot_ssot/jobs/examples.py
chadell: this is true, I will have to convert to None tenant object when the tenant_slug is ""

glennmatthews (approved):

Issue #11, opened 2021-10-28
Capture and report more performance information


dgarros:
### Environment
* Nautobot version:  1.1.3
* nautobot-ssot version:  1.0.1

### Proposed Functionality
Currently only the total duration of the Job execution is being captured and reported in the Sync Result view.

It would be great to have more information available to help identify any performance issue like  : 
- individual time to load each adapter
- time to execute the diff
- time to execute the sync
- memory utilized

 ### Use Case
Improve developer experience and help identify any performance issue

chadell:
@dgarros I guess this issue is solved by #15?

chadell:
Added in #15

PR #15, opened 2021-12-01
Propose refactor to capture performance data from SSoT execution

chadell:
will address #11 

Take this first commit as a proposal of how we could get more insights from SSoT executions.
The idea is to split the current `sync_data` into 4 methods:
* `load_source_adapter`
* `load_target_adapter`
* `calculate_diff`
* `execute_sync`

The benefits I see from this are:
* It's backwards compatible, so you can still overwrite completely the `sync_data` 
* It gives the future plugin developers a more clear implementation guide
* Obviously, it enables us to inject extra performance measurements around this child methods


nautobot_ssot/jobs/base.py
```diff
@@ -47,17 +48,56 @@ class DataSyncBaseJob(BaseJob):
...
 
+    def load_source_adapter(self):
+        """TODO:."""
+        pass
+
+    def load_target_adapter(self):
+        """TODO:."""
+        pass
+
+    def calculate_diff(self):
```
glennmatthews, line 59: Similarly, this should either return the calculated diff, or consistently set it as `self.diff`, so that we can take advantage of https://github.com/networktocode/diffsync/pull/80 and avoid re-calculating the diff unnecessarily.

nautobot_ssot/jobs/base.py
```diff
@@ -47,17 +48,56 @@ class DataSyncBaseJob(BaseJob):
 
     dry_run = BooleanVar()
 
+    def load_source_adapter(self):
+        """TODO:."""
+        pass
+
+    def load_target_adapter(self):
```
glennmatthews, line 55: We should establish a consistent expectation here - should these methods take the source and target adapter as inputs? Or are they expected to construct the source/target adapter as part of the method? Or is the expectation that we'll already have `self.source_adapter` and `self.target_adapter` as instance properties? We should pick one and establish it as the standard.

nautobot_ssot/jobs/base.py
```diff
@@ -47,17 +48,56 @@ class DataSyncBaseJob(BaseJob):
...
+        """TODO:."""
+        pass
+
+    def calculate_diff(self):
+        """TODO:."""
+        pass
+
+    def execute_sync(self):
+        """TODO:."""
+        pass
```
glennmatthews, line 65: I think we should be able to provide default implementations of at least `calculate_diff` and `execute_sync`, as in many cases concrete subclasses shouldn't need to override these.

nautobot_ssot/jobs/base.py
```diff
@@ -47,17 +48,56 @@ class DataSyncBaseJob(BaseJob):
...
         Available instance attributes include:
 
         - self.kwargs     (corresponds to the Job's `data` input, including 'dry_run' option)
         - self.commit     (should generally be True)
         - self.sync       (Sync instance tracking this job execution)
         - self.job_result (as per Job API)
         """
-        pass  # pylint: disable=unnecessary-pass
+        # TODO: most of the log messages from the methods could be standarised here
+        # example: self.log_info(message="Loading current data from source adapter...")
```
glennmatthews, line 80: I agree, these should be logged automatically in the default `sync_data` implementation.

nautobot_ssot/jobs/base.py
```diff
@@ -47,17 +48,56 @@ class DataSyncBaseJob(BaseJob):
...
+            f"Local Load Time: {load_target_adapter_time - load_source_adapter_time}\n",
+            f"Diff Time: {calculate_diff_time - load_target_adapter_time}\n",
+        )
+
+        if not self.kwargs["dry_run"]:
+            self.execute_sync()
+            execute_sync_time = datetime.now()
+            time_message += (f"Sync Time: {execute_sync_time - calculate_diff_time}",)
+
+        self.log_info(message=time_message)
```
glennmatthews, line 100: Personally I'd rather see four separate timing log messages, each at the appropriate point in the logs, instead of one combined message at the end of the logs.

glennmatthews (commented):
I like the idea here!

nautobot_ssot/jobs/base.py
dgarros: Could we provide some sane default for all of them ?

nautobot_ssot/jobs/base.py
dgarros: Could we store these datapoints in the Sync model ? it would be easier to query, analyze, compare and expose to other systems if needed

nautobot_ssot/jobs/base.py
dgarros: My preference would be to have `self.source_adapter` and `self.target_adapter` as instance properties. I think it would make it easier to provide default method if we can access the adapters from the instance directly

nautobot_ssot/jobs/base.py
chadell: I think so, the one from the example for instance

nautobot_ssot/jobs/base.py
chadell: good idea

nautobot_ssot/jobs/base.py
```diff
@@ -33,31 +36,135 @@
...
+        """
+        raise NotImplementedError
+
+    def calculate_diff(self):
+        """Method to calculate the difference from SOURCE to TARGET adapter and store in `self.diff`.
+
+        This is a generic implementation that you could overwrite completely in you custom logic.
+        """
+        if self.source_adapter and self.target_adapter:
+            self.diffsync_flags = (
```
dgarros, line 80: I think it would be good to check if `diffsync_flags` is already defined first, this way it's possible to provide a different set of flags without overwriting the entire method

dgarros (approved):
Looks good to me, great work Christian

nautobot_ssot/jobs/base.py
chadell: sure, I will move the definition to init, so anyone would be able to overwrite them at any time

glennmatthews: Possibly make this a class-level attribute `cls.default_diffsync_flags` or such?

nautobot_ssot/jobs/base.py
```diff
@@ -33,31 +36,135 @@
...
+                | DiffSyncFlags.SKIP_UNMATCHED_DST
+            )
+            self.diff = self.source_adapter.diff_to(self.target_adapter, flags=self.diffsync_flags)
+            self.sync.diff = self.diff.dict()
+            self.sync.save()
+
+    def execute_sync(self):
+        """Method to synchronize the difference from `self.diff`, from SOURCE to TARGET adapter.
+
+        This is a generic implementation that you could overwrite completely in you custom logic.
```
glennmatthews:
```suggestion
        This is a generic implementation that you could overwrite completely in your custom logic.
```

nautobot_ssot/jobs/base.py
```diff
@@ -33,31 +36,135 @@
...
+
+        start_time = datetime.now()
+
+        self.log_info(message="Loading current data from source adapter...")
+        self.load_source_adapter()
+        load_source_adapter_time = datetime.now()
+        self.sync.load_time_source = load_source_adapter_time - start_time
+        self.sync.save()
+        self.log_info(
+            message=f"Source Load Time from {self.source_adapter.__class__.__name__}: {self.sync.load_time_source}"
```
glennmatthews, line 123: DiffSync adapters have a perfectly good `__str__` implementation:

```suggestion
            message=f"Source Load Time from {self.source_adapter}: {self.sync.load_time_source}"
```

nautobot_ssot/jobs/base.py
```diff
@@ -33,31 +36,135 @@
...
+        )
+
+        self.log_info(message="Loading current data from target adapter...")
+        self.load_target_adapter()
+        load_target_adapter_time = datetime.now()
+        self.sync.load_time_target = load_target_adapter_time - load_source_adapter_time
+        self.sync.save()
+        self.log_info(
+            message=f"Target Load Time from {self.target_adapter.__class__.__name__}: {self.sync.load_time_target}"
```
glennmatthews, line 132:
```suggestion
            message=f"Target Load Time from {self.target_adapter}: {self.sync.load_time_target}"
```

nautobot_ssot/jobs/base.py
```diff
@@ -33,31 +36,135 @@
...
+        self.log_info(message="Calculating diffs...")
+        self.calculate_diff()
+        calculate_diff_time = datetime.now()
+        self.sync.diff_time = calculate_diff_time - load_target_adapter_time
+        self.sync.save()
+        self.log_info(message=f"Diff Time: {self.sync.diff_time}")
+
+        if not self.kwargs["dry_run"]:
+            self.log_info(
+                message=f"Syncing from {self.source_adapter.__class__.__name__} to {self.target_adapter.__class__.__name__}..."
```
glennmatthews, line 144:
```suggestion
                message=f"Syncing from {self.source_adapter} to {self.target_adapter}..."
```

nautobot_ssot/jobs/base.py
```diff
@@ -33,31 +36,130 @@
...
+        self.log_info(
+            message=f"Target Load Time from {self.target_adapter.__class__.__name__}: {self.sync.load_time_target}"
+        )
+
+        self.log_info(message="Calculating diffs...")
+        self.calculate_diff()
+        calculate_diff_time = datetime.now()
+        self.sync.diff_time = calculate_diff_time - load_target_adapter_time
+        self.sync.save()
+        self.log_info(message=f"Diff Time: {self.sync.diff_time}")
```
glennmatthews, line 140:
```suggestion
        self.log_info(message=f"Diff Calculation Time: {self.sync.diff_time}")
```

nautobot_ssot/jobs/base.py
```diff
@@ -33,31 +36,130 @@
...
+            self.log_info(message=f"Sync Time: {self.sync.sync_time}")
+
+        if self.kwargs["memory_profiling"]:
+            memory_size, memory_peak = tracemalloc.get_traced_memory()
+            self.sync.memory_usage = tracemalloc.get_tracemalloc_memory()
+            self.sync.save()
+            self.log_info(
+                message=(
+                    f"Traced Memory (Current, Peak): {memory_size}, {memory_peak}\n",
+                    f"Memory Usage: {self.sync.memory_usage}",
```
glennmatthews, line 160: Would probably be useful to include units here (bytes? kilobytes?)

nautobot_ssot/jobs/base.py
```diff
@@ -147,6 +252,13 @@ def __init__(self):
         self.sync = None
         self.kwargs = {}
         self.commit = False
+        self.diff = None
+        self.source_adapter = None
+        self.target_adapter = None
+        # Default diffsync flags. You can overwrite them at any time.
+        self.diffsync_flags = (
+            DiffSyncFlags.CONTINUE_ON_FAILURE | DiffSyncFlags.LOG_UNCHANGED_RECORDS | DiffSyncFlags.SKIP_UNMATCHED_DST
```
glennmatthews, line 260: I don't think we should default `SKIP_UNMATCHED_DST` as a flag - that should be opt-in.

nautobot_ssot/jobs/base.py
```diff
@@ -210,10 +322,12 @@ def run(self, data, commit):
             self.log_failure(message=f"An exception occurred: `{type(exc).__name__}: {exc}`\n```\n{stacktrace}\n```")
 
 
+# pylint: disable=abstract-method
 class DataSource(DataSyncBaseJob):
     """Base class for Jobs that sync data **from** another data source **to** Nautobot."""
 
     dry_run = BooleanVar(description="Perform a dry-run, making no actual changes to Nautobot data.")
+    memory_profiling = BooleanVar(description="Perform a memory profiling analysis.", default=False)
```
glennmatthews, line 330: Seems like this could be set as-is on the `DataSyncBaseJob` class, without needing to be overridden by the DataSource and DataTarget subclasses, since it doesn't differ between them (unlike `dry_run` which has a different description for each).

nautobot_ssot/models.py
```diff
@@ -47,6 +47,11 @@ class Sync(BaseModel):
 
     start_time = models.DateTimeField(blank=True, null=True)
     # end_time is represented by the job_result.completed field
+    load_time_source = models.DurationField(blank=True, null=True)
+    load_time_target = models.DurationField(blank=True, null=True)
+    diff_time = models.DurationField(blank=True, null=True)
+    sync_time = models.DurationField(blank=True, null=True)
+    memory_usage = models.IntegerField(blank=True, null=True)
```
glennmatthews, line 54: Planning to add these to the sync detail page template? In general what's your intended purpose of storing these in the DB given they're already also present in the job logs?

chadell: the idea is following @dgarros 's point about storing them for future retrieval from external source, not only as text.
The idea would to present them as metrics, or directly in the UI/API (not done yet, to confirm).

nautobot_ssot/jobs/base.py
```diff
@@ -33,31 +36,124 @@
...
+        - self.job_result (as per Job API)
+        """
+        raise NotImplementedError
+
+    def calculate_diff(self):
+        """Method to calculate the difference from SOURCE to TARGET adapter and store in `self.diff`.
+
+        This is a generic implementation that you could overwrite completely in your custom logic.
+        """
+        if self.source_adapter and self.target_adapter:
```
glennmatthews, line 79: If either or both are missing, should we at least log a warning, rather than silently doing a no-op?

nautobot_ssot/jobs/base.py
```diff
@@ -33,31 +36,124 @@
...
+            self.diff = self.source_adapter.diff_to(self.target_adapter, flags=self.diffsync_flags)
+            self.sync.diff = self.diff.dict()
+            self.sync.save()
+
+    def execute_sync(self):
+        """Method to synchronize the difference from `self.diff`, from SOURCE to TARGET adapter.
+
+        This is a generic implementation that you could overwrite completely in your custom logic.
+        """
+        if self.source_adapter and self.target_adapter:
```
glennmatthews, line 89: Same comment here.

nautobot_ssot/jobs/base.py
```diff
@@ -33,31 +36,124 @@
...
+        self.log_info(message=f"Target Load Time from {self.target_adapter}: {self.sync.load_time_target}")
+
+        self.log_info(message="Calculating diffs...")
+        self.calculate_diff()
+        calculate_diff_time = datetime.now()
+        self.sync.diff_time = calculate_diff_time - load_target_adapter_time
+        self.sync.save()
+        self.log_info(message=f"Diff Calculation Time: {self.sync.diff_time}")
+
+        if not self.kwargs["dry_run"]:
```
glennmatthews, line 138:
```suggestion
        if self.kwargs["dry_run"]:
            self.log_info("As `dry_run` is set, skipping the actual data sync.")
        else:
```

nautobot_ssot/jobs/base.py
```diff
@@ -33,31 +36,124 @@
...
+            self.log_info(message=f"Syncing from {self.source_adapter} to {self.target_adapter}...")
+            self.execute_sync()
+            execute_sync_time = datetime.now()
+            self.sync.sync_time = execute_sync_time - calculate_diff_time
+            self.sync.save()
+            self.log_info(message="Sync complete")
+            self.log_info(message=f"Sync Time: {self.sync.sync_time}")
+
+        if self.kwargs["memory_profiling"]:
+            memory_size, memory_peak = tracemalloc.get_traced_memory()
```
glennmatthews, line 148: Any reason not to save these two values to `self.sync` as well? Are they less useful?

chadell: fixed

PR #18, opened 2021-12-16
Avoid skipping the diff and sync if one of the adapters is empty

glennmatthews:
Due to a behavioral quirk (bug?) in diffsync, a DIffSync adapter that contains no data will evaluate as falsey:

```python
>>> diffsync = DiffSync()
>>> bool(diffsync)
False
>>> diffsync.add(DiffSyncModel())
>>> bool(diffsync)
True
```

Change the checks that we have on the existence of the two adapters to explicit `is not None` checks to bypass this behavior.


chadell (approved):
LGTM

PR #21, opened 2021-12-23
Add performance stats in detail view

chadell:
View with no memory stats
image

View with memory stats
image


nautobot_ssot/templates/nautobot_ssot/sync_detail.html
```diff
@@ -119,6 +119,41 @@
                         </td>
                     </tr>
                 </table>
+                <div class="panel-heading">
```
glennmatthews, line 122: This should be a separate panel from the one above.

```suggestion
            </div>
            <div class="panel panel-default">
                <div class="panel-heading">
```

nautobot_ssot/templates/nautobot_ssot/sync_detail.html
```diff
@@ -119,6 +119,41 @@
...
+                    <strong>Performance stats</strong>
```
glennmatthews, line 123:
```suggestion
                    <strong>Performance Stats</strong>
```

nautobot_ssot/templates/nautobot_ssot/sync_detail.html
```diff
@@ -119,6 +119,41 @@
...
+                </div>
+                <table class="table table-hover panel-body attr-table">
+                    <tr>
+                        <td>Time loading Source</td>
```
glennmatthews, line 127:
```suggestion
                        <td>Time Loading Source</td>
```

nautobot_ssot/templates/nautobot_ssot/sync_detail.html
```diff
@@ -119,6 +119,41 @@
...
+                        <td>{{ object.load_time_source | shorter_timedelta }}</td>
+                    </tr>
+                    <tr>
+                        <td>Time loading Target</td>
```
glennmatthews, line 131:
```suggestion
                        <td>Time Loading Target</td>
```

nautobot_ssot/templates/nautobot_ssot/sync_detail.html
```diff
@@ -119,6 +119,41 @@
...
+                        <td>Memory usage</td>
+                        <td>{{ object.memory_usage }} bytes</td>
+                    </tr>
+                    <tr>
+                        <td>Memory blocks</td>
+                        <td>{{ object.memory_size }} blocks</td>
+                    </tr>
+                    <tr>
+                        <td>Memory peak blocks</td>
+                        <td>{{ object.memory_peak }} blocks</td>
```
glennmatthews, line 153: Consider using the [`intcomma`](https://docs.djangoproject.com/en/3.2/ref/contrib/humanize/#intcomma) filter to render these large numbers in a more readable way, i.e.:

```jinja2
{% load humanize %}
...
                        <td>{{ object.memory_usage | intcomma }} bytes</td>
```

nautobot_ssot/templates/nautobot_ssot/sync_detail.html
```diff
@@ -119,6 +119,41 @@
...
+                    <tr>
+                        <td>Memory usage</td>
+                        <td>{{ object.memory_usage }} bytes</td>
+                    </tr>
+                    <tr>
+                        <td>Memory blocks</td>
+                        <td>{{ object.memory_size }} blocks</td>
+                    </tr>
+                    <tr>
+                        <td>Memory peak blocks</td>
```
glennmatthews, line 152:
```suggestion
                        <td>Time Calculating Diff</td>
                        <td>{{ object.diff_time | shorter_timedelta }}</td>
                    </tr>
                    <tr>
                        <td>Time Performing Sync</td>
                        <td>{{ object.sync_time | shorter_timedelta }}</td>
                    </tr>
                    {% if object.memory_usage %}
                    <tr>
                        <td>Memory Usage</td>
                        <td>{{ object.memory_usage }} bytes</td>
                    </tr>
                    <tr>
                        <td>Memory Blocks</td>
                        <td>{{ object.memory_size }} blocks</td>
                    </tr>
                    <tr>
                        <td>Memory Peak Blocks</td>
```

glennmatthews, line 152: From a user perspective I don't find these three items very clear. Are `Memory Usage` and `Memory Blocks` measuring the same thing, just in bytes versus blocks? If those two are *not* `Peak` usage, what are they? Can we change any or all of these headings to be clearer in their meaning?

glennmatthews:
Actually, what do you think about potentially moving the *time-based* performance data over to the left, adjacent to the "Duration" item, something like:

```
Duration        0:01:43.4 total
                * 0:00:36.4 loading from {{ object.source }}
                * 0:00:06.4 loading from {{ object.target }}
                * 0:00:02.1 calculating diffs
                * 0:01:00.5 performing sync
```

dgarros:
> Actually, what do you think about potentially moving the _time-based_ performance data over to the left, adjacent to the "Duration" item, something like:
> 
> ```
> Duration        0:01:43.4 total
>                 * 0:00:36.4 loading from {{ object.source }}
>                 * 0:00:06.4 loading from {{ object.target }}
>                 * 0:00:02.1 calculating diffs
>                 * 0:01:00.5 performing sync
> ```

I think that make sense

nautobot_ssot/templates/nautobot_ssot/sync_detail.html
chadell: I agree with you, this must be improved. I will review it and propose a better implementation of the idea, maybe changing a bit how this is tracked.

chadell:
Changes in the HTML consolidating duration data
image

chadell:
@glennmatthews , following your point about memory stats, I refactored it a bit, because I was getting a useless data (memory used by the `tracemalloc` library itself) and now, the approach is, similar to the time performance, for every step, we capture the final memory usage and the peak usage (clearing stats in between), so it will give a better understanding of how "efficient" each step was.

> Because the previous commit was not actually released, I would propose to merge both migration files, do you agree?

image

chadell:
I tried rendering the memory usage stats as a bullet list (above) and as a table (current commit)

image

nautobot_ssot/jobs/base.py
```diff
@@ -110,6 +110,18 @@ def sync_data(self):
...
+        def record_memory_trace(step: str):
+            """Helper function to record memory usage and reset tracemalloc stats."""
+            memory_size, memory_peak = tracemalloc.get_traced_memory()
+            setattr(self.sync, f"{step}_memory_size", memory_size)
+            setattr(self.sync, f"{step}_memory_peak", memory_peak)
+            self.sync.save()
+            self.log_info(
+                message=(f"Traced memory for {step} (Current, Peak): {memory_size} bytes, {memory_peak} bytes")
+            )
+            tracemalloc.clear_traces()
```
glennmatthews, line 123: I'm still struggling to understand the meaning/significance of these various stats, even after a quick skim of the `tracemalloc` docs. I gather that by using `clear_traces` at each stage, we're getting only incremental memory stats for that stage - so the total memory used by the overall sync process is at least (sum of all "current" measurements)? And for any given stage, the actual peak memory usage is (sum of all previous stages' "current" measurements plus the "peak" measurement for the current stage)?

I think at a minimum it would be good to have a paragraph or two in the README explaining how to read these statistics and what meaning to take from them.

nautobot_ssot/templates/nautobot_ssot/sync_detail.html
```diff
@@ -120,6 +129,77 @@
...
+                            </table>
+                        </td>
+                    </tr>
+                    <tr>
+                        <td>Memory used loading {{ object.target }}</td>
+                        <td>
+                            <table class="table table-hover panel-body attr-table">
+                                <tr>
+                                    <td>Final</td>
+                                    <td>{{ object.load_target_memory_size | intcomma }} bytes</td>
```
glennmatthews, line 159: Using `intcomma` is an improvement but the values are still hard to read as raw byte numbers, I think. Maybe we should actually add a custom template filter (similar to Nautobot core's `humanize_speed`) that would render values as "11.42 MiB", etc. rather than "11,420,190 bytes"?

nautobot_ssot/templates/nautobot_ssot/sync_detail.html
```diff
@@ -58,7 +59,15 @@
...
                     <tr>
                         <td>Duration</td>
-                        <td>{{ object.duration | shorter_timedelta }}</td>
+                        <td>
+                            {{ object.duration | shorter_timedelta }} total
+                            <ul>
+                                <li>{{ object.load_time_source | shorter_timedelta }} loading from {{ object.source }}</li>
+                                <li>{{ object.load_time_target | shorter_timedelta }} loading from {{ object.target }}</li>
+                                <li>{{ object.diff_time | shorter_timedelta }} calulating diffs</li>
+                                <li>{{ object.sync_time | shorter_timedelta }} performing sync</li>
```
glennmatthews, line 68: If we're rewriting migrations anyway, my vote would be to also rename `load_time_source` and `load_time_target` to `source_load_time` and `target_load_time` so as to make them more consistent with `diff_time` and `sync_time`.

nautobot_ssot/templates/nautobot_ssot/sync_detail.html
```diff
@@ -120,6 +129,77 @@
...
+                            </table>
+                        </td>
+                    </tr>
+                    <tr>
+                        <td>Memory used calculating diff</td>
+                        <td>
+                            <table class="table table-hover panel-body attr-table">
+                                <tr>
+                                    <td>Final</td>
+                                    <td>{{ object.diff_memory_size | intcomma }} bytes</td>
```
glennmatthews, line 174: Would it be more correct to rename the fields to `diff_memory_final` instead of `diff_memory_size`, etc.?

nautobot_ssot/jobs/base.py
```diff
@@ -110,6 +110,18 @@ def sync_data(self):
...
         """
+
+        def record_memory_trace(step: str):
+            """Helper function to record memory usage and reset tracemalloc stats."""
+            memory_size, memory_peak = tracemalloc.get_traced_memory()
+            setattr(self.sync, f"{step}_memory_size", memory_size)
+            setattr(self.sync, f"{step}_memory_peak", memory_peak)
+            self.sync.save()
+            self.log_info(
+                message=(f"Traced memory for {step} (Current, Peak): {memory_size} bytes, {memory_peak} bytes")
```
glennmatthews, line 121: Would it be more consistent and correct to log as `(Final, Peak)` instead of `(Current, Peak)`?

nautobot_ssot/jobs/base.py
chadell: Your understanding is the same as mine, so for each step we capture the final memory allocation and the peak.
I will improve the README for this stats, for developers to understand what they mean

nautobot_ssot/models.py
```diff
@@ -51,9 +51,14 @@ class Sync(BaseModel):
...
-    memory_size = models.IntegerField(blank=True, null=True)
-    memory_peak = models.IntegerField(blank=True, null=True)
+    load_source_memory_size = models.IntegerField(blank=True, null=True)
+    load_source_memory_peak = models.IntegerField(blank=True, null=True)
+    load_target_memory_size = models.IntegerField(blank=True, null=True)
+    load_target_memory_peak = models.IntegerField(blank=True, null=True)
+    diff_memory_size = models.IntegerField(blank=True, null=True)
+    diff_memory_peak = models.IntegerField(blank=True, null=True)
+    sync_memory_size = models.IntegerField(blank=True, null=True)
+    sync_memory_peak = models.IntegerField(blank=True, null=True)
```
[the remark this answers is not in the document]
chadell: nice! you always teach me something new 😄

nautobot_ssot/templates/nautobot_ssot/sync_detail.html
```diff
@@ -120,6 +129,77 @@
...
+                                    <td>Final</td>
+                                    <td>{{ object.load_source_memory_size | intcomma }} bytes</td>
+                                </tr>
+                                <tr>
+                                    <td>Peak</td>
+                                    <td>{{ object.load_source_memory_size | intcomma }} bytes</td>
+                                </tr>
+                            </table>
+                        </td>
+                    </tr>
```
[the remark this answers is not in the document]
chadell: I was not even aware of `rowspan` 😓

nautobot_ssot/templates/nautobot_ssot/sync_detail.html
chadell: understood

nautobot_ssot/templates/nautobot_ssot/sync_detail.html
```diff
@@ -120,6 +129,77 @@
...
+                                    <td>{{ object.diff_memory_size | intcomma }} bytes</td>
+                                </tr>
+                                <tr>
+                                    <td>Peak</td>
+                                    <td>{{ object.diff_memory_peak | intcomma }} bytes</td>
+                                </tr>
+                            </table>
+                        </td>
+                    </tr>
+                    {% if object.sync_memory_size %}
```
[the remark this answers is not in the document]
chadell: because maybe it was coming from a `dry-run` so there is no `sync` data to render. does it make sense?

nautobot_ssot/templates/nautobot_ssot/sync_detail.html
chadell: so, from your message, I understand that you agree about rewriting migrations?

nautobot_ssot/templates/nautobot_ssot/sync_detail.html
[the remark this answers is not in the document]
glennmatthews: Gotcha, thanks!

nautobot_ssot/templates/nautobot_ssot/sync_detail.html
glennmatthews: Yes, I think that should be fine.

chadell:
New look
Screenshot 2022-01-12 at 07 19 17

docs/developing_jobs.md
```diff
@@ -8,35 +8,55 @@ In brief, the following general steps can be followed:
...
+     ```python
+     def source_load_adapter(self):
+         """Method to instantiate and load the SOURCE adapter into `self.source_adapter`."""
+         self.source_adapter = NautobotRemote(url=self.kwargs["source_url"], token=self.kwargs["source_token"], job=self)
+         self.source_adapter.load()
+     ```
+
+   - `self.calculate_diff`: This method is implemented by default, using the output from load_adapter methods.
+
+   - `self.execue_sync`: This method is implemented by default, using the output from load_adapter methods. Only executed if it's not a `dry-run` execution.
```
glennmatthews, line 40:
```suggestion
   - `self.execute_sync`: This method is implemented by default, using the output from load_adapter methods. Only executed if it's not a `dry-run` execution.
```

docs/developing_jobs.md
```diff
@@ -8,35 +8,55 @@ In brief, the following general steps can be followed:
...
 6. Install your Job via any of the supported Nautobot methods (installation into the `JOBS_ROOT` directory, inclusion in a Git repository, or packaging as part of a plugin) and it should automatically become available!
+
+## Analyze Job performance
+
+The built-in implementation of `sync_data` is composed by 4 steps:
+
+- Load data from source adapter
+- Load data from target adapter
+- Calculating diff
+- Executing synchronization (if not `dry-run)
```
glennmatthews, line 52:
```suggestion
- Loading data from source adapter
- Loading data from target adapter
- Calculating diff
- Executing synchronization (if not `dry-run`)
```

nautobot_ssot/jobs/base.py
```diff
@@ -51,7 +51,7 @@ class DataSyncBaseJob(BaseJob):  # pylint: disable=too-many-instance-attributes
     dry_run = BooleanVar()
     memory_profiling = BooleanVar(description="Perform a memory profiling analysis.", default=False)
 
-    def load_source_adapter(self):
+    def source_load_adapter(self):
```
glennmatthews, line 54: Not a fan of this function name change myself. Either `load_source_adapter` (my preference) or `source_adapter_load` would be better.

nautobot_ssot/templates/nautobot_ssot/sync_detail.html
```diff
@@ -129,72 +131,48 @@
...
-                                    <td>Final</td>
-                                    <td>{{ object.diff_memory_size | intcomma }} bytes</td>
-                                </tr>
-                                <tr>
-                                    <td>Peak</td>
-                                    <td>{{ object.diff_memory_peak | intcomma }} bytes</td>
-                                </tr>
-                            </table>
-                        </td>
+                        <td rowspan="2">Memory used loading {{ object.targe }}</td>
```
glennmatthews, line 150:
```suggestion
                        <td rowspan="2">Memory used loading {{ object.target }}</td>
```

nautobot_ssot/templatetags/humanize_size.py
```diff
@@ -0,0 +1,23 @@
+"""Simple template tag to render a shorter representation of a timedelta object."""
+
+from django import template
+
+
+register = template.Library()
+
+
+@register.filter()
+def humanize_size(size):
```
glennmatthews, line 10: I think `humanize_bytes` would be a better name.

nautobot_ssot/templatetags/humanize_size.py
```diff
@@ -0,0 +1,23 @@
...
+    """Humanize memory size given in bytes.
+
+    Examples:
+        1 => "1.00 B"
```
glennmatthews, line 14: Would be nice if this could be just `1 B` since we don't really deal in fractions of bytes. Not critical though.

docs/developing_jobs.md
```diff
@@ -8,35 +8,55 @@ In brief, the following general steps can be followed:
...
 
-            if not self.kwargs["dry_run"]:
-                self.log_info(message="Syncing from Data Source to Nautobot...")
-                diffsync1.sync_to(diffsync2, flags=diffsync_flags)
-                self.log_info(message="Sync complete")
+     ```python
+     def load_source_adapter(self):
+         """Method to instantiate and load the SOURCE adapter into `self.source_adapter`."""
+         self.source_adapter = NautobotRemote(url=self.kwargs["source_url"], token=self.kwargs["source_token"], job=self)
+         self.source_adapter.load()
```
glennmatthews, line 35: Needs to be changed to `load_target_adapter` and updated accordingly

nautobot_ssot/jobs/base.py
```diff
@@ -110,6 +110,18 @@ def sync_data(self):
         - self.sync       (Sync instance tracking this job execution)
```
glennmatthews, line 110: New screenshot (above) is quite large - at the very least, consider narrowing your browser window to the minimum size that still shows all of the data usefully, as there's a lot of horizontal whitespace that could be omitted.

nautobot_ssot/models.py
```diff
@@ -47,13 +47,18 @@ class Sync(BaseModel):
...
-    memory_size = models.IntegerField(blank=True, null=True)
-    memory_peak = models.IntegerField(blank=True, null=True)
+    source_load_memory_final = models.PositiveBigIntegerField(blank=True, null=True)
+    source_load_memory_peak = models.PositiveBigIntegerField(blank=True, null=True)
+    target_load_memory_final = models.PositiveBigIntegerField(blank=True, null=True)
+    target_load_memory_peak = models.PositiveBigIntegerField(blank=True, null=True)
+    diff_memory_final = models.PositiveBigIntegerField(blank=True, null=True)
+    diff_memory_peak = models.PositiveBigIntegerField(blank=True, null=True)
+    sync_memory_final = models.PositiveBigIntegerField(blank=True, null=True)
+    sync_memory_peak = models.PositiveBigIntegerField(blank=True, null=True)
```
glennmatthews, line 61: Would it be better to default these to 0 instead of allowing them to be blank/null? Not sure.

nautobot_ssot/templates/nautobot_ssot/sync_detail.html
```diff
@@ -3,6 +3,9 @@
 {% load plugins %}
 {% load shorter_timedelta %}
 {% load render_diff %}
+{% load humanize %}
```
glennmatthews, line 6: I *think* we're not using humanize any more?

nautobot_ssot/models.py
chadell: I thought about defaulting to `null` to be more explicit about data that has not been gathered at all and avoid confusing with a execution that is not actually consuming any memory (i.e. a simple `pass`) so we can easily understand if the data was actually the "default" vs something actually gathered.
But maybe you have some other point in mind that I am not considering?

glennmatthews: No, that's a fair point. Thanks.

glennmatthews (approved):

Issue #41, opened 2022-05-02
KeyError with attribute type dict


Thetacz:
### Environment
* Python version:  3.6.15
* Nautobot version:  1.2.10
* nautobot-ssot version:  1.1.0

### Expected Behavior
Being able to open job history detail in SSoT

### Observed Behavior
![image](https://user-images.githubusercontent.com/26114847/166208473-07691f06-e345-4220-9794-e9a65d8f6063.png)
```
08:44:37.325 ERROR   django.request :
  Internal Server Error: /plugins/ssot/history/82d43bc2-fcac-46b4-aee4-237fd6a58c73/
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/usr/local/lib/python3.6/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/local/lib/python3.6/site-packages/django/views/generic/base.py", line 70, in view
    return self.dispatch(request, *args, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/nautobot/utilities/views.py", line 94, in dispatch
    return super().dispatch(request, *args, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/django/views/generic/base.py", line 98, in dispatch
    return handler(request, *args, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/nautobot/core/views/generic.py", line 124, in get
    **self.get_extra_context(request, instance),
  File "/usr/local/lib/python3.6/site-packages/django/shortcuts.py", line 19, in render
    content = loader.render_to_string(template_name, context, request, using=using)
  File "/usr/local/lib/python3.6/site-packages/django/template/loader.py", line 62, in render_to_string
    return template.render(context, request)
  File "/usr/local/lib/python3.6/site-packages/django/template/backends/django.py", line 61, in render
    return self.template.render(context)
  File "/usr/local/lib/python3.6/site-packages/django/template/base.py", line 170, in render
    return self._render(context)
  File "/usr/local/lib/python3.6/site-packages/django/template/base.py", line 162, in _render
    return self.nodelist.render(context)
  File "/usr/local/lib/python3.6/site-packages/django/template/base.py", line 938, in render
    bit = node.render_annotated(context)
[...]

PR #42, opened 2022-05-04
Escape dictionary curly bracket before renderingin HTML

chadell:
Addresses issue #41

### DIFF
--- a/nautobot_ssot/templatetags/render_diff.py
+++ b/nautobot_ssot/templatetags/render_diff.py
@@ -43,7 +43,7 @@ def render_diff_recursive(diff):
                 child_class = "diff-unchanged"
             else:
                 child_class = "diff-changed"
-            child_result += format_html('<li class="{}">{}<ul>', mark_safe(child_class), child)  # nosec
+            child_result += format_html('<li class="{}">{}<ul>', child_class, child)
 
             for attr, value in child_diffs.pop("+", {}).items():
                 child_result += format_html('<li class="diff-added">{}: {}</li>', attr, value)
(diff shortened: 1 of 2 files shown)


glennmatthews (commented):
I'm not sure this is the "best" solution to the issue - I'm concerned that it seems like we're double-interpreting a variable value somewhere in here, and this is maybe only a bandaid over a more fundamental design issue. Might be worth looking at how `render_diff` is wrapping this text into `format_html()` and see if that needs to be revisited somehow instead.

chadell:
The `result` that we send to `format_html(result)` contains this HTML: `<li class="diff-added">cfs: {'asw_owner': ''}</li>`

```
>>> from django.utils.html import format_html
>>> result = "<li class='diff-added'>cfs: {'asw_owner': ''}</li>"
>>> format_html(result)
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.6/site-packages/django/utils/html.py", line 115, in format_html
    return mark_safe(format_string.format(*args_safe, **kwargs_safe))
KeyError: "'asw_owner'"
```

From docs, `format_html(format_string, *args, **kwargs)`  takes that as a format string.
So, my suggestion would be to start using `mark_safe` instead.

chadell:
Any concern/idea about bandit issue with cross-site scripting?

glennmatthews:
> Any concern/idea about bandit issue with cross-site scripting?

Could certainly try a test or two with objects in the diff whose string representation is HTML or JS?

chadell:
In [6e3c3c0](https://github.com/nautobot/nautobot-plugin-ssot/pull/42/commits/6e3c3c06b577f812bd48d990b94d9ef35258239b) I replaced all the f-strings with input data by `format_html` what enforces `escape` to protect against XSS, and then when we reused this "safe" HTML code, we use the `mark_safe` to not escape it again.

glennmatthews (approved):
Can we have a unit test case or two get added for `render_diff`? Otherwise looks good to me!

Issue #47, opened 2022-05-09
`AttributeError` upon end of `calculate_diff`method - `'dict' object has no attribute summary`


Kircheneer:
### Environment
* Python version:  3.8.13
* Nautobot version:  1.3.3
* nautobot-ssot version:  1.1.1

### Expected Behavior

I expected the SSoT job not to crash.

### Observed Behavior

The SSoT job crashed.

```
An exception occurred: AttributeError: 'dict' object has no attribute 'summary'

Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/nautobot_ssot/jobs/base.py", line 327, in run
    self.sync_data()
  File "/usr/local/lib/python3.8/site-packages/nautobot_ssot/jobs/base.py", line 153, in sync_data
    self.calculate_diff()
  File "/usr/local/lib/python3.8/site-packages/nautobot_ssot/jobs/base.py", line 83, in calculate_diff
    self.log_info(message=self.sync.diff.summary())
AttributeError: 'dict' object has no attribute 'summary'
```

### Steps to Reproduce
1. TBD
2.
3.

Kircheneer:
Resolved with #48

PR #48, opened 2022-05-09
Use self.diff instead of self.sync.diff

chadell:
Addresses issue #47

TODO:
- add a test case to catch this issues


chadell:
@kircheneer could you check it?

Kircheneer:
Looks good to me 👍🏻

PR #54, opened 2022-07-09
Drops Python 3.6 Support

jvanderaa:
Dropping Python3.6 support. 

- Several packages have already dropped the support for end of support Python 3.6
- Believe this may be causing other SSOT plugins to still be doing a Python3.6 container image (and thus only i386 containers)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -104,7 +104,7 @@ jobs:
       fail-fast: true
       matrix:
         python-version: ["3.7", "3.8", "3.9"]
-        nautobot-version: ["1.2.11", "1.3.8"]
+        nautobot-version: ["1.2.11", "stable"]
     runs-on: "ubuntu-20.04"
     env:
       INVOKE_NAUTOBOT_SSOT_PYTHON_VER: "${{ matrix.python-version }}"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -104,7 +104,7 @@ jobs:
       fail-fast: true
       matrix:
         python-version: ["3.7", "3.8", "3.9"]
-        nautobot-version: ["1.0.3", "1.1.4"]
+        nautobot-version: ["1.0.3", "1.1.4", "1.2.11", "1.3.8"]
     runs-on: "ubuntu-20.04"
     env:
       INVOKE_NAUTOBOT_SSOT_PYTHON_VER: "${{ matrix.python-version }}"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -104,7 +104,7 @@ jobs:
       fail-fast: true
       matrix:
         python-version: ["3.7", "3.8", "3.9"]
-        nautobot-version: ["1.0.3", "1.1.4", "1.2.11", "1.3.8"]
+        nautobot-version: ["1.2.11", "1.3.8"]
     runs-on: "ubuntu-20.04"
     env:
       INVOKE_NAUTOBOT_SSOT_PYTHON_VER: "${{ matrix.python-version }}"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -103,7 +103,7 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        python-version: ["3.6", "3.7", "3.8", "3.9"]
+        python-version: ["3.7", "3.8", "3.9"]
         nautobot-version: ["1.0.3", "1.1.4"]
     runs-on: "ubuntu-20.04"
     env:
--- a/nautobot_ssot/jobs/base.py
+++ b/nautobot_ssot/jobs/base.py
@@ -1,10 +1,10 @@
 """Base Job classes for sync workers."""
 from collections import namedtuple
 from datetime import datetime
-from packaging import Version
 import traceback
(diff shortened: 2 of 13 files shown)


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "nautobot-ssot"
-version = "1.1.2"
+version = "1.2.0"
```
glennmatthews, line 3: Please don't pre-emptively bump the version in a PR to develop. We should only bump the version when we're preparing to do a release.

tasks.py
```diff
@@ -39,9 +39,9 @@ def is_truthy(arg):
 namespace.configure(
     {
         "nautobot_ssot": {
-            "nautobot_ver": "1.1.4",
+            "nautobot_ver": "1.3.3",
```
glennmatthews, line 42:
```suggestion
            "nautobot_ver": "1.3.7",
```

pyproject.toml
```diff
@@ -90,7 +91,8 @@ no-docstring-rgx="^(_|test_|Meta$)"
 disable = """,
     line-too-long,
     bad-continuation,
-    too-few-public-methods
+    too-few-public-methods,
+    arguments-differ,
```
glennmatthews, line 95: What failures is this addressing? Are you confident that they're noise and not legitimate issues?

pyproject.toml
jvanderaa: Was helping to get something out the door. I prefer it come from the PR that is putting things out personally. I do recognize this and will revert.

pyproject.toml
```diff
@@ -41,7 +40,9 @@ bandit = "*"
 pylint = "*"
 pylint-django = "*"
 pydocstyle = "*"
-flake8 = "*"
+# we need to pin flake8 because of package dependencies that cause it to downgrade and 
+# therefore cause issues with linting since older versions do not take .flake8 as config
+flake8 = "^3.9.2"
```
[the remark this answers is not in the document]
jvanderaa: Agreed, for the explicit call out on it.

tasks.py
jvanderaa: Going a step further and including today's 1.3.8 release.

nautobot_ssot/jobs/base.py
```diff
@@ -268,7 +268,7 @@ def __init__(self):
         # Default diffsync flags. You can overwrite them at any time.
         self.diffsync_flags = DiffSyncFlags.CONTINUE_ON_FAILURE | DiffSyncFlags.LOG_UNCHANGED_RECORDS
 
-    def as_form(self, data=None, files=None, initial=None):
+    def as_form(self, data=None, files=None, initial=None, approval_view=False):
         """Render this instance as a Django form for user inputs, including a "Dry run" field."""
         form = super().as_form(data=data, files=files, initial=initial)
```
glennmatthews, line 273: Tricky here as `approval_view` is only a parameter in nautobot 1.2 and later - perhaps we need something like:

```suggestion
    from nautobot import __version__ as nautobot_version
    from packaging.version import Version
    if Version(nautobot_version) < Version("1.2"):
        form = super().as_form(data=data, files=files, initial=initial)
    else:
        form = super().as_form(data=data, files=files, initial=initial, approval_view=approval_view)
```

jvanderaa:
@glennmatthews may need some thoughts on the new errors. They are not present testing locally. Anything standing out to you?

glennmatthews:
> @glennmatthews may need some thoughts on the new errors. They are not present testing locally. Anything standing out to you?

I suspect this is due to the `@override_settings(JOBS_ROOT=os.path.join(os.path.dirname(os.path.dirname(__file__)), "jobs"))` in the test case, though that doesn't explain why it wouldn't be seen locally but only in CI. I think it's resulting in the `ExampleDataSource` and `ExampleDataTarget` jobs being loaded twice (once as a plugin-provided job and once again from the overridden `JOBS_ROOT` as a "manually installed" job) which is failing since the two jobs are identical and so cannot coexist in the DB.

We might need to instead change how these test cases are set up so that they don't try to load and run a job that isn't actually a Job any more.

.github/workflows/ci.yml
```diff
@@ -103,8 +103,8 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        python-version: ["3.6", "3.7", "3.8", "3.9"]
-        nautobot-version: ["1.0.3", "1.1.4"]
+        python-version: ["3.7", "3.8", "3.9"]
+        nautobot-version: ["1.2.11", "1.3.8"]
```
chadell, line 107:
```suggestion
        nautobot-version: ["1.2.11", "stable"]
```

jvanderaa: @chadell the error is back on the stable front. Is there something else we want to cover on this before just setting it back?

glennmatthews (changes requested):
Dang. Something still isn't quite right with the Dockerfile - in the `1.2.11` case it's doing a `poetry add nautobot==1.2.11` successfully, but then the later:

```
COPY . /source
RUN poetry install --no-interaction --no-ansi
```

is overwriting the updated `poetry.lock` inside the container with the base version from Git and then reinstalling nautobot 1.3.8.

glennmatthews (approved):

PR #70, opened 2022-12-06
Enable Upstream Testing

bryanculver:
- Updated Dockerfile to be compatible with upstream testing
- Added workflow fragment to trigger testing

### DIFF
--- a/.github/workflows/upstream_testing.yml
+++ b/.github/workflows/upstream_testing.yml
@@ -1,7 +1,7 @@
 ---
 name: "Nautobot Upstream Testing"
 
-on: # yamllint disable-line rule:truthy rule:comments
+on:  # yamllint disable-line rule:truthy rule:comments
   schedule:
     - cron: "0 4 */2 * *"  # every other day at midnight
   push:  # yamllint disable rule:empty-values
--- a/.github/workflows/upstream_testing.yml
+++ b/.github/workflows/upstream_testing.yml
@@ -1,0 +1,15 @@
+---
+name: "Nautobot Upstream Testing"
+
+on: # yamllint disable-line rule:truthy rule:comments
+  schedule:
+    - cron: "0 4 */2 * *"  # every other day at midnight
+  push:  # yamllint disable rule:empty-values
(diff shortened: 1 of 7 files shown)


.github/workflows/upstream_testing.yml
```diff
@@ -0,0 +1,15 @@
+---
+name: "Nautobot Upstream Testing"
+
+on: # yamllint disable-line rule:truthy rule:comments
+  schedule:
+    - cron: "0 4 */2 * *"  # every other day at midnight
+  push:  # yamllint disable rule:empty-values
+# TODO: REMOVE PUSH LINE ABOVE
```
bryanculver, line 8:
```suggestion
  push:  # yamllint disable rule:empty-values
# TODO: REMOVE PUSH LINE ABOVE
```
Remove before merging.

chadell:
LGTM

nautobot_ssot/__init__.py
```diff
@@ -21,7 +21,7 @@ class NautobotSSOTPluginConfig(PluginConfig):
     description = "Nautobot Single Source of Truth."
     base_url = "ssot"
     required_settings = []
-    min_version = "1.0.3"
+    min_version = "1.4.0"
```
chadell, line 24: I think we should document the change on the supported version on a compatibility matrix

snaselj: Fixed by updating docs

.github/workflows/upstream_testing.yml
snaselj: Removed

nautobot_ssot/__init__.py
chadell: Thanks @snaselj ,but because this plugin is widely used, we should be more explicit with a table like this

| nautobot-plugin-ssot version | min Nautobot version |
| < 1.2.0 |  1.0.3 |
| 1.3.x | 1.4.0 |

snaselj: I see, this should be part of another PR:
https://github.com/nautobot/nautobot-plugin-ssot/pull/80

Issue #74, opened 2022-12-27
Legacy templates need to extend `generic/object_list.html`


HanlinMiao:
Was not sure about this one but can we change this https://github.com/nautobot/nautobot-plugin-ssot/blob/develop/nautobot_ssot/templates/nautobot_ssot/dashboard.html to extend `generic/object_list.html`?

PR #76, opened 2023-01-20
Changed extends 'base.html' to extends 'generic/object_detail.html' for detail templates

briddo:
Fixes: #74


nautobot_ssot/templates/nautobot_ssot/data_source_target.html
```diff
@@ -1,4 +1,4 @@
-{% extends 'base.html' %}
+{% extends 'generic/object_detail.html' %}
```
glennmatthews, line 1: In addition to this change, need to review whether any or all of the contents of `block header` are now redundant, and if they can be removed to simplify the template.

Additionally, `block content` should be updated to only override `block content_[left|right|full_width]_page` as appropriate.

.gitignore
```diff
@@ -189,6 +189,10 @@ $RECYCLE.BIN/
 *.lnk
 
 ### PyCharm ###
+
+# Jetbrains IDE configs
+.idea/
```
glennmatthews, line 194: Can presumably remove all of the individual `.idea/**/...` entries below since they're covered by this blanket entry now?

gsnider2195:
This one is going to require some not insignificant refactoring. 

- `data_source_target.html` is used in a custom `View` (not `generic.ObjectView`) that maps to multiple custom job classes, not database models. 
- `sync_header.html` is being used as a base template for the main tab and a separate job result tab on the `Sync` object view

nautobot_ssot/templates/nautobot_ssot/data_source_target.html
briddo: After reviewing the `header` block, I think this customisation is necessary.

snaselj:
Rebased on top of `develop`

snaselj:
I'm creating a new PR instead of https://github.com/nautobot/nautobot-plugin-ssot/pull/76 as is seems to me not necessary to make these changes.

bryanculver:
Replaced with #89

PR #88, opened 2023-03-03
fix: Do not fail example data source job with invalid data

snaselj:
# About

I noticed example source job is failing now when I start with empty DB, providing PR that fixes that.

## Done

- Fixed to use "active" status as a fallback for missing status in example data source.

### DIFF
--- a/nautobot_ssot/jobs/examples.py
+++ b/nautobot_ssot/jobs/examples.py
@@ -436,6 +436,10 @@ class NautobotRemote(DiffSync):
             self.job.log_debug(message=f"Loaded {region} from remote Nautobot instance")
 
         for site_entry in self._get_api_data("api/dcim/sites/"):
+            if not site_entry["status"]:
+                self.job.log_debug(message=f"Skipping site with empty status: {site_entry}")
+                continue
+
             site = self.site(
                 name=site_entry["name"],
                 slug=site_entry["slug"],
@@ -448,6 +452,10 @@ class NautobotRemote(DiffSync):
             self.job.log_debug(message=f"Loaded {site} from remote Nautobot instance")
 
         for prefix_entry in self._get_api_data("api/ipam/prefixes/"):
+            if not prefix_entry["status"]:
+                self.job.log_debug(message=f"Skipping prefix with empty status: {prefix_entry}")
+                continue
+
             prefix = self.prefix(
                 prefix=prefix_entry["prefix"],
                 description=prefix_entry["description"],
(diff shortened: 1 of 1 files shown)


nautobot_ssot/jobs/examples.py
```diff
@@ -436,6 +436,10 @@ def load(self):
             self.job.log_debug(message=f"Loaded {region} from remote Nautobot instance")
 
         for site_entry in self._get_api_data("api/dcim/sites/"):
+            if not site_entry["status"]:
```
chadell, line 439: Thanks for the suggestions, but...

Why not apply the same logic as below:

```py
region_name=site_entry["region"]["name"] if site_entry["region"] else None,
```

```py
status_slug=site_entry["status"]["value"] if ite_entry["status"] else None
```

snaselj: Because `status` field is required. I could assign some default value there, but seems to me better to ignore (and log) invalid objects, WDYT?

chadell: so, if it's not required any more in the Nautobot model, maybe we could update to make it optional?
https://github.com/nautobot/nautobot-plugin-ssot/blob/develop/nautobot_ssot/jobs/examples.py#L60

snaselj: There are these options:

- Use some default status for objects that has empty status
- Fix source data in demo.nautobot.com
- Use current solution: skip objects with empty status
- Use `save()` instead of `validated_save()`
  - not nice IMHO, but will copy all data

chadell: I will sync with @mzbroch next week to "Fix source data in demo.nautobot.com" and to understand how this can happen at all.

jdrew82: I think for this we should probably just be defaulting to Active Status. The problem is definitely in the demo data somehow allowing objects that require Status to be created without it blowing up the Job but we should have the Job handling it appropriately instead of just assuming the Status is set right.

snaselj: Updated to use "active" status as a fallback for missing status

jdrew82: Thanks @snaselj! Does that make this ready for review?

snaselj: @jdrew82 
Not yet, dry run for the source job went smoothly, but without dry run, the job get's stuck. Will check it and let you know once everything works.

Kircheneer:
@snaselj is this ready?

snaselj:
Rebased to the latest develop

snaselj:
This PR is ready for review now @Kircheneer

PR #89, opened 2023-03-08
Use generic templates

snaselj:
# About

Fixes https://github.com/nautobot/nautobot-plugin-ssot/issues/74

Replaces: https://github.com/nautobot/nautobot-plugin-ssot/pull/76 as is seems to me not necessary to make these changes.

## Done

- [x] Updated templates to extend generic templates
- [x] Refactored `DataSourceTargetView` to inherit from `ObjectView`
- [x] Removed unused `sync_run.html` template


nautobot_ssot/templates/nautobot_ssot/sync_header.html
```diff
@@ -1,4 +1,4 @@
-{% extends 'base.html' %}
+{% extends 'generic/object_detail.html' %}
```
snaselj, line 1: @gsnider2195 

Regardning the following note:

> * `sync_header.html` is being used as a base template for the main tab and a separate job result tab on the `Sync` object view
 
Do you think it would be better to replace `sync_header.html` with some custom tag and update `sync_detail.html`, `sync_jobresult.html` and `sync_logentries.html` accordingly?

chadell: @glennmatthews do you have any opinion on this?

chadell (approved):
LGTM

nautobot_ssot/templates/nautobot_ssot/sync_header.html
glennmatthews: Not sure I exactly understand what's being proposed and what its implications would be. Does it make any difference in the rendered appearance of each of these pages?

snaselj: @glennmatthews 
I can remove `sync_header.html` and implement `header` block as custom tag instead, UI from user perspective should stay the same. Not sure, whether does it worth, I was just reacting to the note from Garry.

gsnider2195: These templates have a lot of customization so they might not benefit from moving to `generic/object_detail.html`. That file provides 3 top level blocks `header`, `content` and `javascript`. `sync_header.html` overloads the `header` block, all 3 of the tabs overload the `content` block and one of them overloads the `javascript` block. That effectively makes moving to `object_detail.html` a no-op. Was this change driven by the need to simply move away from extending `base.html` or to leverage the benefits of extending `object_detail.html`? If it's the latter, I don't see any benefit changing the base template for `sync_header.html`.

jdrew82: If we've determined that using object_detail is a no-op, can we revert that bit for this PR so we can get the rest merged in?

snaselj: Reverted

Issue #92, opened 2023-03-21
MySQL: OperationalError: (1038, 'Out of sort memory, consider increasing server sort buffer size') 


grelleum:
### Environment
* Python version: 3.8.13
* Nautobot version:  1.5.13
* nautobot-ssot version:  1.2.0

### Expected Behavior
When clicking `Plugins > Single Source of Truth > Dashboard` in the Nautobot UI, I expect to see the SSoT Dashboard.

### Observed Behavior
Page fails to load.  Debug mode presents a traceback with the exception:
```
OperationalError: (1038, 'Out of sort memory, consider increasing server sort buffer size')
```

### Steps to Reproduce
1. Deploy Nautobot using the MySQL backend database.
2. Run several large SSoT jobs.
3. Navigate to `Plugins > Single Source of Truth > Dashboard` in the Nautobot UI.

grelleum:
Same error occurs when navigating to `Plugins > Single Source of Truth > Logs` in the Nautobot UI.

Kircheneer:
@grelleum can this be closed now that #93 is merged?

grelleum:
yes, thank you

PR #93, opened 2023-03-21
fix-92: Defer loading JSONField 'diff' in queryset

grelleum:
Addresses issue #92 

This PR resolve two possible `out of sort memory` errors when running against a MySQL databse.
1. Navigating to `Plugins > Single Source of Truth > Dashboard`
2. Navigating to `Plugins > Single Source of Truth > Logs`

Additionally, the page load time is significantly faster.

### DIFF
--- a/nautobot_ssot/forms.py
+++ b/nautobot_ssot/forms.py
@@ -24,7 +24,7 @@ class SyncLogEntryFilterForm(BootstrapMixin, forms.ModelForm):
     """Form for filtering SyncLogEntry records."""
 
     q = forms.CharField(required=False, label="Search")
-    sync = forms.ModelChoiceField(queryset=Sync.objects.all(), required=False)
+    sync = forms.ModelChoiceField(queryset=Sync.objects.defer("diff").all(), required=False)
     action = forms.ChoiceField(choices=add_blank_choice(SyncLogEntryActionChoices), required=False)
     status = forms.ChoiceField(choices=add_blank_choice(SyncLogEntryStatusChoices), required=False)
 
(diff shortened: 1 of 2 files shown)


chadell (approved):
LGTM

nautobot_ssot/forms.py
```diff
@@ -24,7 +24,7 @@ class SyncLogEntryFilterForm(BootstrapMixin, forms.ModelForm):
     """Form for filtering SyncLogEntry records."""
 
     q = forms.CharField(required=False, label="Search")
-    sync = forms.ModelChoiceField(queryset=Sync.objects.all(), required=False)
+    sync = forms.ModelChoiceField(queryset=Sync.objects.defer("diff").all(), required=False)
```
Kircheneer, line 27: Do we need to `defer` or can we outright use `only` instead? From the docs
> only use defer() when you cannot, at queryset load time, determine if you will need the extra fields or not
This would mean that we _might_ need the diff for the filter form. Is this the case?

nautobot_ssot/views.py
```diff
@@ -24,7 +24,7 @@
 class DashboardView(ObjectListView):
     """Dashboard / overview of SSoT."""
 
-    queryset = Sync.objects.all()
+    queryset = Sync.objects.defer("diff").all()
```
Kircheneer, line 27: Same comment as above should apply here if it applies there.

nautobot_ssot/forms.py
Kircheneer: May not work, because `values` returns a dictionary, but I think we should evaluate this.

nautobot_ssot/views.py
grelleum: same as above, `.values()` provides tuples, not objects.

nautobot_ssot/forms.py
grelleum: specifically, when trying `.values()` I get the error:
```
'dict' object has no attribute 'get_source_url'
```

Issue #91, opened 2023-03-20
Prometheus metrics


Kircheneer:
### Environment
* Nautobot version:  1.5.13 (to get `metrics.py` support)
* nautobot-plugin-ssot version:  latest

### Proposed Functionality

Expose Prometheus metrics for this app. Examples:

- `nautobot_ssot_duration_seconds{phase="loading-remote", job="example-data-source"} 10.0`
- `nautobot_ssot_duration_seconds{phase="diff", job="example-data-source"} 2.0`
- `nautobot_ssot_memory_profiling_?{?) ?` - this one is not as well thought out but I think it would be interesting to add a metric for this

Further metrics could be useful, these are just examples I was able to think of.

### Use Case

Allow users of the platform to generate time series from their Nautobot instance. In the end this could for example lead to creation of NOC dashboards or alerting based on this data.

jdrew82:
I'd say it'd be useful to have all of the results of the diff be provided as metrics. That and possibly an average over time could be useful.

jdrew82:
It could also be useful to have the metrics around how long it takes to load data from each SoR and also the time to calculate the diff or complete the sync. This would allow an end user to see when a sync is taking longer than it normally does and why much more easily.

PR #95, opened 2023-03-24
Add Metrics

jdrew82:
This PR is focused on adding Prometheus metrics for the SSoT framework to address #91. This should add the following metrics:

`nautobot_ssot_duration_microseconds{phase="source_load_time", job=<job_slug>}`
`nautobot_ssot_sync_totals{sync_type="total_syncs"}`
`nautobot_ssot_sync_memory_usage{phase="source_load_memory_final"}`


nautobot_ssot/metrics.py
```diff
@@ -22,25 +22,37 @@ def metric_ssot_jobs():
...
+                ssot_job_durations.add_metric(
+                    labels=["target_load_time", job.slug],
+                    value=(last_job_sync.target_load_time.seconds * 1000000)
+                    + last_job_sync.target_load_time.microseconds,
+                )
+
+            if last_job_sync.diff_time:
+                ssot_job_durations.add_metric(
+                    labels=["diff_time", job.slug],
+                    value=(last_job_sync.diff_time.seconds * 1000000) + last_job_sync.diff_time.microseconds,
```
chadell, line 42: I thought the `microseconds` was including the `seconds`... so, this means that you need adding both metrics to get the final one?

jdrew82: I thought so too and that's why I originally used just the microseconds attribute. When I got to testing though the conversion didn't match up until I included the seconds and multiplier. It became really evident with the diff_time:

```
>>> sync.diff_time
datetime.timedelta(seconds=1, microseconds=69464)
```

jdrew82:
Here's a screenshot I got of the metrics after a few tests:
Screenshot 2023-03-27 at 8 21 57 AM

jdrew82:
I've also added in a metric for the latest metrics for CRUD operations from the diff summary. As the summary appeared to not be saved anywhere and just logged I've gone ahead and added an attribute to the Sync object to save it. I did see that the list appears to calculate these numbers when displaying the Sync history list. I'm happy to update the method these metrics are obtained to be similar if someone can help explain how they're derived to me if we don't want to have the `summary` attribute. Personally, I think it makes sense to save that small dictionary of data instead of re-calculating it repeatedly but I understand not wanting to add unnecessary attributes to the Sync class.

nautobot_ssot/metrics.py
```diff
@@ -0,0 +1,164 @@
...
+        value=Sync.objects.filter(job_result__status="completed").count(),
+    )
+
+    sync_gauge.add_metric(labels=["failed_syncs"], value=Sync.objects.filter(job_result__status="failed").count())
+
+    sync_gauge.add_metric(labels=["errored_syncs"], value=Sync.objects.filter(job_result__status="errored").count())
+
+    sync_gauge.add_metric(labels=["pending_syncs"], value=Sync.objects.filter(job_result__status="pending").count())
+
+    sync_gauge.add_metric(labels=["running_syncs"], value=Sync.objects.filter(job_result__status="running").count())
```
chadell, line 81:
```suggestion
    for status_type in ("failed", "errored", "pending", "running"):
        sync_gauge.add_metric(labels=[f"{status_type}_syncs"], value=Sync.objects.filter(job_result__status=status_type).count())
```

chadell (approved):
LGTM. I don't see any problem with the `summary`field

nautobot_ssot/metrics.py
jdrew82: Good catch! I've updated to this pattern and also included the completed gauge too.

nautobot_ssot/metrics.py
```diff
@@ -0,0 +1,156 @@
...
+def metric_ssot_jobs():
+    """Extracts duration of latest SSoT Job run.
+
+    Yields:
+        GaugeMetricFamily: Prometheus Metrics
+    """
+    ssot_job_durations = GaugeMetricFamily(
+        "nautobot_ssot_duration_microseconds",
+        "Nautobot SSoT Job Phase Duration in microseconds",
+        labels=["phase", "job"],
```
Kircheneer, line 19: The unit should be seconds (see [here](https://prometheus.io/docs/practices/naming/#base-units) for a list of base units).

nautobot_ssot/metrics.py
```diff
@@ -0,0 +1,156 @@
...
+    """Calculate total number of Syncs.
+
+    Yields:
+        GaugeMetricFamily: Prometheus Metrics
+    """
+    sync_gauge = GaugeMetricFamily("nautobot_ssot_sync_totals", "Nautobot SSoT Sync Totals", labels=["sync_type"])
+
+    sync_gauge.add_metric(labels=["total_syncs"], value=Sync.objects.all().count())
+
+    for status_type in ("completed", "failed", "errored", "pending", "running"):
```
Kircheneer, line 70: Should this be iterating over a choices variable from core?

nautobot_ssot/metrics.py
```diff
@@ -0,0 +1,156 @@
...
+    )
+
+    for job in Job.objects.filter(slug__icontains="ssot"):
+        last_job_sync = Sync.objects.filter(job_result__job_model_id=job.id).last()
+        if last_job_sync and last_job_sync.summary:
+            for operation, value in last_job_sync.summary.items():
+                sync_ops.add_metric(
+                    labels=[job.slug, operation],
+                    value=value,
+                )
```
Kircheneer, line 95: This should output an empty metric if there are no jobs present.

Kircheneer: Empty as in sample set to 0, no labels.

nautobot_ssot/metrics.py
```diff
@@ -0,0 +1,156 @@
...
+
+
+def metric_sync_operations():
+    """Extracts the diff summary operations from each Job's last Sync.
+
+    Yields:
+        GuageMetricFamily: Prometheus Metrics
+    """
+    sync_ops = GaugeMetricFamily(
+        "nautobot_ssot_sync_operations", "Nautobot SSoT operations by Job", labels=["job", "operation"]
```
Kircheneer, line 85:
```suggestion
        "nautobot_ssot_sync_total", "Nautobot SSoT operations by Job", labels=["job", "operation"]
```

nautobot_ssot/metrics.py
```diff
@@ -0,0 +1,156 @@
...
+    Yields:
+        GaugeMetricFamily: Prometheus Metrics
+    """
+    memory_gauge = GaugeMetricFamily(
+        "nautobot_ssot_sync_memory_usage", "Nautobot SSoT Sync Memory Usage", labels=["phase"]
+    )
+
+    last_sync = Sync.objects.filter(source_load_memory_final__isnull=False).last()
+
+    if last_sync:
```
Kircheneer, line 112: This should have an `else` clause that supplies an empty metrics (value 0 and no labels).

nautobot_ssot/metrics.py
```diff
@@ -0,0 +1,156 @@
...
+
+
+def metric_memory_usage():
+    """Extracts memory usage for latest SSoT Sync where memory profiling was used.
+
+    Yields:
+        GaugeMetricFamily: Prometheus Metrics
+    """
+    memory_gauge = GaugeMetricFamily(
+        "nautobot_ssot_sync_memory_usage", "Nautobot SSoT Sync Memory Usage", labels=["phase"]
```
Kircheneer, line 107: Should this have a job label?

Kircheneer:
```suggestion
        "nautobot_ssot_sync_memory_usage_bytes", "Nautobot SSoT Sync Memory Usage", labels=["phase"]
```

pyproject.toml
```diff
@@ -27,10 +27,11 @@ packages = [
 
 [tool.poetry.dependencies]
 python = "^3.7"
-nautobot = "^1.4.0"
+nautobot = "^1.5.13"
 diffsync = "^1.6.0"
 Markdown = "!=3.3.5"
 packaging = "^21.3"
+nautobot-capacity-metrics = "^2.0.0"
```
Kircheneer, line 34: What is this needed for?

Kircheneer: If this is not needed, we need to pull in `prometheus_client` as an explicit dependency.

Kircheneer (changes requested):

nautobot_ssot/metrics.py
```diff
@@ -0,0 +1,156 @@
...
+    Yields:
+        GaugeMetricFamily: Prometheus Metrics
+    """
+    ssot_job_durations = GaugeMetricFamily(
+        "nautobot_ssot_duration_microseconds",
+        "Nautobot SSoT Job Phase Duration in microseconds",
+        labels=["phase", "job"],
+    )
+
+    for job in Job.objects.filter(slug__icontains="ssot"):
```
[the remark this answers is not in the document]
jdrew82: As long as the name of the project that the Job comes from follows our standard of having SSoT in the name then it should be fine. From what I saw the slug is essentially the full python path to the Job so should contain `ssot` as long as they are following the standard. Unfortunately, I couldn't think of another method of determining SSoT-specific Jobs. If you can think of a method I'm open to changing this.

nautobot_ssot/metrics.py
jdrew82: This isn't a choice option that a user can select. It's automatically set from the Job results.

nautobot_ssot/metrics.py
```diff
@@ -0,0 +1,156 @@
...
+    yield ssot_job_durations
+
+
+def metric_syncs():
+    """Calculate total number of Syncs.
+
+    Yields:
+        GaugeMetricFamily: Prometheus Metrics
+    """
+    sync_gauge = GaugeMetricFamily("nautobot_ssot_sync_totals", "Nautobot SSoT Sync Totals", labels=["sync_type"])
```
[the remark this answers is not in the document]
jdrew82: Why would it be total as opposed to totals? It's a `GaugeMetricFamily` which implies a collection of things so should be plural. There are multiple metrics included in it. Why are you thinking just `total`?

pyproject.toml
jdrew82: This was added because I thought we need the capacity metrics App to enable metric capability in Nautobot. I believe that has the prometheus_client dependency.

nautobot_ssot/metrics.py
jdrew82: Again, why `total` here as opposed to `totals`? You've got multiple items here. It's not just one operation total.

nautobot_ssot/metrics.py
jdrew82: Done.

nautobot_ssot/metrics.py
jdrew82: We can do that, sure. I wasn't sure how applicable that'd be. My understanding is that you'd most likely use memory profiling for troubleshooting purposes so it wouldn't be enabled on Jobs all the time and you wouldn't necessarily care about it for all Jobs.

nautobot_ssot/metrics.py
Kircheneer: The sample is a float value, so if you pass in a float you don't lose any precision.

nautobot_ssot/metrics.py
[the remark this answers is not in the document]
Kircheneer: Sorry, I should have been more clear here - this is a standard from the Prometheus side ([ref](https://prometheus.io/docs/practices/naming/)).

nautobot_ssot/metrics.py
Kircheneer: My suggestion was to loop over this maybe: https://github.com/nautobot/nautobot/blob/develop/nautobot/extras/choices.py#L189

Then if something changes in Nautobot core the metric is automatically appended to.

nautobot_ssot/metrics.py
Kircheneer: Same reasoning as above.

nautobot_ssot/metrics.py
Kircheneer: I think this is still interesting to know. If I am building a dashboard about it, the data is a lot more comparable between different jobs if I have this label. Otherwise I can only graph the different jobs in a single graph.

pyproject.toml
Kircheneer: Yup - I think it makes more sense pulling in `prometheus_client` directly, although even the `nautobot` dependency should take care of that so it might not make sense to be more specific then necessary here. Still I think we should remove `nautobot-capacity-metrics` as I don't think it provides any value here.

pyproject.toml
```diff
@@ -27,10 +27,11 @@ packages = [
 
 [tool.poetry.dependencies]
 python = "^3.7"
-nautobot = "^1.4.0"
+nautobot = "^1.5.13"
```
Kircheneer, line 30: Nice learning from another similar MR - we don't have to change this. If the version is smaller than 1.5.13 it just doesn't render any metrics - if the version is bigger then or equal to it renders them. This should maybe go in the docs instead?

nautobot_ssot/metrics.py
jdrew82: Strange. When I was testing this and used the seconds attribute it appeared to be returned as an int not a float. Is there a specific way to get it as a float? That's why I went to microseconds originally.

Kircheneer: Oh sorry, misunderstanding - my point was to get the value in microseconds and then divide it by a 1000 and pass that to the metric.

nautobot_ssot/metrics.py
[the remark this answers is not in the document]
Kircheneer: This should work:

```
[sync["job_result__job_model__name"] for sync in Sync.objects.values("job_result__job_model__name").order_by().distinct()]
```

nautobot_ssot/metrics.py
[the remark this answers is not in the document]
jdrew82: Got it. I wasn't aware there were naming standards. Updated accordingly.

nautobot_ssot/metrics.py
jdrew82: Corrected.

nautobot_ssot/metrics.py
jdrew82: Refactored to include Job as a label.

pyproject.toml
jdrew82: Changed to prometheus client as requested.

jdrew82: So I'm not sure this is going to work. The capacity-metrics App is needed for the `register_metric_func` function that's used in the `__init__.py` file. I can't find that function in prometheus-client. Should I revert to capacity-metrics?

Kircheneer: Check [this](https://docs.nautobot.com/projects/core/en/stable/plugins/development/#prometheus-metrics) - specifically the part on `metrics.py` - you don't need register_metric_func anymore.

nautobot_ssot/metrics.py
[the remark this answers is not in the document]
jdrew82: This only works if there has already been a Sync attempted for the Job so a Sync object is created. My method allows for all SSoT Jobs regardless of whether they've been ran yet to show metrics.

pyproject.toml
jdrew82: Ah, I was going off [this](https://github.com/nautobot/nautobot-plugin-capacity-metrics#option-2---registry-for-third-party-plugins) in capacity-metrics that said they had to be registered. I'll remove that then.

nautobot_ssot/metrics.py
jdrew82: Done.

nautobot_ssot/metrics.py
jdrew82: This has been changed back to seconds but including microseconds to get a more precise measurement.

nautobot_ssot/metrics.py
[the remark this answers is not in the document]
jdrew82: @Kircheneer: Do you still want to use Sync objects or is looking for `ssot` in the Job slug acceptable? I want to get this PR merged in if possible.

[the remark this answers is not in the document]
Kircheneer: Does it make sense to report phase duration for jobs that haven't yet run (i.e. that don't have `Sync` object instances)? I don't think so, and therefore I think it should be alright if we went by the `Sync` object route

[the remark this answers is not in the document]
jdrew82: Typically, I'd agree with you, but I could see a new user being confused as to why they don't see the metric when they first enable them and it might be confusing that they are required to complete a Sync before the metrics work.

Kircheneer (approved):

jdrew82:
Closing to follow the documented release pattern.

PR #104, opened 2023-04-26
Add docs on performance optimizations.

### LINKED ISSUE #98: Document how to efficiently load data from Nautobot
### Environment
* Nautobot version:  all
* nautobot-ssot version: all

### Proposed Functionality

The documentation shows how to use features such as `Queryset.select_related` to use efficient DB queries in order to increase the performance of SSoT jobs.

### Use Case

Big SSoT jobs can take quite long. I would like them to be faster

### Example

```
for region in Region.objects.all():
    for site in Site.objects.filter(region=region):
        for location in Location.objects.filter(location_type__name="Floor", site=site):
            for sub_location in Location.objects.filter(location_type__name="Room", parent=location):
                pass
```

is quite slow, because it has a a query per `sub_location`.

```
for sub_location in Location.objects.filter(location_type__name="Room").select_related(
    "parent",
    "parent__site",
    "parent__site__region"
):
    pass
```

is just one big query with `JOIN`.

Kircheneer:
Closes #98


jdrew82 (approved):
LGTM!

docs/user/developing_jobs.md
```diff
@@ -42,21 +44,115 @@ In brief, the following general steps can be followed:
...
-The built-in implementation of `sync_data` is composed by 4 steps:
+## Optimizing for Execution Time
+
+When syncing large amounts of data, job execution time may start to become an issue. Fortunately, there are a number of ways you can go about optimizing your jobs performance.
+
+### Optimizing Database Queries
+
+As an SSoT job typically has lots of database interaction for loading, creating, updating, and deleting objects, this is a common source of performance issues.
+
+The following is an example of an efficient `load` function that can be greatly improved:
```
itdependsnetworks, line 55:
```suggestion
The following is an example of an inefficient `load` function that can be greatly improved:
```

docs/user/developing_jobs.md
```diff
@@ -42,21 +44,115 @@ In brief, the following general steps can be followed:
...
+- diffsync [logging configuration](https://diffsync.readthedocs.io/en/latest/api/diffsync.logging.html?highlight=logging)
+- `print` calls
+- other, external frameworks you are using
+
+## Analyzing Job Performance
+
+In general there are two different metrics to optimize for when developing SSoT jobs:
+
+- CPU time (maps directly to total execution time)
+- Memory usage
```
itdependsnetworks, line 136: Should IO be on this list?

Kircheneer: I don't think IO is something you optimize for, stdout IO is something you can do less of to optimize for CPU time, isn't it?

itdependsnetworks: One of the issues I was told was basically the same premise you described in the example but for external systems. e.g. Make an API call to each device, interface, asset information in ServiceNow, vs making a paginated call to Devices, Interfaces, and Assets, (so only 3 paginated calls vs N).

Kircheneer: Ohh I understand. Yes that is 100% worth calling out, too. Thanks!

docs/user/developing_jobs.md
```diff
@@ -42,21 +44,115 @@ In brief, the following general steps can be followed:
...
 6. Install your Job via any of the supported Nautobot methods (installation into the `JOBS_ROOT` directory, inclusion in a Git repository, or packaging as part of a plugin) and it should automatically become available!
 
-## Analyze Job performance
-
-The built-in implementation of `sync_data` is composed by 4 steps:
+## Optimizing for Execution Time
+
+When syncing large amounts of data, job execution time may start to become an issue. Fortunately, there are a number of ways you can go about optimizing your jobs performance.
+
+### Optimizing Database Queries
```
chadell, line 51:
```suggestion
### Optimizing Nautobot Database Queries
```

docs/user/developing_jobs.md
```diff
@@ -42,21 +44,115 @@ In brief, the following general steps can be followed:
...
+
+As an SSoT job typically has lots of database interaction for loading, creating, updating, and deleting objects, this is a common source of performance issues.
```
chadell, line 53:
```suggestion
As an SSoT job typically has lots of Nautobot database interaction (i.e. Nautobot is always the source or destination target) for loading, creating, updating, and deleting objects, this is a common source of performance issues.
```

chadell (commented):
I don't see any mention of the atomic transaction topic, and how to leverage `post_run`.
Also, I think it could be interesting to introduce the trade-off of using bulk operations

docs/user/developing_jobs.md
chadell: Also, it would be good to provide some ideas to improve, with the tradeoffs (i.e. overloading the target endpoint)

Kircheneer:
> I don't see any mention of the atomic transaction topic, and how to leverage `post_run`. Also, I think it could be interesting to introduce the trade-off of using bulk operations

I wanted to (for now) stick to things for which I concretely know that they don't pose any negatives, with `post_run` losing the atomic transaction and resulting in abnormal job behavior (you have to manually reload the job result, it shows as completed too early) and bulk operations not running custom model validation or even `save` methods. 

These optimizations could be documented at a later point in time IMO.

docs/user/developing_jobs.md
```diff
@@ -42,21 +44,115 @@ In brief, the following general steps can be followed:
...
+            child_region.add_child(site)
+        location, location_created = self.get_or_instantiate("location", ids={"name": location.name})
+        if location_created:
+            self.add(location)
+            site.add_child(location)
+```
+
+As an additional bonus, this way the code has less levels of indentation.
+
+The essence of this is that you should make liberal use of `select_related` to join together the database tables you need into a single, big query rather than a bunch of small queries.
```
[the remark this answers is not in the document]
Kircheneer: Addressed, thanks!

chadell (commented):
Add `.idea` to gitignore

chadell:
> 

Yes, they can be added in another PR, but they are in the scope of #98 issue

docs/user/developing_jobs.md
chadell: You could use threading or asyncio to parallelize IO, but this could overload the other endpoint.

docs/user/developing_jobs.md
```diff
@@ -42,21 +44,125 @@ In brief, the following general steps can be followed:
...
+
+- diffsync [logging configuration](https://diffsync.readthedocs.io/en/latest/api/diffsync.logging.html?highlight=logging)
+- `print` calls
+- other, external frameworks you are using
+
+### Minimizing external IO
+
+In most if not all cases, the side of an SSoT job that interacts with the non-Nautobot system will be accessed through some form of IO as for example HTTP requests via the network. Depending on the amount of requests, request/response size and the latency to the remote system this can take a lot of time. Care should be taken when crafting the IO interaction, using bulk endpoints instead of querying each individual record on the remote system where possible.
+
```
itdependsnetworks, line 139:
```suggestion

As an example, you could take the following proposal:

* Collect Sites
    * For each Site, collect all Devices
        * For each Device collect the Interface information
        * For each Device collect the VLAN information

However, this would be preferred:

* Collect all Sites
* Collect all Devices
* Collect all Interfaces
* Collect all VLANs

```

Kircheneer: I worked this in, thanks!

Kircheneer:
I have responded to most the requests now. Open points:

## Using `bulk_create`

I don't feel good recommending this in the documentation as is right now because of its severe trade-offs:
- No change log
- No model validation (depending on this model this ranges from okay to very bad - take a look at the `Cable` model for example: https://github.com/nautobot/nautobot/blob/develop/nautobot/dcim/models/cables.py#L149 - leaving it up to the user to re-implement these validations feels like a very bad idea and could lead to their instance being rendered unusable)
- No custom save methods (again, this depends on the model - its really bad for device https://github.com/nautobot/nautobot/blob/develop/nautobot/dcim/models/devices.py#L814 - not using this custom `save` method would lead to no device components being created

I argue that we need explicit Nautobot core support for this to be reasonable.

## Parallelizing the load functions

I think this needs to be supported from SSoT itself rather than a recommendation in the docs for people to build it themselves. To me it only spells trouble when users would have to convert their custom methods of parallelization to a method that eventually will be supported out of this app.

@chadell thoughts?

docs/user/developing_jobs.md
Kircheneer: https://github.com/nautobot/nautobot-plugin-ssot/pull/104#issuecomment-1531198525

chadell:
> on't feel good recommending this in the documentation as is right now because of its severe trade-offs:
> 
> * No change log
> * No model validation (depending on this model this ranges from okay to very bad - take a look at the `Cable` model for example: https://github.com/nautobot/nautobot/blob/develop/nautobot/dcim/models/cables.py#L149 - leaving it up to the user to re-implement these validations feels like a very bad idea and could lead to their instance being rendered unusable)
> * No custom save methods (again, this depends on the model - its really bad for device https://github.com/nautobot/nautobot/blob/develop/nautobot/dcim/models/devices.py#L814 - not using this custom `save` method would lead to no device components being created
> 
> I argue that we need explicit Nautobot core support for this to be reasonable.

1. If we do not document it (even for discouraging), someone else will ask the question again. I prefer to be explicit about it, so everyone can make their own decisions. I mean, documenting != recommending.
2. Parallelization of load jobs should be out of the scope, as it will be tackled after Nautobot 2.0 release. The other question is parallelizing a single job with concurrent API operations. Again, I am not saying we recommend doing it, but explaining potential pros/cons will make everyone understand it.

docs/user/developing_jobs.md
Kircheneer: Addressed

Kircheneer:
All feedback addressed

chadell (approved):
LGTM

itdependsnetworks (approved):

PR #106, opened 2023-05-02
Release v1.3.0

jdrew82:
This PR is to prep for the v1.3.0 release.


.cookiecutter.json
```diff
@@ -0,0 +1,25 @@
...
+    "plugin_slug": "nautobot-ssot",
+    "project_slug": "nautobot-plugin-ssot",
+    "repo_url": "https://github.com/nautobot/nautobot-plugin-ssot",
+    "base_url": "ssot",
+    "min_nautobot_version": "1.4.0",
+    "max_nautobot_version": "1.9999",
+    "nautobot_version": "1.5.1",
+    "camel_name": "NautobotSsot",
+    "project_short_description": "Nautobot Single Source of Truth",
+    "version": "1.2.0",
```
glennmatthews, line 18: Should this be 1.3.0? I'm not entirely clear on the purpose of this file - do we need to keep this field in sync with the app version number?

.cookiecutter.json
```diff
@@ -0,0 +1,25 @@
+{
+  "cookiecutter": {
+    "codeowner_github_usernames": "@glennmatthews @jathanism @lampwins @chadell",
```
glennmatthews, line 3: Should this be `@nautobot/plugin-ssot` instead?

docs/admin/compatibility_matrix.md
```diff
@@ -0,0 +1,10 @@
+# Compatibility Matrix
+
+While that last supported version will not be strictly enforced--via the max_version setting, any issues with an updated Nautobot supported version in a minor release, will require a bug to be raised and a fix in Nautobot core to address, with no fixes expected in this plugin. This allows the Nautobot Single Source of Truth App the ability to quickly take advantage of the latest features.
```
glennmatthews, line 3: I think this is boilerplate text, but I honestly have no idea what this paragraph is telling me.

Kircheneer (approved):

glennmatthews (approved):

.cookiecutter.json
jdrew82: This file is used to re-generate the cookie I believe. I'm not entirely sure it's required to be updated as we update the repo. Thoughts @Kircheneer and @chadell?

docs/admin/compatibility_matrix.md
jdrew82: This was already in the repo before I created this PR.

.cookiecutter.json
Kircheneer: Paging @snaselj

.cookiecutter.json
jdrew82: Again, this was generated when the cookie was first created and I'm not sure it needs to be updated as we update the project?

PR #125, opened 2023-05-25
Update to DiffSync 1.8.0 minimum

jdrew82:
This PR just updates the framework to use DiffSync 1.8.0 minimum along with updating all the other project dependencies.


jdrew82:
> Does the move to Poetry 1.5 and removal of the `category` keys from the lockfile pose any backward-compatibility issues with older versions of Poetry?

I would expect that to only affect the dev environment as the lockfile isn't used otherwise to my understanding.

Kircheneer:
@jdrew82 can this be closed?

Kircheneer:
Closing, feel free to re-open if this is deemed necessary.

PR #138, opened 2023-06-12
Add Arista CloudVision integration

### LINKED ISSUE #132: Migrate Arista CloudVision SSoT Plugin to This Repo
### Environment
* Nautobot version:  
* nautobot-plugin-chatops version:  

### Proposed Functionality

After #128 and #115, merge https://github.com/nautobot/nautobot-plugin-ssot-arista-cloudvision into this repo as an `integration`.

### Use Case

Make installation, discovery, maintaining easier.

snaselj:
# Closes #132  

- Adds Arista CloudVision integration from nautobot-plugin-ssot-arista-cloudvision.

## TODO

- [ ] Test integration.


snaselj:
Rebased on top of develop

chadell:
@qduk @jdrew82 could you review it?

pyproject.toml
```diff
@@ -62,6 +65,8 @@ mkdocs-version-annotations = "1.0.0"
 mkdocstrings = "0.20"
 mkdocstrings-python = "0.8.3"
 requests-mock = "^1.10.0"
+nautobot-device-lifecycle-mgmt = "^1.0.2"
```
jdrew82, line 68: This should optional too. The support for the DLC App isn't required for this integration. It's simply there to be used if found.

jdrew82 (changes requested):
I don't see you updating the Dockerfile for the dev environment here. We should have that added so the dev environment has the App for coding against.

pyproject.toml
snaselj: `nautobot-device-lifecycle-mgmt` is dev dependency, installed in dev container only. Up to the user, whether installed in production.

Would you like it to be an optional extra dependency instead?

snaselj:
> I don't see you updating the Dockerfile for the dev environment here. We should have that added so the dev environment has the App for coding against.

`[all]` and dev dependencies are installed to dev image from poetry / pip, `poetry.lock` is updated.

Is there anything else you are missing in the Dockerfile?

pyproject.toml
jdrew82: I see that the original repo has it as a dev dependency so I understand why you placed it there. I misread it as a regular dependency. That's probably fine. I believe other Apps have it as a regular dependency but marked optional. I'm not sure what's the better pattern there. As for the Dockerfile, you need to update one of the install lines like [here](https://github.com/nautobot/nautobot-plugin-ssot-arista-cloudvision/blob/c4b3f3c1fc0e54a207afe5b15fad2410b48e8941/development/Dockerfile#L67C1-L68C1) as if you don't specify the install it will actually remove it from the installed modules.

jdrew82:
> > I don't see you updating the Dockerfile for the dev environment here. We should have that added so the dev environment has the App for coding against.
> 
> `[all]` and dev dependencies are installed to dev image from poetry / pip, `poetry.lock` is updated.
> 
> Is there anything else you are missing in the Dockerfile?

I addressed this in an above comment.

pyproject.toml
snaselj: I made `nautobot-device-lifecycle-mgmt` an optional dependency, as it is IMHO a better approach and updated the Dockerfile to take into account all dependencies when creating the list of `dev` dependencies.

However `nautobot-device-lifecycle-mgmt` was installed in the `dev` image before this change as well.

Could you pls. test it, whether the image contain all necessary dependencies?

jdrew82 (approved):
LGTM!

pyproject.toml
Kircheneer: @jdrew82 can you test this so we can move forward?https://github.com/networktocode/circuit-maintenance-parser

jdrew82: @Kircheneer I no longer have access to a CVP instance unfortunately so we can't test this unless we get our CVP instance up and working again.

jdrew82:
@snaselj can you rebase this on develop so the conflicts go away?

pyproject.toml
jdrew82: Confirmed that the DLC App is available in the dev environment so we should be g2g on this one.

jdrew82:
This should be ready to merge once the rebase happens or conflicts are fixed.

snaselj:
> This should be ready to merge once the rebase happens or conflicts are fixed.

Conflicts are resolved by merging to the latest `develop`.

Issue #38, opened 2022-04-13
Make Logs Available Via API


jdrew82:
### Environment
* Nautobot version:  1.2.8
* nautobot-ssot version:  1.1.0

### Proposed Functionality
As an end-user, it'd be very helpful to have the SSoT logs available via the API for programmatic searching. This is especially important as the number of SoRs imported and the number of objects imported increases.

### Use Case
As an end-user, when I have an inquiry about why a large number of devices weren't imported it can be rather arduous to review the logs via the GUI. Being able to programmatically search them via the API would make this task much easier.

Kircheneer:
#97 is another way to solve your specific use case, although I don't think this invalidates API access to the logs entirely.

PR #143, opened 2023-06-21
Add API Endpoints

jdrew82:
This PR is for addressing #38 to make the Sync and SyncLogEntry objects available via API endpoints.

### DIFF
--- a/nautobot_ssot/api/views.py
+++ b/nautobot_ssot/api/views.py
@@ -1,0 +1,32 @@
+"""Views for SSoT APIs."""
+from rest_framework.routers import APIRootView
+
+from nautobot.extras.api.views import NautobotModelViewSet
+
+from nautobot_ssot.api import serializers
+from nautobot_ssot import filters
+from nautobot_ssot import models
+
+
+class SsotRootView(APIRootView):
+    """SSoT API root view."""
+
+    def get_view_name(self):
+        """API for SSoT root view boilerplate."""
+        return "Single Source of Truth"
+
+
+class SyncViewSet(NautobotModelViewSet):  # pylint: disable=too-many-ancestors
+    """API viewset for interacting with Sync objects."""
+
+    queryset = models.Sync.objects.all()
+    serializer_class = serializers.SyncListSerializer
+    filterset_class = filters.SyncFilterSet
+
+
+class SyncLogEntryViewSet(NautobotModelViewSet):  # pylint: disable=too-many-ancestors
+    """API viewset for interacting with SyncLogEntry objects."""
+
+    queryset = models.SyncLogEntry.objects.all()
+    serializer_class = serializers.SyncLogEntrySerializer
+    filterset_class = filters.SyncLogEntryFilterSet
--- a/nautobot_ssot/api/views.py
+++ b/nautobot_ssot/api/views.py
@@ -8,7 +8,7 @@ from nautobot_ssot import filters
 from nautobot_ssot import models
 
 
-class SsotRootView(APIRootView):
+class SSOTRootView(APIRootView):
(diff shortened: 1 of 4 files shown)


nautobot_ssot/api/views.py
```diff
@@ -0,0 +1,32 @@
...
+from rest_framework.routers import APIRootView
+
+from nautobot.extras.api.views import NautobotModelViewSet
+
+from nautobot_ssot.api import serializers
+from nautobot_ssot import filters
+from nautobot_ssot import models
+
+
+class SsotRootView(APIRootView):
```
Kircheneer, line 11:
```suggestion
class SSotRootView(APIRootView):
```

Nitpick

Kircheneer (changes requested):

nautobot_ssot/api/views.py
jdrew82: I thought PEP8 says it should be camelcase?

nautobot_ssot/tests/test_api.py
```diff
@@ -1,15 +1,151 @@
...
 
 from nautobot.users.models import Token
+from nautobot.utilities.testing.api import APIViewTestCases
+
+from nautobot_ssot import models
 
 User = get_user_model()
 
 
+class SyncAPITest(APIViewTestCases.APIViewTestCase):  # pylint: disable=too-many-ancestors
```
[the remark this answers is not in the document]
jdrew82: I think some of the tests are certainly valid. I just wasn't sure how to go about actually testing them. I can work on building them out though. This was meant as an initial stab at adding the API endpoints to satisfy the referenced issue and to also get ready for the Nautobot 2.0 React UI.

nautobot_ssot/api/views.py
glennmatthews: It's OK to have multiple consecutive capitals - I'd vote for `SSOTRootView` myself as I personally prefer acronyms to stay capitalized rather than being treated as TitleCase words. (`nautobot.ipam.models.IPAddress`, not `IpAddress`)

jdrew82: Alright, works for me.

nautobot_ssot/tests/test_api.py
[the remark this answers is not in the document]
Kircheneer: I believe that we should get the unit tests in along with the feature, otherwise its unclear when/if the unit tests will ever be enabled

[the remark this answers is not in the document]
jdrew82: Agreed, I just need time to figure out the tests. I'm not familiar with how the API unit tests work so still working through that.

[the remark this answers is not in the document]
Kircheneer: Here you should use separate subclasses of `APIViewTestCases` as this is a read-only API. That should cut down on the necessary `@skip` annotations.

nautobot_ssot/tests/test_api.py
```diff
@@ -1,15 +1,151 @@
...
+    @skip("Not implemented")
+    def test_list_objects_filtered(self):
+        pass
+
+    @skip("Not implemented")
+    def test_list_objects_unknown_filter_no_strict_filtering(self):
+        pass
+
+
+class SyncLogEntryAPITest(APIViewTestCases.APIViewTestCase):  # pylint: disable=too-many-ancestors
```
Kircheneer, line 78: Same thing about the different base class here since this is also read-only

jdrew82:
Putting this on hold until I have time to work on the tests after the 2.0 release.

Kircheneer:
Closing until we pick it back up.

Issue #45, opened 2022-05-06
TypeError: Object of type date is not JSON serializable


progala:
### Environment
* Python version:  Example: 3.7.12
* Nautobot version:  Example: 1.2.8
* nautobot-ssot version:  1.1.0

### Expected Behavior

SSoT synchronization completing successfully for the object with a field of type `datetime.date`.

### Observed Behavior

An exception is raised, see details below, when the SSoT tries to save the sync diff dict.

```
An exception occurred: TypeError: Object of type date is not JSON serializable

Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/nautobot_ssot/jobs/base.py", line 326, in run
    self.sync_data()
  File "/source/ssot/jobs.py", line 220, in sync_data
    self.sync.save()
  File "/usr/local/lib/python3.7/site-packages/django/db/models/base.py", line 754, in save
    force_update=force_update, update_fields=update_fields)
  File "/usr/local/lib/python3.7/site-packages/django/db/models/base.py", line 792, in save_base
    force_update, using, update_fields,
  File "/usr/local/lib/python3.7/site-packages/django/db/models/base.py", line 873, in _save_table
    forced_update)
  File "/usr/local/lib/python3.7/site-packages/django/db/models/base.py", line 926, in _do_update
    return filtered._update(values) > 0
  File "/usr/local/lib/python3.7/site-packages/django/db/models/query.py", line 803, in _update
    return query.get_compiler(self.db).execute_sql(CURSOR)
  File "/usr/local/lib/python3.7/site-packages/django/db/models/sql/compiler.py", line 1522, in execute_sql
    cursor = super().execute_sql(result_type)
  File "/usr/local/lib/python3.7/site-packages/django/db/models/sql/compiler.py", line 1143, in execute_sql
    sql, params = self.as_sql()
  File "/usr/local/lib/python3.7/site-packages/django/db/models/sql/compiler.py", line 1488, in as_sql
    val = field.get_db_prep_save(val, connection=self.connection)
  File "/usr/local/lib/python3.7/site-packages/django/db/models/fields/__init__.py", line 823, in get_db_prep_save
    return self.get_db_prep_value(value, connection=connection, prepared=False)
  File "/usr/local/lib/python3.7/site-packages/django/db/models/fields/__init__.py", line 818, in get_db_prep_value
    value = self.get_prep_value(value)
[...]

PR #146, opened 2023-07-03
Use DjangoJSONEncoder for diff field

jmcgill298:
The DjangoJSONEncoder provide JSON encoding for additional object types, such as datetime and uuid objects, two types that are often used in the Sync.diff field.


chadell (approved):
LGTM

chadell:
fixes #45

PR #150, opened 2023-07-13
Release v2.0.0b2

jdrew82:
This PR is to prepare for cutting release for v2.0.0-beta2. 

- Updated version to specify beta2 to match Nautobot.
- Added release notes to summarize changes.

### DIFF
--- a/docs/admin/release_notes/version_2.0.md
+++ b/docs/admin/release_notes/version_2.0.md
@@ -1,7 +1,7 @@
 
 # v2.0 Release Notes
 
-## v2.0.0-beta.2 - 2023-07-13
+## v2.0.0-beta.1 - 2023-07-13
 
 ### Added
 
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "nautobot-ssot"
-version = "2.0.0"
+version = "2.0.0-beta.2"
```
Kircheneer, line 3: Shouldn't this be beta 1 of SSoT?

docs/admin/release_notes/version_2.0.md
```diff
@@ -0,0 +1,21 @@
+
+# v2.0 Release Notes
+
+## v2.0.0-beta.2 - 2023-07-13
```
Kircheneer, line 4: Same here

Kircheneer (changes requested):
Just the version number

pyproject.toml
jdrew82: I was doing this to match Nautobot but I'll change it to beta 1. Will also add a compatibility matrix to documentation so it's clear what version is supported for which Nautobot version.

Kircheneer (approved):
LGTM!

PR #69, opened 2022-11-23
Add Initial Mixins

itdependsnetworks:

PR #153, opened 2023-07-28
Implementing base models/adapters for Nautobot side functionality.

Kircheneer:
Similar approach and borrowing lots of ideas from #69 but with automatic inference.


nautobot_ssot/tests/test_base_models.py
```diff
@@ -0,0 +1,107 @@
...
+    description: Optional[str] = None
+    group__name: Optional[str] = None
+    tags: List[TagDict] = []
+
+
+class TestAdapter(BaseAdapter):
+    """An adapter for testing the `BaseAdapter` base class."""
+
+    top_level = ("tenant",)
+    tenant = NautobotTenant
```
Kircheneer, line 34: This is the interesting part, as this is all that's needed to have an adapter that does all CRUD operations.

chadell: it looks clean :)

lampwins:
Is the goal here to get to dynamic pydantic models based on the underlying django models? Curious if you have explored https://github.com/jordaneremieff/djantic/?

itdependsnetworks:
The inference of data types would be great but the inference of the full model would not. This does not conflict with anything that @lampwins indicated, but just want to be clear of the intention as auto discovery of the whole model would be problematic given how complex Nautobot's data model is.

docs/user/developing_jobs.md
```diff
@@ -6,43 +6,96 @@ Familiarity with [DiffSync](https://diffsync.readthedocs.io/en/latest/) and with
...
+    _model = Tenant
+    _modelname = "tenant"
+    _identifiers = ("name",)
+    _attributes = ("description",)
+
+    name: str
+    description: str
+```
+
+As you can see when looking at the source code of the `nautobot.tenancy.models.Tenant` model, the fields on the model (`name` and `description`) match the names of the fields on that model one-to-one. This enables the base class `NautobotModel` to dynamically load, create, update and delete your tenants without you needing to implement this functionality yourself.
```
itdependsnetworks, line 38:
```suggestion
As you can see when looking at the source code of the `nautobot.tenancy.models.Tenant` model, the fields on the model (`name` and `description`) match the names of the fields on that model exactly. This enables the base class `NautobotModel` to dynamically load, create, update and delete your tenants without you needing to implement this functionality yourself.
```

itdependsnetworks: one-to-one is a term used in django orm and databases, my mind goes to that definition.

docs/user/developing_jobs.md
```diff
@@ -6,43 +6,96 @@ Familiarity with [DiffSync](https://diffsync.readthedocs.io/en/latest/) and with
...
+
+For a full how-to on modelling different kinds of fields, check the corresponding [page](modelling.md) in the docs.
```
itdependsnetworks, line 40:
```suggestion
The above example shows the simplest field type (an attribute on the model), however, to build a production implementation you will need to understand how to identify different variants of fields by following the [modeling docs](modeling.md).
```

itdependsnetworks: english spelling is modeling, will need to adjust the filename to match.

docs/user/developing_jobs.md
```diff
@@ -6,43 +6,96 @@ Familiarity with [DiffSync](https://diffsync.readthedocs.io/en/latest/) and with
...
+from nautobot_ssot.contrib import BaseAdapter
+
+from your_ssot_app.models import DiffSyncDevice, DiffSyncPrefix, DiffSyncIPAddress
+
+class NautobotAdapter(BaseAdapter):
+    top_level = ("device", "prefix")
+
+    device = DiffSyncDevice
+    prefix = DiffSyncPrefix
+    ip_address = DiffSyncIPAddress  # Not in top_level because it's a child of the prefix model
```
itdependsnetworks, line 56:
```suggestion
    ip_address = DiffSyncIPAddress  # Not in the `top_level` tuple, since it's a child of the prefix model
```

docs/user/developing_jobs.md
```diff
@@ -6,43 +6,96 @@ Familiarity with [DiffSync](https://diffsync.readthedocs.io/en/latest/) and with
...
+```
+
+The `load` function is already implemented on this adapter and will automatically and recursively traverse any children relationships for you.
```
itdependsnetworks, line 59:
```suggestion
The `load` function is already implemented on this adapter and will automatically and recursively traverse any children relationships for you, provided the models are defined correctly.
```

docs/user/developing_jobs.md
```diff
@@ -6,43 +6,96 @@ Familiarity with [DiffSync](https://diffsync.readthedocs.io/en/latest/) and with
...
-     ```python
-     def load_source_adapter(self):
-         """Method to instantiate and load the SOURCE adapter into `self.source_adapter`."""
-         self.source_adapter = NautobotRemote(url=self.kwargs["source_url"], token=self.kwargs["source_token"], job=self)
-         self.source_adapter.load()
-     ```
+### Implementing `create`, `update` and `delete`
 
-   - `self.load_target_adapter`: This is mandatory to be implemented. As an example:
+In case you are synchronizing data _to_ and not _from_ Nautobot, you can entirely skip this step. The `nautobot_ssot.contrib.NautobotModel` class takes care of automatically.
```
itdependsnetworks, line 63:
```suggestion
If you are synchronizing data _to_ Nautobot and not _from_ Nautobot, you can entirely skip this step. The `nautobot_ssot.contrib.NautobotModel` class provides this functionality automatically.
```

docs/user/developing_jobs.md
```diff
@@ -6,43 +6,96 @@ Familiarity with [DiffSync](https://diffsync.readthedocs.io/en/latest/) and with
...
-     def load_target_adapter(self):
-        """Method to instantiate and load the TARGET adapter into `self.target_adapter`."""
-        self.target_adapter = NautobotLocal(job=self)
-        self.target_adapter.load()
-     ```
+If you need to perform the `create`, `update` and `delete` operations on the remote system however, it will be up to you to implement those on your model.
 
-   - `self.calculate_diff`: This method is implemented by default, using the output from load_adapter methods.
+!!! note
+    You still want your models to adhere to the [modelling guide](modelling.md), because doing so gives you the auto-generated `load` function for the diffsync adapter on the Nautobot side.
```
itdependsnetworks, line 68:
```suggestion
    You still want your models to adhere to the [modeling guide](modeling.md), since it provides you the auto-generated `load` function for the diffsync adapter on the Nautobot side.
```

docs/user/developing_jobs.md
```diff
@@ -6,43 +6,96 @@ Familiarity with [DiffSync](https://diffsync.readthedocs.io/en/latest/) and with
...
+
+### Loading from the Remote System
+
+Regardless of which direction you are synchronizing data in, you need to write the `load` method for the remote adapter yourself. You can find many examples of how this can be done in the `nautobot_ssot.integrations` module, which contains pre-existing integrations with remote systems.
+
+!!! note
+    As the features in `nautobot_ssot.contrib` are still very new, these existing integrations won't necessarily use them.
```
itdependsnetworks, line 75:
```suggestion
    As the features in `nautobot_ssot.contrib` are still very new, most existing integrations do not use them.
```

docs/user/developing_jobs.md
```diff
@@ -6,43 +6,96 @@ Familiarity with [DiffSync](https://diffsync.readthedocs.io/en/latest/) and with
...
+
+### Putting it Together in a Job
+
+Develop a Job class, derived from either the `nautobot_ssot.jobs.base.DataSource` or `DataTarget` classes provided by this plugin, and implement the adapters to populate the self.source_adapter and self.target_adapter that are used by the built-in implementation of sync_data. This sync_data method is an opinionated way of running the process including some performance data, more in next section, but you could overwrite it completely or any of the key hooks that it calls:
```
itdependsnetworks, line 79: I have trouble following this one, mind taking another crack at it?

docs/user/modelling.md
```diff
@@ -0,0 +1,128 @@
+# Modelling
```
itdependsnetworks, line 1:
```suggestion
# Modeling
```

docs/user/modelling.md
```diff
@@ -0,0 +1,128 @@
...
+
+This page describes how to model various kinds of fields on a `nautobot_ssot.contrib.NautobotModel` subclass.
+
+## Normal Fields
+
+For normal, non-relationship, non-custom fields on a model, all you need to do is to ensure that field name on your SSoT model class matches that of the Nautobot model class. To ensure this is the case, you can check out the model either by reading the corresponding source code on [GitHub](https://github.com/nautobot/nautobot) or through the [Nautobot Shell](https://docs.nautobot.com/projects/core/en/stable/administration/nautobot-shell/) - in there you can use commands like `dir(Tenant)` to get an overview of available fields for the different models.
```
itdependsnetworks, line 7: This is one of the reasons I thought the graphql idea was good. Obviously not for this PR, but in general, we need a better story for newcomers to understand how to build this out.

docs/user/modelling.md
```diff
@@ -0,0 +1,128 @@
...
+
+!!! note
+    If you want to sync data into a model provided by a [Nautobot App](https://docs.nautobot.com/projects/core/en/stable/plugins/), you need to navigate to its respective source code repository. Furthermore, you may need to manually import the model if you're using the Nautobot Shell approach.
+
+## Custom Fields
+
+For [custom fields](https://docs.nautobot.com/projects/core/en/stable/models/extras/customfield/), you will need to use a `nautobot_ssot.contrib.CustomFieldAnnotation`. Given a custom field called "Test" on the circuit provider model, this is how you could map that custom field into the diffsync model field `test_custom_field`:
```
itdependsnetworks, line 14:
```suggestion
For [custom fields](https://docs.nautobot.com/projects/core/en/stable/models/extras/customfield/), you will need to use the `nautobot_ssot.contrib.CustomFieldAnnotation` class. Given a custom field called "Test" on the circuit provider model, this is how you could map that custom field into the diffsync model field `test_custom_field`:
```

docs/user/modelling.md
```diff
@@ -0,0 +1,128 @@
...
+
+```python
+from typing_extensions import Annotated  # Python<3.9 - you need to install this module
+# from typing import Annotated  # Python>=3.9
+from nautobot.circuits.models import Provider
+from nautobot_ssot.contrib import NautobotModel, CustomFieldAnnotation
+class ProviderModel(NautobotModel):
```
itdependsnetworks, line 21:
```suggestion
from nautobot_ssot.contrib import NautobotModel, CustomFieldAnnotation

class ProviderModel(NautobotModel):
```

docs/user/modelling.md
```diff
@@ -0,0 +1,128 @@
...
+    _modelname = "provider"
+    _identifiers = ("name",)
+    _attributes = ("custom_field",)
+
+    name: str
+    test_custom_field: Annotated[str, CustomFieldAnnotation(name="Test")]
+```
+
+!!! note
+    This is necessary because custom field names may include spaces, which are un-representable in Python object field names unless you resort to `setattr` shenanigans, which would get us into trouble here.
```
itdependsnetworks, line 32:
```suggestion
    Defining a `CustomFieldAnnotation` variable is necessary since custom field names may include spaces, which are un-representable in Python object field names.
```

docs/user/modelling.md
```diff
@@ -0,0 +1,128 @@
...
+
+## Many-to-one Relationships
+
+For many-to-one relationships (i.e. [foreign keys](https://docs.djangoproject.com/en/3.2/topics/db/examples/many_to_one/) or [generic foreign keys](https://docs.djangoproject.com/en/3.2/ref/contrib/contenttypes/#generic-relations)) a slightly different approach is employed. We need to add a field on our model for each field that we need in order to uniquely identify our related object behind the many-to-one relationship. We can do this using a [familiar synta](https://docs.djangoproject.com/en/3.2/topics/db/queries/#lookups-that-span-relationships) of double underscore separated paths. Assuming we want to synchronize prefixes and associate them with locations, we may be faced with the problems that locations aren't uniquely identified by name alone, but rather need the location type as well, we can address this as follows.
```
itdependsnetworks, line 36:
```suggestion
For many-to-one relationships (i.e. [foreign keys](https://docs.djangoproject.com/en/3.2/topics/db/examples/many_to_one/) or [generic foreign keys](https://docs.djangoproject.com/en/3.2/ref/contrib/contenttypes/#generic-relations)) a slightly different approach is employed. We need to add a field on our model for each field that we need in order to uniquely identify our related object behind the many-to-one relationship. We can do this using a [familiar syntax](https://docs.djangoproject.com/en/4.2/topics/db/queries/#lookups-that-span-relationships) of double underscore separated paths. Assuming we want to synchronize `prefixes` and associate them with `locations`, we may be faced with the problems that locations aren't uniquely identified by name alone, but rather need the location type as well, we can address this as follows.
```

docs/user/modelling.md
```diff
@@ -0,0 +1,128 @@
...
+
+```python
+from nautobot.ipam.models import Prefix
+from nautobot_ssot.contrib import NautobotModel
+class PrefixModel(NautobotModel):
```
itdependsnetworks, line 41:
```suggestion
from nautobot_ssot.contrib import NautobotModel

class PrefixModel(NautobotModel):
```

docs/user/modelling.md
```diff
@@ -0,0 +1,128 @@
...
+    _attributes = ("location__name", "location__location_type__name")
+
+    network: str
+    prefix_length: int
+
+    location__name: str
+    location__location_type__name: str
+```
+
+Now, on model `create` or `update`, the SSoT framework will dynamically pull in the location with the specified name and location type name, uniquely identifying the location and populating the foreign key.
```
itdependsnetworks, line 53: it's not immediately clear to me to what `to add a field on our model` is exactly constructed.

docs/user/modelling.md
```diff
@@ -0,0 +1,128 @@
...
+
+### Special Case: Generic Foreign Key
+
+In the case of a [generic foreign key](https://docs.djangoproject.com/en/3.2/topics/db/examples/many_to_many/), we are faced with a problem. Normally, the [content type](https://docs.djangoproject.com/en/3.2/ref/contrib/contenttypes/) of a relationship field can be inferred from the model class. In the case of generic foreign keys however, this is not the case. In the example of the [IP address](https://docs.nautobot.com/projects/core/en/stable/models/ipam/ipaddress/?h=ip+address) model in Nautobot, the `assigned_object` field can point to either a device's or a VM's interface. We address this the following way:
```
itdependsnetworks, line 57:
```suggestion
In the case of a [generic foreign key](https://docs.djangoproject.com/en/3.2/topics/db/examples/many_to_many/), we are faced with a problem. Normally, the [content type](https://docs.djangoproject.com/en/3.2/ref/contrib/contenttypes/) of a relationship field can be inferred from the model class. In the case of generic foreign key**s** however, this is not the case. In the example of the [IP address](https://docs.nautobot.com/projects/core/en/stable/models/ipam/ipaddress/?h=ip+address) model in Nautobot, the `assigned_object` field can point to either a device's or a VM's interface. We address this the following way:
```

docs/user/modelling.md
```diff
@@ -0,0 +1,128 @@
...
+
+```python
+from nautobot.ipam.models import IPAddress
+from nautobot_ssot.contrib import NautobotModel
+class NautobotIPAddress(NautobotModel):
```
itdependsnetworks, line 62:
```suggestion
from nautobot_ssot.contrib import NautobotModel

class NautobotIPAddress(NautobotModel):
```

docs/user/modelling.md
```diff
@@ -0,0 +1,128 @@
...
+    prefix_length: int
+    status__name: str
+    assigned_object__app_label: str
+    assigned_object__model: str
+    assigned_object__device__name: str
+    assigned_object__name: str
+```
+
+!!! warning
+    A limitation of this approach is that we are locked into a single kind of content type unless the field names for all content types match. In this specific case, the VM interface model does not have a `device` field. This is a known issue.
```
itdependsnetworks, line 87: we are locked into a single kind of content type unless the field names -> we are locked into a single kind of content type [per something] unless the field names

docs/user/modelling.md
```diff
@@ -0,0 +1,128 @@
...
+
+## One-to-many and many-to-many Relationships
+
+For "to-many" relationships such as (generic) foreign keys traversed backwards or [many-to-many relationships](https://docs.djangoproject.com/en/3.2/topics/db/examples/many_to_many/), we need to employ a different mechanism. First, we need to again identify which fields of the related object we are interested in for queries to it. In this case, our example will be using an interface model for which we also want to sync associated IP addresses. For our scenario, lets assume that our IP addresses can be uniquely identified through the `host` and `prefix_length` fields:
```
itdependsnetworks, line 91:
```suggestion
For "to-many" relationships such as (generic) foreign keys traversed backwards or [many-to-many relationships](https://docs.djangoproject.com/en/3.2/topics/db/examples/many_to_many/), we need to employ a different mechanism. Again we start by identifying which fields of the related object we are interested in for queries **to** the model. In this case, our example will be using an `Interface` model for which we also want to sync the associated IP addresses. For our scenario, lets assume that our IP addresses can be uniquely identified through the `host` and `prefix_length` fields:
```

docs/user/modelling.md
```diff
@@ -0,0 +1,128 @@
...
+```python
+from typing_extensions import TypedDict  # Python<3.9 - you need to install this module
+# from typing import TypedDict  # Python>=3.9
+
+class IPAddressDict(TypedDict):
+    host: str
+    prefix_length: int
+```
+
+Having defined this, we can now again define our diffsync model:
```
itdependsnetworks, line 102:
```suggestion
Having defined this, we can now define our diffsync model:
```

docs/user/modelling.md
```diff
@@ -0,0 +1,128 @@
...
+
+```python
+from typing import List  # use the builtin 'list' from Python 3.9 on
+from nautobot.dcim.models import Interface
+from nautobot_ssot.contrib import NautobotModel
+class NautobotInterface(NautobotModel):
```
itdependsnetworks, line 108:
```suggestion
from nautobot_ssot.contrib import NautobotModel

class NautobotInterface(NautobotModel):
```

docs/user/modelling.md
```diff
@@ -0,0 +1,128 @@
...
+
+    name: str
+    device__name: str
+    ip_addresses: List[IPAddressDict] = []
+```
+
+!!! note
+    In this example we can also see a case where a foreign key (name `device`, through `device__name`) constitutes part of the `_identfiers` field - this is a common pattern as model relationships define their uniqueness across multiple models in Nautobot.
+
+Through us defining this, Nautobot will now be able to dynamically load IP addresses related to our interfaces from Nautobot, and also set the related objects on the relationship in the `create` and `update` methods.
```
itdependsnetworks, line 125:
```suggestion
Through us defining the model, Nautobot will now be able to dynamically load IP addresses related to our interfaces from Nautobot, and also set the related objects on the relationship in the `create` and `update` methods. 
```

docs/user/modelling.md
```diff
@@ -0,0 +1,128 @@
+# Modelling
+
+This page describes how to model various kinds of fields on a `nautobot_ssot.contrib.NautobotModel` subclass.
+
```
itdependsnetworks, line 4:
```suggestion
## Quick Reference

| Type | action | notes |
| ----- | ------ | ----- |
| Normal fields | standard attribute | N/A |
| Custom Fields | f"{field_nam}_custom_field" | need to use custom class, etc. etc. |

```

itdependsnetworks: ^^ not ready in it's current form, but you should be able to see where I am going with this.

mkdocs.yml
```diff
@@ -102,6 +102,7 @@ nav:
       - Getting Started: "user/app_getting_started.md"
       - Using the App: "user/app_use_cases.md"
       - Developing Jobs: "user/developing_jobs.md"
+      - Modelling: "user/modelling.md"
```
itdependsnetworks, line 105:
```suggestion
      - Modelling: "user/modeling.md"
```

docs/user/developing_jobs.md
```diff
@@ -6,43 +6,96 @@ Familiarity with [DiffSync](https://diffsync.readthedocs.io/en/latest/) and with
...
+5. Putting it together into a job
 
-   - Define your models based on the data you want to sync. For example, if you are syncing specific attributes (e.g. tags) of a device, but not the devices themselves, your parent model
-     should be the tags. This approach prevents unnecessary `create`, `update`, and `delete` calls for models that are not sync'd.
-   - For each model class, implement the `create`, `update`, and `delete` DiffSyncModel APIs for writing data to the Data Target system.
+### Defining the Models
 
-2. Define a `DiffSync` adapter class for loading initial data from Nautobot and constructing instances of each `DiffSyncModel` class to represent that data.
-3. Define a `DiffSync` adapter class for loading initial data from the Data Source or Data Target system and constructing instances of the `DiffSyncModel` classes to represent that data.
+The models use diffsync, which in turn uses [pydantic](https://docs.pydantic.dev/latest/) for its data modelling. Nautobot SSoT comes with a set of classes out of the box that implement a lot of functionality for you automatically using the classes from `nautobot_ssot.contrib`, provided you adhere to a set of rules.
```
itdependsnetworks, line 19:
```suggestion
The models use diffsync, which in turn uses [pydantic](https://docs.pydantic.dev/latest/) for its data modeling. Nautobot SSoT comes with a set of classes out of the box that implement a lot of functionality for you automatically using the classes from `nautobot_ssot.contrib`, provided you adhere to a set of rules.
```

development/creds.example.env
```diff
@@ -32,15 +32,15 @@ NAUTOBOT_ARISTACV_CVP_TOKEN="changeme"
...
-NAUTOBOT_APIC_USERNAME_NTC=admin
-NAUTOBOT_APIC_PASSWORD_NTC=super_secret_password
-NAUTOBOT_APIC_VERIFY_NTC=False
-NAUTOBOT_APIC_SITE_NTC="NTC ACI"
-NAUTOBOT_APIC_BASE_URI_DEVNET=https://sandboxapicdc.cisco.com
-NAUTOBOT_APIC_USERNAME_DEVNET=admin
-NAUTOBOT_APIC_PASSWORD_DEVNET=super_secret_password
-NAUTOBOT_APIC_VERIFY_DEVNET=False
-NAUTOBOT_APIC_SITE_DEVNET="DevNet Sandbox"
+# NAUTOBOT_APIC_BASE_URI_NTC=https://aci.cloud.networktocode.com
```
itdependsnetworks, line 35: ???

docs/user/developing_jobs.md
```diff
@@ -6,43 +6,96 @@ Familiarity with [DiffSync](https://diffsync.readthedocs.io/en/latest/) and with
...
 ## Quickstart
 
-In brief, the following general steps can be followed:
+As a general overview, here are the different things to be done to have a functioning SSoT job:
 
-1. Define one or more `DiffSyncModel` data model class(es) representing the common data record(s) to be synchronized between the two systems.
+1. Define the base models that define the data contract
+2. Implement the Nautobot adapter automatically through `nautobot_ssot.contrib.BaseAdapter`
+3. If Nautobot is the source of the job implement `create`, `update`, `delete` methods for the models in the remote system
+4. Implement the remote adapters `load` method, responsible for loading the data from the remote system into the adapter
```
chadell, line 14:
```suggestion
4. Implement the adapters `load` method for the non Nautobot targets, responsible for loading the data from the remote system into the adapter 
```

docs/user/developing_jobs.md
chadell: why if I want to reimplement the functionality to add some customization? maybe a note to show that this is possible?

docs/user/developing_jobs.md
```diff
@@ -6,43 +6,96 @@ Familiarity with [DiffSync](https://diffsync.readthedocs.io/en/latest/) and with
...
+
+```python
+def load_target_adapter(self):
+   """Method to instantiate and load the TARGET adapter into `self.target_adapter`."""
+   self.target_adapter = NautobotLocal(job=self)
+   self.target_adapter.load()
+```
 
-   - `self.execute_sync`: This method is implemented by default, using the output from load_adapter methods. Only executed if it's not a `dry-run` execution.
+The methods `calculate_diff` and `execute_sync` are both implemented be default, using the data that is loaded into the adapters through the respective methods. Note that `execute_sync` will _only_ execute when dry-run is set to false.
```
chadell, line 95:
```suggestion
The methods `calculate_diff` and `execute_sync` are both implemented by default, using the data that is loaded into the adapters through the respective methods. Note that `execute_sync` will _only_ execute when dry-run is set to false.
```

docs/user/developing_jobs.md
```diff
@@ -6,43 +6,96 @@ Familiarity with [DiffSync](https://diffsync.readthedocs.io/en/latest/) and with
...
 
-5. Optionally, on your Job class, also implement the `lookup_object`, `data_mappings`, and/or `config_information` APIs (to provide more information to the end user about the details of this Job), as well as the various metadata properties on your Job's `Meta` inner class. Refer to the example Jobs provided in this plugin for examples and further details.
-6. Install your Job via any of the supported Nautobot methods (installation into the `JOBS_ROOT` directory, inclusion in a Git repository, or packaging as part of a plugin) and it should automatically become available!
+Optionally, on your Job class, also implement the `lookup_object`, `data_mapping`, and/or `config_information` APIs (to provide more information to the end user about the details of this Job), as well as the various metadata properties on your Job's Meta inner class. Refer to the example Jobs provided in this plugin for examples and further details.
```
chadell, line 97: maybe a link to docs about these methods?

nautobot_ssot/contrib.py
```diff
@@ -0,0 +1,347 @@
...
+
+        This then maps the model field 'is_global' to the custom field 'Is Global'.
+    """
+
+    name: str
+
+
+class BaseAdapter(DiffSync):
+    """
+    Adapter for loading from Nautobot.
```
chadell, line 41: from Nautobot?

chadell: I would propose `NautobotAdapter`...

pyproject.toml
```diff
@@ -38,7 +38,7 @@ nautobot = "^1.4.0"
 nautobot-device-lifecycle-mgmt = { version = "^1.0.2", optional = true }
 oauthlib = { version = ">=3.1.0", optional = true }
 packaging = ">=21.3, <24"
-prometheus-client = "~0.14.1"
+prometheus-client = ">=0.14.1"
```
chadell, line 41: not sure why this is relevant in this PR

development/creds.example.env
Kircheneer: If I leave these in, it automatically creates `Site` objects which mess up the example SSoT job.

docs/user/modelling.md
Kircheneer: Yeah I wasn't happy with this, it requires somewhat intricate knowledge. Would be cool to have a data model published somewhere in a very easy to understand format

docs/user/modelling.md
Kircheneer: I don't understand, why is it important to highlight the "s" here?

itdependsnetworks: First you write:

> In the case of a [generic foreign key]

Then you write:

> In the case of generic foreign keys however

After reading that 5 times I am still not sure what two things are being compared. It seemed logical that you were comparing `generic foreign key` to `generic foreign keys` based on the paragraph structure ... but perhaps that is wrong?? So emphasis is my best guess at what is being stated as different.

itdependsnetworks: I would say a common theme is to be explicit, often by removing pronouns but in this case a bit different. It's very complex topic and I am not always understanding "what exactly" is being referred to.

development/creds.example.env
itdependsnetworks: ok, would just check in with whomever owns aci

docs/user/modelling.md
Kircheneer: Great idea! Put something in there.

docs/user/modeling.md
```diff
@@ -0,0 +1,144 @@
...
+    assigned_object__device__name: str
+    assigned_object__name: str
+```
+
+!!! warning
+    A limitation of this approach is that we are locked into a single kind of content type unless the field names for all content types match. In this specific case, the VM interface model does not have a `device` field. This is a known issue.
+
+## One-to-many and many-to-many Relationships
+
+For "to-many" relationships such as (generic) foreign keys traversed backwards or [many-to-many relationships](https://docs.djangoproject.com/en/3.2/topics/db/examples/many_to_many/), we need to employ a different mechanism. Again we start by identifying which fields of the related object we are interested in for queries **to** the model. In this case, our example will be using an `Interface` model for which we also want to sync the associated IP addresses. For our scenario, lets assume that our IP addresses can be uniquely identified through the `host` and `prefix_length` fields:
```
itdependsnetworks, line 106:
```suggestion
For "*-to-many" relationships such as (generic) foreign keys traversed backwards or [many-to-many relationships](https://docs.djangoproject.com/en/3.2/topics/db/examples/many_to_many/), we need to employ a different mechanism. Again we start by identifying which fields of the related object we are interested in for queries **to** the model. In this case, our example will be using an `Interface` model for which we also want to sync the associated IP addresses. For our scenario, lets assume that our IP addresses can be uniquely identified through the `host` and `prefix_length` fields:
```

development/creds.example.env
chadell: isn't this a problem in a scenario when you want to use both?

nautobot_ssot/contrib.py
```diff
@@ -0,0 +1,354 @@
...
+            is_global: Annotated[bool, CustomFieldAnnotation(name="Is Global")
+        ```
+
+        This then maps the model field 'is_global' to the custom field 'Is Global'.
+    """
+
+    name: str
+
+
+class BaseAdapter(DiffSync):
```
chadell, line 43: I would rename it to something more explicit such as `NautobotAdapter`

nautobot_ssot/contrib.py
```diff
@@ -0,0 +1,354 @@
...
+    @staticmethod
+    def _get_parameter_names(diffsync_model):
+        """Ignore the differences between identifiers and attributes, because at this point they don't matter to us."""
+        return list(diffsync_model._identifiers) + list(diffsync_model._attributes)  # pylint: disable=protected-access
+
+    def _load_models(self, diffsync_model):
+        """Given a diffsync model class, load a list of models from the database and return them."""
+        parameter_names = self._get_parameter_names(diffsync_model)
+
+        # Here we identify any foreign keys so that we can load them directly in the first query.
```
chadell, line 59:
```suggestion
        # Here we identify any foreign keys (identified by "__") so that we can load them directly in the first query.
```

nautobot_ssot/contrib.py
```diff
@@ -0,0 +1,354 @@
...
+
+    This adapter is able to infer how to load data from Nautobot based on how the models attached to it are defined.
+    """
+
+    @staticmethod
+    def _get_parameter_names(diffsync_model):
+        """Ignore the differences between identifiers and attributes, because at this point they don't matter to us."""
+        return list(diffsync_model._identifiers) + list(diffsync_model._attributes)  # pylint: disable=protected-access
+
+    def _load_models(self, diffsync_model):
```
chadell, line 55: it seems to me that the name is misleading. You aren't loading "models" but "objects" (`self._load_single_object`)

nautobot_ssot/contrib.py
```diff
@@ -0,0 +1,354 @@
...
+                continue
+
+            # Handle custom fields. See CustomFieldAnnotation docstring for more details.
+            is_custom_field = False
+            metadata_for_this_field = getattr(type_hints[parameter_name], "__metadata__", [])
+            for metadata in metadata_for_this_field:
+                if isinstance(metadata, CustomFieldAnnotation):
+                    parameters[parameter_name] = database_object.cf[metadata.name]
+                    is_custom_field = True
+                    continue
```
chadell, line 86: why not?

```suggestion
                    break
```

chadell, line 86: why not?

```suggestion
                    break
```

nautobot_ssot/contrib.py
```diff
@@ -0,0 +1,354 @@
...
+        # If the lookup doesn't point anywhere, check whether it is using the convention for generic foreign keys.
+        except AttributeError:
+            if lookups[-1] in ["app_label", "model"]:
+                return getattr(ContentType.objects.get_for_model(related_object), lookups[-1])
+        return None
+
+
+class NautobotModel(DiffSyncModel):
+    """
+    Base model for any diffsync models interfacing with Nautobot.
```
chadell, line 198: Jsut to make explicit that this is not covering Nautobot via API

```suggestion
    Base model for any diffsync models interfacing with Nautobot via ORM.
```

nautobot_ssot/contrib.py
```diff
@@ -0,0 +1,354 @@
...
+        try:
+            diffsync_model = getattr(self, model_name)
+        except AttributeError as error:
+            raise AttributeError(
+                f"Please define {model_name} to be the diffsync model on this adapter as a class level attribute."
+            ) from error
+        return diffsync_model
+
+    @staticmethod
+    def _handle_to_many_relationship(database_object, diffsync_model, parameter_name):
```
chadell, line 145: you are also using "handle" in the `NautobotModel` methods. Maybe here we could use something like "get"? (because you are getting things)

nautobot_ssot/contrib.py
```diff
@@ -0,0 +1,354 @@
...
+        try:
+            obj.validated_save()
+        except ValidationError as error:
+            raise ValidationError(f"Parameters: {parameters}") from error
+
+        # Set many-to-many fields after saving
+        cls._handle_many_to_many_fields(many_to_many_fields, obj)
+
+    @classmethod
+    def _handle_many_to_many_fields(cls, many_to_many_fields, obj):
```
chadell, line 304: similiar to my other comment, here, instead of "handle" may be "set" would be more explicit?

development/creds.example.env
Kircheneer: I think that we need to re-think how the ACI integration does this, its a little too invasive for my taste. I don't think me changing the development environment here should really impact anybody, so I will create an issue from this discussion.

docs/user/developing_jobs.md
```diff
@@ -6,43 +6,96 @@ Familiarity with [DiffSync](https://diffsync.readthedocs.io/en/latest/) and with
 
 ## Quickstart
 
-In brief, the following general steps can be followed:
+As a general overview, here are the different things to be done to have a functioning SSoT job:
 
-1. Define one or more `DiffSyncModel` data model class(es) representing the common data record(s) to be synchronized between the two systems.
+1. Define the base models that define the data contract
+2. Implement the Nautobot adapter automatically through `nautobot_ssot.contrib.BaseAdapter`
+3. If Nautobot is the source of the job implement `create`, `update`, `delete` methods for the models in the remote system
```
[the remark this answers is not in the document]
Kircheneer: I re-worded this entirely, this should be better now.

Kircheneer:
> Really well done!!
> 
> I think an example would be great as well. In my mind it showing how to reuse the model for the other side will be pretty helpful.

I have put a full example into the "Develop Jobs" section of the docs with the latest revision

pyproject.toml
Kircheneer: Doesn't need to be, removed it

docs/user/developing_jobs.md
```diff
@@ -4,17 +4,90 @@ A goal of this plugin is to make it relatively quick and straightforward to deve
...
+```python
+# example_ssot_plugin/jobs.py
+from typing import Optional
+
+from diffsync import DiffSync
+from nautobot.ipam.models import VLAN
+from nautobot.extras.jobs import Job
+from nautobot_ssot.contrib import NautobotModel, NautobotAdapter
+from nautobot_ssot.jobs import DataSource
+from remote_system import APIClient  # This is only an example
```
itdependsnetworks, line 32:
```suggestion
from . import DictionaryAsAnAPIClient
```

itdependsnetworks: The thought is to couple this with a dictionary to be an actual usable example. 

e.g. 

```python
# sot_file.py

DictionaryAsAnAPIClient = {} # actual testable data.
```

development/creds.example.env
chadell: the integration with different controllers is gonna be implemented with the `controllers` models coming soon to Nautobot, so we could select different url endpoints easily. But, until then, I would keep as it is

nautobot_ssot/contrib.py
chadell: I don't see any answer to this point

docs/user/modeling.md
```diff
@@ -0,0 +1,144 @@
...
+
+!!! note
+    If you want to sync data into a model provided by a [Nautobot App](https://docs.nautobot.com/projects/core/en/stable/plugins/), you need to navigate to its respective source code repository. Furthermore, you may need to manually import the model if you're using the Nautobot Shell approach.
+
+## Custom Fields
+
+For [custom fields](https://docs.nautobot.com/projects/core/en/stable/models/extras/customfield/), you will need to use the `nautobot_ssot.contrib.CustomFieldAnnotation` class. Given a custom field called "Test" on the circuit provider model, this is how you could map that custom field into the diffsync model field `test_custom_field`:
+
+```python
+from typing_extensions import Annotated  # Python<3.9 - you need to install this module
```
chadell, line 29: I think it would be great for compatibility with something like:

```py
try:
    from typing import Annotated
except ModuleNotFoundError:
    from typing_extensions import Annotated
```

docs/user/modeling.md
```diff
@@ -0,0 +1,144 @@
...
+    _modelname = "provider"
+    _identifiers = ("name",)
+    _attributes = ("custom_field",)
+
+    name: str
+    test_custom_field: Annotated[str, CustomFieldAnnotation(name="Test")]
+```
+
+!!! note
+    Defining a `CustomFieldAnnotation` variable is necessary since custom field names may include spaces, which are un-representable in Python object field names.
```
chadell, line 45: why not use a name with an space in the example?

docs/user/modeling.md
```diff
@@ -0,0 +1,144 @@
...
+from typing_extensions import Annotated  # Python<3.9 - you need to install this module
+# from typing import Annotated  # Python>=3.9
+from nautobot.circuits.models import Provider
+from nautobot_ssot.contrib import NautobotModel, CustomFieldAnnotation
+
+class ProviderModel(NautobotModel):
+    _model = Provider
+    _modelname = "provider"
+    _identifiers = ("name",)
+    _attributes = ("custom_field",)
```
chadell, line 38: to me, it's not clear the relationship between  "custom_field" and `test_custom_field`.

It maybe be that we only need to define one attribute "custom_field" and add add many different custom fields? or maybe, the magic comes from appending the slugify name of the custom field in the class field? I think that an example with two custom fields would deambiguate this question

docs/user/modeling.md
```diff
@@ -0,0 +1,144 @@
...
+    name: str
+    test_custom_field: Annotated[str, CustomFieldAnnotation(name="Test")]
+```
+
+!!! note
+    Defining a `CustomFieldAnnotation` variable is necessary since custom field names may include spaces, which are un-representable in Python object field names.
+
+## Many-to-one Relationships
+
+For many-to-one relationships (i.e. [foreign keys](https://docs.djangoproject.com/en/3.2/topics/db/examples/many_to_one/) or [generic foreign keys](https://docs.djangoproject.com/en/3.2/ref/contrib/contenttypes/#generic-relations)) a slightly different approach is employed. We need to add a field on our model for each field that we need in order to uniquely identify our related object behind the many-to-one relationship. We can do this using a [familiar synta](https://docs.djangoproject.com/en/3.2/topics/db/queries/#lookups-that-span-relationships) of double underscore separated paths. Assuming we want to synchronize prefixes and associate them with locations, we may be faced with the problems that locations aren't uniquely identified by name alone, but rather need the location type as well, we can address this as follows.
```
chadell, line 49:
```suggestion
For many-to-one relationships (i.e. [foreign keys](https://docs.djangoproject.com/en/3.2/topics/db/examples/many_to_one/) or [generic foreign keys](https://docs.djangoproject.com/en/3.2/ref/contrib/contenttypes/#generic-relations)) a slightly different approach is employed. We need to add a field on our model for each field that we need in order to uniquely identify our related object behind the many-to-one relationship. We can do this using a [familiar syntax](https://docs.djangoproject.com/en/3.2/topics/db/queries/#lookups-that-span-relationships) of double underscore separated paths. Assuming we want to synchronize prefixes and associate them with locations, we may be faced with the problems that locations aren't uniquely identified by name alone, but rather need the location type as well, we can address this as follows.
```

docs/user/modeling.md
```diff
@@ -0,0 +1,144 @@
...
+
+```python
+from nautobot.ipam.models import Prefix
+from nautobot_ssot.contrib import NautobotModel
+
+class PrefixModel(NautobotModel):
+    _model = Prefix
+    _identifiers = ("network", "prefix_length")
+    _attributes = ("location__name", "location__location_type__name")
```
chadell, line 58: so, the magic is that the model inspection will understand that `location` is a FK, and with these attributes we help it to identify the target, isn't it?

docs/user/modeling.md
```diff
@@ -0,0 +1,144 @@
...
+
+!!! warning
+    A limitation of this approach is that we are locked into a single kind of content type per model and foreign key, unless the field names for all content types match. In this specific case, the VM interface model does not have a `device` field. This is a known issue.
+
+## One-to-many and many-to-many Relationships
+
+For "*-to-many" relationships such as (generic) foreign keys traversed backwards or [many-to-many relationships](https://docs.djangoproject.com/en/3.2/topics/db/examples/many_to_many/), we need to employ a different mechanism. Again we start by identifying which fields of the related object we are interested in for queries **to** the model. In this case, our example will be using an `Interface` model for which we also want to sync the associated IP addresses. For our scenario, lets assume that our IP addresses can be uniquely identified through the `host` and `prefix_length` fields:
+
+```python
+from typing_extensions import TypedDict  # Python<3.9 - you need to install this module
```
chadell, line 109: smae comment as above for showing more compatible code

docs/user/performance.md
```diff
@@ -0,0 +1,181 @@
+# Optimizing for Execution Time
```
chadell, line 1: is this coming from another place? (I don't see the relationship with this PR)

chadell (approved):
I'm good with the PR even though I still added some comments in terms of documentation.

docs/user/modeling.md
Kircheneer: Actually, this is just a mistake. The attribute name still needs to match what's in `_attributes`

docs/user/modeling.md
Kircheneer: correct!

Kircheneer: I have added an example query to further illustrate this

docs/user/modeling.md
```diff
@@ -106,8 +116,10 @@ class NautobotIPAddress(NautobotModel):
 For "*-to-many" relationships such as (generic) foreign keys traversed backwards or [many-to-many relationships](https://docs.djangoproject.com/en/3.2/topics/db/examples/many_to_many/), we need to employ a different mechanism. Again we start by identifying which fields of the related object we are interested in for queries **to** the model. In this case, our example will be using an `Interface` model for which we also want to sync the associated IP addresses. For our scenario, lets assume that our IP addresses can be uniquely identified through the `host` and `prefix_length` fields:
 
 ```python
-from typing_extensions import TypedDict  # Python<3.9 - you need to install this module
-# from typing import TypedDict  # Python>=3.9
+try:
+    from typing import TypedDict  # Python>=3.9
+except ImportError:
+    from typing_extensions import TypedDict  # Python<3.9 - you need to install this module
```
itdependsnetworks, line 122: this should be handled in pyproject.toml:

```toml
typing_extensions = {version = "*", python = "<3.8"}
```

docs/user/modeling.md
```diff
@@ -0,0 +1,156 @@
...
+
+This page describes how to model various kinds of fields on a `nautobot_ssot.contrib.NautobotModel` subclass.
+
+## Quick Reference
+
+The following table describes in brief the different types of model fields and how they are handled.
+
+| Type of field             | Field name                | Notes                                                                                                                                            |
+|---------------------------|---------------------------|--------------------------------------------------------------------------------------------------------------------------------------------------|
+| Normal fields             | Has to match ORM exactly  | Make sure that the name matches the name in the ORM model.                                                                                       |
```
itdependsnetworks, line 11: Can we break these out by the most specific use case, even if it is repeating itself. I think it should be at least:

- Normal fields on the model
- Custom fields
- django.db.models.OneToOneField
- django.db.models.ForeignKey
- django.db.models.ManyToManyField:
- django.contrib.contenttypes.fields.GenericRelation
- django.contrib.contenttypes.fields.GenericForeignKey

docs/user/modeling.md
Kircheneer: Yes, but this doesn't help the person that copy pastes this example code, and they may also not be using poetry

docs/user/modeling.md
itdependsnetworks: didn't follow why this was resolved?

Kircheneer: I put it into a column in the end, sorry I should have mentioned this.

Kircheneer: This way I didn't have to duplicate anything, I just mapped it in in the last column

Kircheneer: The Markdown table looks terrible

nautobot_ssot/contrib.py
```diff
@@ -0,0 +1,354 @@
...
+    @staticmethod
+    def _handle_foreign_key(database_object, parameter_name):
+        """Handle a single foreign key field.
+
+        Given the object from the database as well as the name of parameter in the form of
+        f'{foreign_key_field_name}__{remote_field_name}'
+        return the field at 'remote_field_name' on the object behind the foreign key at 'foreign_key_field_name'.
+
+        Furthermore, 'remote_field_name' may be a series of '__' delimited lookups.
+        """
```
glennmatthews, line 175: Same request here.

nautobot_ssot/contrib.py
```diff
@@ -0,0 +1,354 @@
...
+    Base model for any diffsync models interfacing with Nautobot through the ORM.
+
+    This provides the `create`, `update` and `delete` operations in a generic fashion, meaning you don't have to
+    implement them yourself.
+
+    In order to accomplish this, the `_model` field has to be set on subclasses to map them to the corresponding ORM
+    model class.
+    """
+
+    _model = None
```
glennmatthews, line 207: Would typing here solve the `not-callable` pylint warning disabled below?

```suggestion
    _model: django.db.models.Model
```

nautobot_ssot/contrib.py
```diff
@@ -0,0 +1,354 @@
...
+            cls._check_field(field)
+
+            # Prepare handling of foreign keys
+            if "__" in field:
+                related_model, lookup = field.split("__", maxsplit=1)
+                django_field = cls._model._meta.get_field(related_model)
+                foreign_keys[related_model][lookup] = value
+                # Add a special key to the dictionary to point to the related model's class
+                foreign_keys[related_model]["_model_class"] = django_field.related_model
+                continue
```
glennmatthews, line 265: Comment with an example, perhaps?

nautobot_ssot/contrib.py
```diff
@@ -0,0 +1,354 @@
...
+                    ) from error
+                related_model = ContentType.objects.get(app_label=app_label, model=model).model_class()
+            # Set the foreign key to 'None' when none of the fields are set to anything
+            if not any(related_model_dict.values()):
+                setattr(obj, field_name, None)
+                continue
+            try:
+                related_object = related_model.objects.get(**related_model_dict)
+            except related_model.DoesNotExist as error:
+                raise ValueError(f"Couldn't find {field_name} instance with: {related_model_dict}.") from error
```
glennmatthews, line 353: Maybe also catch `MultipleObjectsReturned` and reraise?

docs/user/modeling.md
itdependsnetworks: I had the same comment, this is just an md, not a python file so there is no update to this lib, just the example user needs to know.

glennmatthews: If using `typing_extensions` is necessary to take advantage of this functionality of the plugin,  then it should be (at the very minimum an optional) dependency of the plugin.

docs/user/developing_jobs.md
```diff
@@ -4,224 +4,155 @@ A goal of this plugin is to make it relatively quick and straightforward to deve
...
-Due to these caveats it is recommended that you check carefully whether this optimization actually benefits your use case or not before applying it in production code.
+Regardless of which direction you are synchronizing data in, you need to write the `load` method for the remote adapter yourself. You can find many examples of how this can be done in the `nautobot_ssot.integrations` module, which contains pre-existing integrations with remote systems.
 
 !!! note
-    In Nautobot 2.0, jobs will no longer be atomic by default so this section will not apply anymore.
-
-#### Using Bulk ORM Operations
-
-Bulk ORM operations are available within Django to perform multiple inserts/updates in a single DB query. Take a look at the documentation for [bulk_create](https://docs.djangoproject.com/en/3.2/ref/models/querysets/#bulk-create) for example. These bulk operations should only be used very carefully in Nautobot. The following caveats apply when using them:
+As the features in `nautobot_ssot.contrib` are still very new, most existing integrations do not use them.
```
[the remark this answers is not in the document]
jdrew82: I feel like we should at least be updating the Example Job to utilize the Mixin since it's supposed to represent our recommendation around how to create an SSoT App.

docs/user/developing_jobs.md
```diff
@@ -4,224 +4,155 @@ A goal of this plugin is to make it relatively quick and straightforward to deve
...
 
-## Analyzing Job Performance
+The methods [`calculate_diff`][nautobot_ssot.jobs.base.DataSyncBaseJob.calculate_diff] and [`execute_sync`][nautobot_ssot.jobs.base.DataSyncBaseJob.calculate_diff] are both implemented by default, using the data that is loaded into the adapters through the respective methods. Note that `execute_sync` will _only_ execute when dry-run is set to false.
 
-In general there are two different metrics to optimize for when developing SSoT jobs:
+Optionally, on your Job class, also implement the [`lookup_object`][nautobot_ssot.jobs.base.DataSyncBaseJob.calculate_diff], [`data_mapping`][nautobot_ssot.jobs.base.DataSyncBaseJob.calculate_diff], and/or [`config_information`][nautobot_ssot.jobs.base.DataSyncBaseJob.config_information] APIs (to provide more information to the end user about the details of this Job), as well as the various metadata properties on your Job's Meta inner class. Refer to the example Jobs provided in this plugin for examples and further details.
```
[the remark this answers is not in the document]
jdrew82: Are the `lookup_object`, `data_mapping`, and `config_information` intended to be links? If so, then they should be wrapped in paranthesis. Also, it appears you use calculate_diff twice. Regarding Glenn's point, it might be best to instead use a permalink to the part of the code being referenced?

docs/user/modeling.md
```diff
@@ -0,0 +1,156 @@
...
+
+## Custom Fields
+
+For [custom fields](https://docs.nautobot.com/projects/core/en/stable/models/extras/customfield/), you will need to use the `nautobot_ssot.contrib.CustomFieldAnnotation` class. Given a custom field called "Test" on the circuit provider model, this is how you could map that custom field into the diffsync model field `test_custom_field`:
+
+```python
+try:
+    from typing import Annotated  # Python>=3.9
+except ModuleNotFoundError:
+    from typing_extensions import Annotated  # Python<3.9 - you need to install this module
```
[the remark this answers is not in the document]
Kircheneer: Done, thanks!

docs/user/developing_jobs.md
```diff
@@ -4,224 +4,155 @@ A goal of this plugin is to make it relatively quick and straightforward to deve
...
+    _modelname = "tenant"
+    _identifiers = ("name",)
+    _attributes = ("description",)
+
+    name: str
+    description: str
+```
 
-Similar to the database example further up, this suffers from having to perform a lookup (or in this case two) per instance of the lowest item in the hierarchy. It could be optimized (given the availability of bulk endpoints in the remote system) to look something like the following:
+As you can see when looking at the source code of the `nautobot.tenancy.models.Tenant` model, the fields on the model (`name` and `description`) match the names of the fields on that model exactly. This enables the base class `NautobotModel` to dynamically load, create, update and delete your tenants without you needing to implement this functionality yourself.
```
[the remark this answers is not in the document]
Kircheneer: Provided a link, thanks!

docs/user/developing_jobs.md
```diff
@@ -4,224 +4,155 @@ A goal of this plugin is to make it relatively quick and straightforward to deve
...
+Having created all your models, creating the Nautobot side adapter is very straight-forward:
 
-Finally, there are a couple of further ideas that could be used to improve performance. These aren't as well analyzed as the prior ones and there might be built-in support in this app for the in the future:
-
-#### Escaping the Atomic Transaction
+```python
+from nautobot_ssot.contrib import NautobotAdapter
 
-In Nautobot 1.x, all Jobs are executed in an [atomic transaction](https://en.wikipedia.org/wiki/Atomicity_(database_systems)). Atomic transactions often incur a performance overhead. The following example highlights how to have your sync operation "escape" the atomic transaction:
+from your_ssot_app.models import DiffSyncDevice, DiffSyncPrefix, DiffSyncIPAddress
```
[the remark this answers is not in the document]
Kircheneer: Good catch, thanks!

nautobot_ssot/contrib.py
```diff
@@ -0,0 +1,354 @@
...
+            ) from error
+        return diffsync_model
+
+    @staticmethod
+    def _handle_to_many_relationship(database_object, diffsync_model, parameter_name):
+        """Handle a single one- or many-to-many relationship field.
+
+        one- or many-to-many relationships are type annotated as a list of typed dictionaries. The typed
+        dictionary type expresses, which attributes we are interested in for diffsync.
+        """
```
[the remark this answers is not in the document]
Kircheneer: Done!

nautobot_ssot/contrib.py
Kircheneer: Done

nautobot_ssot/contrib.py
Kircheneer: Yes, thanks!

nautobot_ssot/contrib.py
Kircheneer: Done

nautobot_ssot/contrib.py
Kircheneer: Done

docs/user/modeling.md
Kircheneer: Fixed your point @chadell, and good point @glennmatthews - I have added a TODO comment for 2.0

docs/user/developing_jobs.md
```diff
@@ -4,224 +4,155 @@ A goal of this plugin is to make it relatively quick and straightforward to deve
...
 
-    def post_run(self):
-        super().execute_sync()
+    device = DiffSyncDevice
+    prefix = DiffSyncPrefix
+    ip_address = DiffSyncIPAddress  # Not in the `top_level` tuple, since it's a child of the prefix model
 ```
 
-Using this example, the CRUD (create, update and delete) operations of your job will not happen as part of the atomic database transaction, because `post_run` is run outside of that. This brings with it the following caveats:
+The `load` function is already implemented on this adapter and will automatically and recursively traverse any children relationships for you, provided the models are defined correctly.
```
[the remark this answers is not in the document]
Kircheneer: Yes!

docs/user/developing_jobs.md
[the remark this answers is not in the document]
Kircheneer: I have done this for the example integration, added a link here.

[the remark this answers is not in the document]
Kircheneer: Not a link but a reference rather

docs/user/modeling.md
```diff
@@ -0,0 +1,156 @@
...
+| [*-to-one relationships](#-to-one-relationships)   | Django lookup syntax      | See [here](https://docs.djangoproject.com/en/3.2/topics/db/queries/#lookups-that-span-relationships) - your model fields need to use this syntax | `django.db.models.OneToOneField`, `django.db.models.ForeignKey`, `django.contrib.contenttypes.fields.GenericForeignKey`                                                                                                 |
+| [*-to-many relationships](#-to-many-relationships) | Has to match ORM exactly  | In case of a generic foreign key see [here](#special-case-generic-foreign-key)                                                                   | `django.db.models.ManyToManyField`, `django.contrib.contenttypes.fields.GenericRelation`, `django.db.models.ForeignKey` [backwards](https://docs.djangoproject.com/en/3.2/topics/db/queries/#backwards-related-objects) |
```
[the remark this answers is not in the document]
Kircheneer: Oh I missed those.. I have added an entry into this table but I don't want to further balloon this PR up so I will open an issue once its merged

docs/user/modeling.md
```diff
@@ -0,0 +1,156 @@
...
+
+```python
+try:
+    from typing import TypedDict  # Python>=3.9
+except ImportError:
+    from typing_extensions import TypedDict  # Python<3.9 - you need to install this module
+
+class IPAddressDict(TypedDict):
+    host: str
+    prefix_length: int
```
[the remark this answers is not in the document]
Kircheneer: Added one

docs/user/modeling.md
```diff
@@ -0,0 +1,156 @@
...
+    status__name: str
+    assigned_object__app_label: str
+    assigned_object__model: str
+    assigned_object__device__name: str
+    assigned_object__name: str
+```
+
+!!! warning
+    A limitation of this approach is that we are locked into a single kind of content type per model and foreign key, unless the field names for all content types match. In this specific case, the VM interface model does not have a `device` field. This is a known issue.
```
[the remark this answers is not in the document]
Kircheneer: I added a note saying it will error out. Will add an issue to fix this once merged

glennmatthews (approved):

PR #168, opened 2023-08-23
Update for Nautobot 2.0.0.rc1 Support

jdrew82:
This PR is for updating the 2.0 branch to support Nautobot 2.0.0rc1. I found a few instances where slug was still in use for Platform and Status so I've corrected those and associated tests.

### DIFF
--- a/development/Dockerfile
+++ b/development/Dockerfile
@@ -76,3 +76,5 @@ RUN pip install -c constraints.txt -r poetry_freeze_dev.txt
 
 COPY development/nautobot_config.py ${NAUTOBOT_ROOT}/nautobot_config.py
 # !!! USE CAUTION WHEN MODIFYING LINES ABOVE
+
+USER root
--- a/nautobot_ssot/integrations/aristacv/signals.py
+++ b/nautobot_ssot/integrations/aristacv/signals.py
@@ -122,9 +122,11 @@ def post_migrate_create_platform(apps=global_apps, **kwargs):
     Manufacturer = apps.get_model("dcim", "Manufacturer")
     Platform.objects.update_or_create(
         name="arista.eos.eos",
-        napalm_driver="eos",
-        network_driver="arista_eos",
-        manufacturer=Manufacturer.objects.get(name="Arista"),
+        defaults={
+            "napalm_driver": "eos",
+            "network_driver": "arista_eos",
+            "manufacturer": Manufacturer.objects.get(name="Arista"),
+        },
     )
 
     if APP_SETTINGS.get("create_controller"):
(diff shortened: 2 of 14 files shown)


development/Dockerfile
```diff
@@ -76,3 +76,5 @@ RUN pip install -c constraints.txt -r poetry_freeze_dev.txt
 
 COPY development/nautobot_config.py ${NAUTOBOT_ROOT}/nautobot_config.py
 # !!! USE CAUTION WHEN MODIFYING LINES ABOVE
+
+USER root
```
Kircheneer, line 80: Is this a standard pattern for 2.0 migrations? If so, why?

nautobot_ssot/integrations/aristacv/signals.py
```diff
@@ -120,9 +120,10 @@ def post_migrate_create_platform(apps=global_apps, **kwargs):
     """Callback function for post_migrate() -- create Arista Platform."""
     Platform = apps.get_model("dcim", "Platform")
     Manufacturer = apps.get_model("dcim", "Manufacturer")
-    Platform.objects.get_or_create(
+    Platform.objects.update_or_create(
         name="arista.eos.eos",
         napalm_driver="eos",
+        network_driver="arista_eos",
         manufacturer=Manufacturer.objects.get(name="Arista"),
     )
```
Kircheneer, line 128:
```suggestion
    Platform.objects.update_or_create(
        name="arista.eos.eos",
        napalm_driver="eos",
        network_driver="arista_eos",
        manufacturer=Manufacturer.objects.get(name="Arista"),
    )
```
```suggestion
    Platform.objects.update_or_create(
        name="arista.eos.eos",
        defaults={
        "napalm_driver": "eos",
        "network_driver": "arista_eos",
        "manufacturer": Manufacturer.objects.get(name="Arista"),
        }
    )
```

Otherwise the updating will never happen and I believe it might crash when it encounters a platform called `arista.eos.eos` that doesn't have the other parameters matching

Kircheneer (changes requested):
Just a couple of minor things, after that we should be good

development/Dockerfile
glennmatthews: Temporary workaround for a bug in the rc1 docker images. Will be fixed/unnecessary in rc2 and later.

nautobot_ssot/integrations/aristacv/signals.py
jdrew82: Ahhh, wasn't aware of that. Good catch!

development/Dockerfile
jdrew82: This was added temporarily to address an issue with coverage not working in CI. Will be removed in future when we have a container that doesn't have the issue.

jdrew82:
@Kircheneer Tweaked the Platform update in CVP `signals.py`.

Kircheneer (approved):

PR #181, opened 2023-08-25
Python 3.11 Support

nniehoff:


pyproject.toml
```diff
@@ -26,8 +27,8 @@ packages = [
 ]
 
 [tool.poetry.dependencies]
-python = ">=3.8, <3.11"
-nautobot = "2.0.0rc1"
+python = ">=3.8, <3.12"
+nautobot = "2.0.0rc2"
```
Kircheneer, line 31:
```suggestion
nautobot = ">=2.0.0rc2"
```

No more breaking changes at this point, we can leave the dependency open ended in case there are more RCs

nniehoff: @Kircheneer Nautobot 1.6.x supports 3.11

jdrew82: This will be addressed with the 1.5.0 release that drops py3.7. I think this has been fixed in the next-2.0 branch too.

jdrew82:
Closing as this is address by #207.

PR #184, opened 2023-09-01
implements caching mechanism into the NautobotAdapter

Kircheneer:


chadell:
it would be good to get in some debug output the number of hits in the cache

nautobot_ssot/contrib.py
```diff
@@ -44,13 +45,54 @@ class ProviderModel(NautobotModel):
...
+        content_type = ContentType.objects.get_for_model(model_class)
+        model_cache_key = f"{content_type.app_label}.{content_type.model}"
+        if cached_object := self._cache[model_cache_key].get(parameter_set):
+            self._cache_hits[model_cache_key] += 1
+            return cached_object
+        # As we are using `get` here, this will error if there is not exactly one object that corresponds to the
+        # parameter set. We intentionally pass these errors through.
+        self._cache[model_cache_key][parameter_set] = model_class.objects.get(**dict(parameter_set))
+        return self._cache[model_cache_key][parameter_set]
+
```
chadell, line 95: it's the same, but what about this?

```suggestion
          else:
            # As we are using `get` here, this will error if there is not exactly one object that corresponds to the
            # parameter set. We intentionally pass these errors through.
           cached_object = self._cache[model_cache_key][parameter_set] = model_class.objects.get(**dict(parameter_set))
        return cached_object

```

chadell (approved):
it looks good to me

Kircheneer:
Should evaluate whether using [this](https://docs.djangoproject.com/en/3.2/topics/cache/) is a good idea or not.

nautobot_ssot/contrib.py
```diff
@@ -44,13 +45,54 @@ class ProviderModel(NautobotModel):
...
+        """Retrieve an object from the ORM or the cache."""
+        parameter_set = frozenset(parameters.items())
+        content_type = ContentType.objects.get_for_model(model_class)
+        model_cache_key = f"{content_type.app_label}.{content_type.model}"
+        if cached_object := self._cache[model_cache_key].get(parameter_set):
+            self._cache_hits[model_cache_key] += 1
+            return cached_object
+        # As we are using `get` here, this will error if there is not exactly one object that corresponds to the
+        # parameter set. We intentionally pass these errors through.
+        self._cache[model_cache_key][parameter_set] = model_class.objects.get(**dict(parameter_set))
```
lampwins, line 93: We should either not make any assumptions about returning a single instance, and always use `.filter()` to allow for use cases where I want a set of objects instead of a single, or offer two versions of `get_from_orm_cache` for those two different use cases.

nautobot_ssot/contrib.py
```diff
@@ -44,13 +45,54 @@ class ProviderModel(NautobotModel):
...
+        super().__init__(*args, **kwargs)
+        self.invalidate_cache()
+
+    def invalidate_cache(self, zero_out_hits=True):
+        """Invalidates all the objects in the ORM cache."""
+        self._cache = defaultdict(dict)
+        if zero_out_hits:
+            self._cache_hits = defaultdict(int)
+
+    def get_from_orm_cache(self, parameters: Dict, model_class: Type[Model]):
```
lampwins, line 83: This is good and covers many use cases, but there are situations where I do not want my cache lookup key to be tied to how the object(s) are queried. Perhaps I have some complex query logic and I just want a simple key/value cache that is disconnected from my actual query.

nautobot_ssot/contrib.py
Kircheneer: I believe that always using `filter` is a very good idea, I will have to think about list vs. queryset here though and how lazy evaluation of querysets comes into play.

Kircheneer:
I intend to keep working on this, I just had to delete the fork for a different reason.

PR #203, opened 2023-09-08
Sync Dev to Main

jdrew82:
This PR is to push the 1.4.0 release from `dev` to `main`. It's all of the work for the integration of the other SSoT Apps.


chadell (approved):
LGTM

Kircheneer (changes requested):
Branches should be the other way around, right?

Kircheneer (approved):
Upon reconsideration and due to the fact 1.4.0 was cut directly from `develop` and has not made its way into `main` yet, I agree with this PR.

PR #206, opened 2023-09-13
Update docs pins for py3.7 compatibility

cmsirbu:
Docs builds have been failing due to second order dependencies (griffe through mkdocstrings-python) not supporting Python 3.7 anymore: https://readthedocs.org/projects/nautobot-plugin-ssot/builds/

Need to hard pin last supported versions until Python 3.7 is dropped.


jdrew82 (approved):
LGTM!

PR #207, opened 2023-09-13
Drop Python 3.7 Support

jdrew82:
This PR is to update the project to drop support for Python 3.7 and update support to include 3.11.


chadell (approved):
LGTM

PR #213, opened 2023-09-18
fix: :bug: Several fixes in the ACI integration

chadell:


jdrew82:
What's the reasoning behind this? I didn't create the original ACI integration so I'm not sure why these strings were added to the DeviceRole slugs.

chadell:
> What's the reasoning behind this? I didn't create the original ACI integration so I'm not sure why these strings were added to the DeviceRole slugs.

I'm using it, and I'm fixing some bugs, still not completed.
The idea, I guess, was to differentiate the role from ACI and other, but then this custom slug makes the name inconsistent and leads to errors.

nautobot_ssot/integrations/aci/diffsync/adapters/aci.py
```diff
@@ -350,12 +351,25 @@ def load_devices(self):
...
+                        comments="",
+                    )
+                    self.add(_devicetype)
                 model = value["model"]
+
             new_device = self.device(
                 name=value["name"],
                 device_type=model,
-                device_role=value["role"],
+                device_role=value["model"],
```
chadell, line 372:
```suggestion
                device_role=value["role"],
```

chadell:
@jdrew82 it is ready for review

nautobot_ssot/integrations/aci/diffsync/adapters/aci.py
```diff
@@ -350,8 +351,21 @@ def load_devices(self):
...
+                    self.get("device_type", {"model": value["model"], "part_nbr": ""})
+                except ObjectNotFound:
+                    _devicetype = self.device_type(
+                        model=value["model"],
+                        manufacturer="Cisco",
+                        part_nbr="",
+                        u_height=1,
+                        comments="",
+                    )
+                    self.add(_devicetype)
```
Kircheneer, line 366: https://diffsync.readthedocs.io/en/latest/api/diffsync.html#diffsync.DiffSync.get_or_instantiate

```suggestion
                self.get_or_instantiate("device_type", ids={"model": value["model"], "part_nbr": ""}, attrs={"manufacturer": "Cisco", "u_height": 1, "comments": ""})
```

This should probably also work and be more concise

chadell: you are the expert ;)

PR #221, opened 2023-09-25
Add Device42 integration

jdrew82:
This PR is to add an intergration for Device42. I've added the plugin settings and environment variables to the dev environment.


chadell:
The PR makes sense. I understand that the idea for now is import as it was created, without using the new utilities: `NautobotModel`

jdrew82:
> The PR makes sense. I understand that the idea for now is import as it was created, without using the new utilities: `NautobotModel`

Correct. This was just to get it integrated for the LTM support users. Ultimately all future development is 2.0 only with bug fixes in LTM.

chadell (approved):
LGTM

PR #224, opened 2023-09-29
Allow Skipping Conflicting Apps Check

snaselj:
# Closes NaN

## What's Changed

- Added NAUTOBOT_SSOT_ALLOW_CONFLICTING_APPS environment variable to allow conflicting apps to remain in PLUGINS during the upgrade process.


docs/admin/upgrade.md
```diff
@@ -25,5 +25,14 @@ To prevent conflicts during `nautobot-ssot` upgrade:
 
 These steps will help prevent issues during `nautobot-ssot` upgrades. Always back up your data and thoroughly test your configuration after these changes.
 
+!!! note
+    It's possible to allow conflicting apps to remain in `PLUGINS` during the upgrade process. You can specify the following environment variable in `development/development.env` to allow conflicting apps:
```
Kircheneer, line 29: This path is specific to the dev environment, should we just tell people to make sure the environment variable is set, and not suggest a specific way?

snaselj: Fixed

Kircheneer (approved):
LGTM

PR #229, opened 2023-09-29
Add the ability to customize queryset used on NautobotModels

qduk:
This is my attempt to add to @Kircheneer great work! I was having some trouble dealing with the `prefetch_related_parameters`. I ended up duplicating the code in the NautobotModel but I'm sure there is a more succinct way I'm missing. I can definitely add more tests as well. I just didn't want to go crazy in case we go a different direction.


docs/user/modeling.md
```diff
@@ -156,3 +156,26 @@ Through us defining the model, Nautobot will now be able to dynamically load IP
...
+    _modelname = "tenant"
+    _identifiers = ("name",)
+    _attributes = ("description",)
+
+    name: str
+    description: str
+
+    @classmethod
+    def get_queryset(cls):
+        return Tenant.objects.filter(name__starts="s")
```
Kircheneer, line 180:
```suggestion
        return Tenant.objects.filter(name__startswith="s")
```

See [here](https://docs.djangoproject.com/en/4.2/ref/models/querysets/#startswith)

docs/user/modeling.md
```diff
@@ -156,3 +156,26 @@ Through us defining the model, Nautobot will now be able to dynamically load IP
 
 !!! note
     Although `Interface.ip_addresses` is a generic relation, there is only one content type (i.e. `ipam.ipaddress`) that may be related through this relation, therefore we don't have to specific this in any way.
+
+
+## Model Querysets
+
+By default, the builtin NautobotAdapter will load all objects of a given model. However, you can provide your own queryset by overriding the `get_queryset` method. Here is an example where the adapter would only load Tenant objects whose name starts with an "s".
```
Kircheneer, line 163: Perhaps reword this to focus more on the word 'filtering' - we should also mention 'queryset' but if we have users that don't have in-depth enough Django knowledge this might help them find this feature better.

nautobot_ssot/contrib.py
```diff
@@ -257,6 +252,16 @@ class NautobotModel(DiffSyncModel):
 
     _model: Model
 
+    @classmethod
```
Kircheneer, line 255: I _think_ this probably doesn't need to be a `classmethod`?

nautobot_ssot/contrib.py
```diff
@@ -257,6 +252,16 @@ class NautobotModel(DiffSyncModel):
...
+    def get_queryset(cls):
+        """Get the queryset used to load the models data from Nautobot."""
+        parameter_names = list(cls._identifiers) + list(cls._attributes)
+
+        # Here we identify any foreign keys (i.e. fields with '__' in them) so that we can load them directly in the
+        # first query if this function hasn't been overridden.
+        prefetch_related_parameters = [parameter.split("__")[0] for parameter in parameter_names if "__" in parameter]
+        return cls._model.objects.prefetch_related(*prefetch_related_parameters).all()
```
Kircheneer, line 263: How about we have two methods:
- `_get_queryset` used internally to get the actual queryset, call `get_queryset` and adds the `prefetch_related`
- `get_queryset`, which defaults to `return self._model.objects.all()` but may be overwritten

This way we don't loose the performance benefits when overwriting/pass the burden of implementing that to the user

nautobot_ssot/tests/test_contrib.py
```diff
@@ -309,6 +309,68 @@ class Adapter(NautobotAdapter):
...
+
+        new_tenant_name = "NASA"
+        Tenant.objects.create(name=new_tenant_name)
+        adapter = Adapter()
+        adapter.load()
+        diffsync_tenant_1 = adapter.get(TenantModel, new_tenant_name)
+        diffsync_tenant_2 = adapter.get(TenantModel, "Test")
+
+        self.assertEqual(new_tenant_name, diffsync_tenant_1.name)
+        self.assertEqual("Test", diffsync_tenant_2.name)
```
Kircheneer, line 340: Can we drop this test? I am not sure it provides extra coverage in comparison to the rest of the existing tests, as they will be implicitly testing the default implementation. Thoughts?

nautobot_ssot/tests/test_contrib.py
```diff
@@ -309,6 +309,68 @@ class Adapter(NautobotAdapter):
...
+            top_level = ("tenant",)
+            tenant = TenantModel
+
+        new_tenant_name = "NASA"
+        Tenant.objects.create(name=new_tenant_name)
+        adapter = Adapter()
+        adapter.load()
+        diffsync_tenant = adapter.get(TenantModel, new_tenant_name)
+
+        self.assertEqual(new_tenant_name, diffsync_tenant.name)
```
Kircheneer, line 372: Maybe we can create another tenant that falls outside of the filter and ensure it isn't loaded?

Kircheneer (changes requested):
Thanks for the PR, really appreciate it! I have a couple of remarks, but its looking really good already 🚀

nautobot_ssot/tests/test_contrib.py
qduk: Yeah I was thinking the same thing as I wrote it. I think it helped me figure out the structure of the test so I ended up leaving it. I'll yank it out. Thanks!

nautobot_ssot/contrib.py
qduk: After the refactor I still have this as a `@classmethod`. We can meet and discuss.

Kircheneer: Nevermind, I now understand why this has to be like this.

nautobot_ssot/contrib.py
```diff
@@ -59,13 +59,8 @@ def _get_parameter_names(diffsync_model):
...
         """Given a diffsync model class, load a list of models from the database and return them."""
         parameter_names = self._get_parameter_names(diffsync_model)
-
-        # Here we identify any foreign keys (i.e. fields with '__' in them) so that we can load them directly in the
-        # first query.
-        prefetch_related_parameters = [parameter.split("__")[0] for parameter in parameter_names if "__" in parameter]
-
-        # TODO: Allow for filtering, i.e. not getting all models from a table but just some.
-        for database_object in diffsync_model._model.objects.prefetch_related(*prefetch_related_parameters).all():
+        print(type(diffsync_model))
```
Kircheneer, line 62:
```suggestion
```

Kircheneer (changes requested):
Nice, we're pretty much there at this point, just the deletion of a `print` call is missing :)

docs/user/modeling.md
```diff
@@ -156,3 +156,27 @@ Through us defining the model, Nautobot will now be able to dynamically load IP
 
 !!! note
     Although `Interface.ip_addresses` is a generic relation, there is only one content type (i.e. `ipam.ipaddress`) that may be related through this relation, therefore we don't have to specific this in any way.
+
+
+## Filtering Object Loaded From Nautobot
```
Kircheneer, line 161:
```suggestion
## Filtering Objects Loaded From Nautobot
```

Kircheneer (approved):
LGTM

PR #230, opened 2023-09-29
Update to Nautobot 2.0

jdrew82:
This PR is for merging everything from `next-2.0` into develop for the 2.0 release to support Nautobot 2.0.


Kircheneer:
Open points:

- [ ] CI
- [ ] Updating lifecycle management dependency

jdrew82:
As we discussed I'm going to push what I've got working of D42 up for that PR and then we can merge it. That should also fix the issues you mentioned are pending.

PR #231, opened 2023-10-01
Update Device42 Integration for Nautobot 2.0

jdrew82:
This PR is to contain all the changes required for getting the Device42 integration updated for Nautobot 2.0.


jdrew82:
@Kircheneer I'm having issues with some of the tests from your `test_contrib.py` file for the Mixin stuff. I've marked them as skip for now. Can you fix the tests for Nautobot 2.0 when you get a chance? This should also fix #230 and have that ready to merge.

nautobot_ssot/integrations/device42/diffsync/models/base/ipam.py
```diff
@@ -45,10 +45,11 @@ class IPAddress(DiffSyncModel):
 
     _modelname = "ipaddr"
     _identifiers = ("address", "vrf")
-    _attributes = ("available", "label", "device", "interface", "primary", "tags", "custom_fields")
+    _attributes = ("subnet", "available", "label", "device", "interface", "primary", "tags", "custom_fields")
     _children = {}
 
     address: str
+    subnet: str
```
Kircheneer, line 52: Should this be `subnet_mask` or `prefix_length`? Or is it a link to the corresponding prefix?

nautobot_ssot/jobs/examples.py
```diff
@@ -4,6 +4,7 @@
```
Kircheneer, line 4: While I see that these are necessary changes, I'm just wondering why they are introduced in this PR?

Kircheneer: Also the implementation of this example job is swapped over to the `contrib` base models in `develop`, so this may unfortunately need to be redone once more when we get the latest changes from develop into the 2.0 stream

nautobot_ssot/tests/test_contrib.py
```diff
@@ -228,19 +228,20 @@ class Adapter(NautobotAdapter):
...
+        diffsync_ip_address = adapter.get(NautobotIPAddress, {"host": "192.0.2.1", "mask_length": 24})
 
-        self.assertEqual(diffsync_ip_address.assigned_object__app_label, "dcim")
-        self.assertEqual(diffsync_ip_address.assigned_object__model, "interface")
-        self.assertEqual(diffsync_ip_address.assigned_object__device__name, self.device_name)
-        self.assertEqual(diffsync_ip_address.assigned_object__name, self.interface_name)
+        # self.assertEqual(diffsync_ip_address.assigned_object__app_label, "dcim")
+        # self.assertEqual(diffsync_ip_address.assigned_object__model, "interface")
+        # self.assertEqual(diffsync_ip_address.assigned_object__device__name, self.device_name)
+        # self.assertEqual(diffsync_ip_address.assigned_object__name, self.interface_name)
```
Kircheneer, line 243: This still looks like it needs fixing - did something horribly break here causing you to comment it out?

whitej6: Previously IPAddresses were a GFK from ipam.IPAddress to dcim.Interface but now they are ManyToManyField from dcim.Interface to ipam.IPAddress. The `assigned_object` attr is no longer relevant

nautobot_ssot/jobs/examples.py
whitej6: @Kircheneer the amount of work to refactor the example job so it is 2.0 compatible is out of scope at this time. We can re-introduce this later if we decide it's needed.

nautobot_ssot/integrations/device42/diffsync/models/base/ipam.py
whitej6: FK to parent prefix

jdrew82: This is to hold the actual prefix for designating the parent prefix during IP Address creation.

Issue #241, opened 2023-10-11
Infoblox Not Filtering Subnets


jdrew82:
### Environment
* Python version:  3.8
* Nautobot version:  1.6.0
* nautobot-ssot version:  1.6.0
* nautobot-ssot integration: Infoblox

### Expected Behavior
Specifying the `import_subnets` plugin setting should limit the subnets that are imported to only the ones in the list.

### Observed Behavior
All subnets are imported from Infoblox instance.

### Steps to Reproduce
1. Install Nautobot and SSoT.
2. Setup Infoblox integration and include subnets to filter on.
3. Run import Job.

jdrew82:
This should be fixed now so closing.

PR #242, opened 2023-10-11
Fix Infoblox import_subnets setting

jdrew82:
This PR should fix the bug that was mentioned in #241. It looks like one of the plugin settings keys got missed being updated for the `infoblox_` prepend. This corrects it.


jdrew82:
I've gone ahead and updated this to include the refactor from the [old repo](https://github.com/nautobot/nautobot-plugin-ssot-infoblox/pull/143). This should allow support for containers to be included in the subnet filter setting.

nautobot_ssot/integrations/infoblox/diffsync/adapters/infoblox.py
```diff
@@ -45,13 +45,56 @@ def __init__(self, *args, job=None, sync=None, conn=None, **kwargs):
             )
             raise PluginImproperlyConfigured
 
+    # credit to @Eric-Jckson for the conainer code addition: https://github.com/nautobot/nautobot-plugin-ssot-infoblox/pull/143
```
Kircheneer, line 48: Can we do `Co-Authored-By` instead? I would prefer not to have this comment here because it might quickly lose meaning compared to the co-authorship

nautobot_ssot/integrations/infoblox/diffsync/adapters/infoblox.py
```diff
@@ -45,13 +45,56 @@ def __init__(self, *args, job=None, sync=None, conn=None, **kwargs):
...
+            if root_containers:
+                stack = [root_containers[0]]
+
+            while stack:
+                current_node = stack.pop()
+                flattened_tree.append(current_node)
+                children = self.conn.get_child_network_containers(prefix=current_node["network"])
+                stack.extend(children)
+
+            return flattened_tree
```
Kircheneer, line 66: Can this be a normal method instead of a nested function?

nautobot_ssot/integrations/infoblox/diffsync/adapters/infoblox.py
```diff
@@ -45,13 +45,56 @@ def __init__(self, *args, job=None, sync=None, conn=None, **kwargs):
...
+        def remove_duplicates(network_list: list) -> list:
+            """Removes duplicate networks from a list of networks."""
+            seen_networks = set()
+            new_list = []
+            for network in network_list:
+                if network["network"] not in seen_networks:
+                    new_list.append(network)
+                    seen_networks.add(network["network"])
+
+            return new_list
```
Kircheneer, line 77: Same here

nautobot_ssot/integrations/infoblox/diffsync/adapters/infoblox.py
```diff
@@ -45,13 +45,56 @@ def __init__(self, *args, job=None, sync=None, conn=None, **kwargs):
...
+                # If it doesn't, we can just get all subnets from the container
+                if tree:
+                    for subnet in tree:
+                        subnets.extend(self.conn.get_child_subnets_from_container(prefix=subnet["network"]))
+                else:
+                    subnets.extend(self.conn.get_all_subnets(prefix=prefix))
+
+            # Remove duplicates if a child subnet is included infoblox_import_subnets config
+            subnets = remove_duplicates(subnets)
+            all_networks = remove_duplicates(containers) + subnets
```
Kircheneer, line 97: What if we just use a set right away, do we then need this duplicate removal?

nautobot_ssot/integrations/infoblox/utils/client.py
```diff
@@ -1224,7 +1224,8 @@ def update_ipaddress(self, ip_address, **data):  # pylint: disable=inconsistent-
         logger.info("Infoblox IP Address updated: %s", response.json())
         return response.json()
 
-    def get_network_containers(self):
+    # credit to @Eric-Jckson for this code: https://github.com/nautobot/nautobot-plugin-ssot-infoblox/pull/143
```
Kircheneer, line 1227: Same comment as above

nautobot_ssot/integrations/infoblox/utils/client.py
```diff
@@ -1248,10 +1249,101 @@ def get_network_containers(self):
...
+
+        params.update({"network_container": prefix})
+
+        try:
+            response = self._request("GET", url_path, params=params)
+        except HTTPError as err:
+            logger.info(err.response.text)
+            return []
+        logger.info(response.json())
+        return response.json().get("result")
```
Kircheneer, line 1349: Keeping in sync with the approach from above

```suggestion
        response = response.json()
        logger.info(response)
        return response.get("result")
```

Kircheneer (changes requested):

nautobot_ssot/integrations/infoblox/diffsync/adapters/infoblox.py
jdrew82: Done.

nautobot_ssot/integrations/infoblox/diffsync/adapters/infoblox.py
jdrew82: Done

nautobot_ssot/integrations/infoblox/diffsync/adapters/infoblox.py
jdrew82: Done as part of moving functions to client utils.

nautobot_ssot/integrations/infoblox/utils/client.py
jdrew82: Done as part of moving nested functions to client utils.

nautobot_ssot/integrations/infoblox/diffsync/adapters/infoblox.py
jdrew82: I believe these are lists of dicts so aren't going to be hashable for this case.

PR #243, opened 2023-10-11
Fix Infoblox import_subnet for ltm-1.6

jdrew82:
This PR addresses the bug in #241 for the ltm-1.6 branch.


jdrew82:
I've gone ahead and updated this to include the refactor from the [old repo](https://github.com/nautobot/nautobot-plugin-ssot-infoblox/pull/143). This should allow support for containers to be included in the subnet filter setting.

Kircheneer (approved):
LGTM

Issue #257, opened 2023-10-30
KeyError When Processing SSoT CustomFields


Renrut5:
### Environment
* Python version:  3.8
* Nautobot version:  2.0.3
* nautobot-ssot version:  2.0.0, `develop` branch

### Expected Behavior
Custom field values to be populated during SSoT import.

### Observed Behavior
`KeyError` raised during SSoT import when processing custom fields.

Observed behavior happens in `nautobot_ssot/contrib.py`, line 94
- [Class and Method: `NautobotAdapter._load_single_object()`](https://github.com/nautobot/nautobot-plugin-ssot/blob/develop/nautobot_ssot/contrib.py#L94)

Specifically, the KeyError happens on the `database_object.cf['metadata.name']` field.

Troubleshooting has shown that the `database_object.cf['metadata.name']` field does not include all custom fields registered to the object.

During troubleshooting, I attempted to force a key into that dictionary if it did not exist already. The annotation in the DiffSync model used the CustomField key (using `edn_v6`), not the label name (did not use `EDN v6`).

```
for metadata in metadata_for_this_field:
      if isinstance(metadata, CustomFieldAnnotation):
          if metadata.name in database_object.cf:
              # Add dictionary key if it doesn't exist
              database_object.cf[metadata.name] = ""
          parameters[parameter_name] = database_object.cf[metadata.name]
          is_custom_field = True
          break
    if is_custom_field:
      continue
```

This worked, but limited to only text-based custom fields. This was not compatible with other custom field types.

Traceback:
```
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/nautobot_ssot/jobs/base.py", line 322, in run
    self.sync_data(memory_profiling)
  File "/usr/local/lib/python3.8/site-packages/nautobot_ssot/jobs/base.py", line 145, in sync_data
    self.load_target_adapter()
  File "/source/nautobot_ssot_customername/jobs.py", line 41, in load_target_adapter
    self.target_adapter.load()
  File "/source/nautobot_ssot_customername/diffsync/adapters/nautobot.py", line 40, in load
    self._load_objects(diffsync_model)
  File "/usr/local/lib/python3.8/site-packages/nautobot_ssot/contrib.py", line 75, in _load_objects
    self._load_single_object(database_object, diffsync_model, parameter_names)
  File "/source/nautobot_ssot_customername/diffsync/adapters/nautobot.py", line 64, in _load_single_object
[...]

PR #259, opened 2023-10-30
add fix for issue #257

Renrut5:
One important note is that instead of the custom field `label` being used, I used the `key` to make it work. Using the label creates a different custom field that isn't registered and then will start throwing out warnings that the the target object doesn't have that field.

### DIFF
--- a/nautobot_ssot/contrib.py
+++ b/nautobot_ssot/contrib.py
@@ -91,6 +91,12 @@ class NautobotAdapter(DiffSync):
             metadata_for_this_field = getattr(type_hints[parameter_name], "__metadata__", [])
             for metadata in metadata_for_this_field:
                 if isinstance(metadata, CustomFieldAnnotation):
+                    try:
+                        parameters[parameter_name] = database_object.cf[metadata.name]
+                    except KeyError:
+                        # Custom field not in values passed, field should be skipped and value removed from object if set
+                        # To skip or remove value, simply don't set the key in the parameters dict
+                        pass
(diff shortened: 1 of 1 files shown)


mzbroch:
1. Please use `key in dict` - we are assuming the key could not be in a dict 
2. Please see my comment https://github.com/nautobot/nautobot-plugin-ssot/issues/257#issuecomment-1786071434 
3. We should avoid ingesting data based on the `NautobotModel` attributes without checking for CustomField types

nautobot_ssot/contrib.py
```diff
@@ -91,7 +91,8 @@ def _load_single_object(self, database_object, diffsync_model, parameter_names):
             metadata_for_this_field = getattr(type_hints[parameter_name], "__metadata__", [])
             for metadata in metadata_for_this_field:
                 if isinstance(metadata, CustomFieldAnnotation):
-                    parameters[parameter_name] = database_object.cf[metadata.name]
+                    if metadata.name in database_object.cf:
+                        parameters[parameter_name] = database_object.cf[metadata.name]
```
Kircheneer, line 95: Just a note to myself for later review:

If the pydantic model field that corresponds to this parameter doesn't have a default and we don't hit this `if` here, this will likely result in an error further down:

```
        try:
            diffsync_model = diffsync_model(**parameters)
        except pydantic.ValidationError as error:
            raise ValueError(f"Parameters: {parameters}") from error
```

Probably we want to handle this explicitly and provide an error message if it is the case.

PR #263, opened 2023-11-02
Ssot snow delete function

collin-wicker:
Added delete functionality to ServiceNowCRUDMixin in models.py. Disabled cacheing functionality in models.py as it was causing unpredictable results when testing deleting records from ServiceNow. Tested this code with a ServiceNow personal development environment and everything worked as intended in the current format. 

I was unsure as to whether the delete functionality should be available by default in the ServiceNow job but it is currently available by default. Please contact me with any questions/concerns.

### DIFF
--- a/nautobot_ssot/integrations/servicenow/diffsync/adapter_servicenow.py
+++ b/nautobot_ssot/integrations/servicenow/diffsync/adapter_servicenow.py
@@ -4,6 +4,7 @@ from base64 import b64encode
 import json
 import os
 
+from collections import defaultdict
 from diffsync import DiffSync
 from diffsync.enum import DiffSyncFlags
 from diffsync.exceptions import ObjectAlreadyExists, ObjectNotFound
@@ -15,6 +16,10 @@ from . import models
 
 class ServiceNowDiffSync(DiffSync):
     """DiffSync adapter using pysnow to communicate with a ServiceNow server."""
+
+    # create defaultdict object to store objects that should be deleted from ServiceNow if they do not
+    # exist in Nautobot
+    objects_to_delete = defaultdict(list)
 
     company = models.Company
     device = models.Device  # child of location
@@ -293,3 +298,23 @@ class ServiceNowDiffSync(DiffSync):
         self.bulk_create_interfaces()
 
         source.tag_involved_objects(target=self)
+
+        # If there are objects inside any of the lists in objects_to_delete then iterate over those objects
+        # and remove them from ServiceNow
+        if (
+            self.objects_to_delete["interface"]
+            or self.objects_to_delete["device"]
+            or self.objects_to_delete["product_model"]
+            or self.objects_to_delete["location"]
+            or self.objects_to_delete["company"]
+        ):
+            for grouping in (
+                "interface",
+                "device",
+                "product_model",
+                "location",
+                "company",
+            ):
+                for sn_object in self.objects_to_delete[grouping]:
+                    sn_object.delete()
+                self.objects_to_delete[grouping] = []
--- a/nautobot_ssot/integrations/servicenow/diffsync/models.py
+++ b/nautobot_ssot/integrations/servicenow/diffsync/models.py
@@ -15,7 +15,7 @@ class ServiceNowCRUDMixin:
     _sys_id_cache = {}
     """Dict of table -> column_name -> value -> sys_id."""
 
-    def map_data_to_sn_record(self, data, mapping_entry, existing_record=None):
+    def map_data_to_sn_record(self, data, mapping_entry, existing_record=None, clear_cache=False):
         """Map create/update data from DiffSync to a corresponding ServiceNow data record."""
         record = existing_record or {}
         for mapping in mapping_entry.get("mappings", []):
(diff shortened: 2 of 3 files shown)


nautobot_ssot/integrations/servicenow/jobs.py
```diff
@@ -20,13 +20,12 @@
...
 
-    debug = BooleanVar(description="Enable for more verbose logging.")
+    debug = BooleanVar(description="Enable for more verbose logging.", default=True)
 
-    # TODO: not yet implemented
-    # delete_records = BooleanVar(
-    #     description="Delete records from ServiceNow if not present in Nautobot",
-    #     default=False,
-    # )
+    delete_records = BooleanVar(
```
Kircheneer, line 25: Is this used anywhere?

nautobot_ssot/integrations/servicenow/diffsync/adapter_servicenow.py
```diff
@@ -293,3 +298,23 @@ def sync_complete(self, source, diff, flags=DiffSyncFlags.NONE, logger=None):
...
+            for grouping in (
+                "interface",
+                "device",
+                "product_model",
+                "location",
+                "company",
+            ):
+                for sn_object in self.objects_to_delete[grouping]:
+                    sn_object.delete()
+                self.objects_to_delete[grouping] = []
```
Kircheneer, line 320: Can we use the `delete` method on the models directly to delete records? I thought this `complete_sync` hook is only necessary so that we can use bulk operations

nautobot_ssot/integrations/servicenow/diffsync/models.py
```diff
@@ -82,7 +83,29 @@ def update(self, attrs):
...
+                f"in table {entry['table']}"
+            )
+            return None
+        try:
+            self.diffsync.job.logger.warning(f"{self._modelname} {self.get_identifiers()} will be deleted.")
+            _object = sn_resource.get(query=query)
+            self.diffsync.objects_to_delete[self._modelname].append(_object)
+            print(self.diffsync.objects_to_delete[self._modelname])
+            super().delete()
+        except Exception as exception:
```
Kircheneer, line 105: Can you narrow this exception down a little so its not a blanket `except Exception`?

nautobot_ssot/integrations/servicenow/diffsync/models.py
```diff
@@ -29,17 +29,18 @@ def map_data_to_sn_record(self, data, mapping_entry, existing_record=None):
                 sys_id = None
                 if "column" not in mapping["reference"]:
                     raise NotImplementedError
-                column_name = mapping["reference"]["column"]
+                # column_name = mapping["reference"]["column"]
                 if value is not None:
+                    # Cacheing disabled as it was introducing instability and unpredictable results during delete method testing
```
Kircheneer, line 34: Did you figure out why this was the case? Can we possibly fix the cache mechanism instead? I reckon its existence is probably necessary for some bigger deployments. Also if we do decide to omit caching, can you delete the commented out code?

Kircheneer (changes requested):
Thanks for the PR! It would probably be good if @glennmatthews could take a look as well

nautobot_ssot/integrations/servicenow/jobs.py
collin-wicker: There are several areas in the models.py file that use diffsync.job.logger.debug but I believe I defaulted the debug value to True in the jobs.py file when I was testing and must have forgotten to set it back to False.

nautobot_ssot/integrations/servicenow/diffsync/adapter_servicenow.py
jdrew82: I believe this is done to control the order of objects being deleted to ensure that dependent objects are deleted in right order.

collin-wicker: I am running the actual delete method from this location in order to control the order in which the delete method is processed for each object. The actual delete method in models.py is not actually deleting the objects but is adding them to "objects_to_delete" dictionary of lists so that they can be processed in the appropriate order (interface > device > product_model > location > company) as that order is different from what is specified in the adapters in "top_level" (company > location). I am using this method because I observed this pattern in the delete method for the ACI integration. 

I will see if I can add some logic to the diffsync/models.py delete method to resolve this.

collin-wicker: I did a bit of research and attempted to find another way to delete the items in SNOW but was unable to find another method of controller the delete order as the delete order. If this is an issue and you are aware of another method of altering the processing order for deletes please let me know.

nautobot_ssot/integrations/servicenow/diffsync/models.py
collin-wicker: I figured out what the issue with the caching functionality was and handled it in the new commits. Please review the newest commits for the changes.

nautobot_ssot/integrations/servicenow/diffsync/models.py
collin-wicker: After updating the code in the latest commit I don't believe a try/except clause is necessary here as I'm not having any issues with errors in my tests. I have eliminated this try/except clause in the latest commits.

nautobot_ssot/integrations/servicenow/jobs.py
collin-wicker: I have removed the "default=True" for the delete checkbox in the latest commits.

collin-wicker (commented):
Please review latest commits for requested changes and comments.

jdrew82 (approved):
LGTM!

Issue #248, opened 2023-10-17
Error on Delete Orphaned Data Compliance Data from metrics


etiennelatulippe:
### Environment
* Python version:  3.11.6
* Nautobot version:  2.0.1
* nautobot-ssot version:  2.0.0

### Expected Behavior
Being able to migrate from nautobot v1.62 

### Observed Behavior

```
19:27:07.153 DEBUG   nautobot.core.celery: Importing system Jobs
19:27:07.157 WARNING nautobot.extras.utils: Job class "CloudVisionDataSource" name "CloudVision ⟹ Nautobot" is not unique, changing to "CloudVision ⟹ Nautobot (2)".
19:27:07.162 INFO    nautobot.extras.utils: Refreshed Job "SSoT - Arista CloudVision: CloudVision ⟹ Nautobot (2)" from <CloudVisionDataSource>
19:27:07.163 WARNING nautobot.extras.utils: Job class "CloudVisionDataTarget" name "Nautobot ⟹ CloudVision" is not unique, changing to "Nautobot ⟹ CloudVision (2)".
19:27:07.166 INFO    nautobot.extras.utils: Refreshed Job "SSoT - Arista CloudVision: Nautobot ⟹ CloudVision (2)" from <CloudVisionDataTarget>
19:27:07.169 INFO    nautobot.extras.utils: Refreshed Job "System Jobs: Git Repository: Sync" from <GitRepositorySync>
19:27:07.175 INFO    nautobot.extras.utils: Refreshed Job "System Jobs: Git Repository: Dry-Run" from <GitRepositoryDryRun>
19:27:07.179 INFO    nautobot.extras.utils: Refreshed Job "nautobot_data_validation_engine.jobs: Run Registered Data Compliance Rules" from <RunRegisteredDataComplianceRules>
19:27:07.183 INFO    nautobot.extras.utils: Refreshed Job "nautobot_data_validation_engine.jobs: Delete Orphaned Data Compliance Data" from <DeleteOrphanedDataComplianceData>
Traceback (most recent call last):
  File "/usr/local/bin/nautobot-server", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/usr/local/lib/python3.11/site-packages/nautobot/core/cli/__init__.py", line 54, in main
    run_app(
  File "/usr/local/lib/python3.11/site-packages/nautobot/core/runner/runner.py", line 297, in run_app
    management.execute_from_command_line([runner_name, command] + command_args)
[...]

PR #264, opened 2023-11-02
Add check for missing job_class attribute on 1.x Jobs

jdrew82:
We were told through the public Slack that the issue detailed in #248 is still occurring but at a different part of the code. I've added the same check that is done for the `metric_ssot_jobs()` function to the `metric_sync_operations()` and `metric_memory_usage()`. This should hopefully fix the issue now.

### DIFF
--- a/nautobot_ssot/jobs/__init__.py
+++ b/nautobot_ssot/jobs/__init__.py
@@ -43,6 +43,8 @@ def get_data_jobs():
     data_sources = []
     data_targets = []
     for job in sync_jobs:
+        if job.job_class is None or not issubclass(job.job_class, (DataSource, DataTarget)):
+            continue
         if issubclass(job.job_class, DataSource):
             data_sources.append(job)
         if issubclass(job.job_class, DataTarget):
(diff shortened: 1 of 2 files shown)


nautobot_ssot/jobs/__init__.py
```diff
@@ -43,6 +43,8 @@ def get_data_jobs():
     data_sources = []
     data_targets = []
     for job in sync_jobs:
+        if job.job_class is None or not issubclass(job.job_class, (DataSource, DataTarget)):
```
Kircheneer, line 46:
```suggestion
        if job.job_class is None:
```

I would argue that this is enough since the `issubclass` check happens right after, right?

Kircheneer (changes requested):
Just a minor nitpick, looks good apart from this.

nautobot_ssot/jobs/__init__.py
jdrew82: Agreed, but this check should speed up the process since it eliminates all Jobs before they get to the subclass check.

Kircheneer (approved):
Fine by me!

Issue #274, opened 2023-11-15
AssertionError at /graphql/


Danmanny:
### Environment
* Python version:  3.10.13
* Nautobot version:  2.0.0
* nautobot-ssot version:  2.0.1

Work in a Nautobot v2.0.0 environment and access graphql.

Cannot access graphql

### Steps to Reproduce
1. Install Nautobot 2.0.0
2. Access Graphql

jdrew82:
Investigated this and it appears to be related to some of the CustomFields being created in various integration signals. A few have dashes in the key when dashes will break GraphQL. CustomFields should always use underscores in keys.

jeffkala:
Hit this as well with infoblox.

AssertionError: Names must match /^[_a-zA-Z][_a-zA-Z0-9]*$/ but "cf_ssot-synced-to-infoblox" does not.

jdrew82:
> Hit this as well with infoblox.
> 
> AssertionError: Names must match /^[_a-zA-Z][_a-zA-Z0-9]*$/ but "cf_ssot-synced-to-infoblox" does not.

That's been found and is addressed in the fix PR.

Kircheneer:
@jdrew82 does #275 cover all of these?

jdrew82:
This should be resolved now so closing.

PR #275, opened 2023-11-15
Fix CustomField Bug

jdrew82:
This PR is to fix the bug detailed in #274. I believe this is around CustomFields that have dashes in the key when it should be underscores instead. There is a migration included to update any user that might have the old CustomFields and migrate them to the new ones.

### DIFF
--- a/docs/user/integrations/ipfabric.md
+++ b/docs/user/integrations/ipfabric.md
@@ -36,7 +36,7 @@ There are several options available.
 
 - **Debug**: Enables more verbose logging that can be useful for troubleshooting synchronization issues.
 - **Safe Delete Mode**: Delete operations changes the object status to a predefined value (configurable via settings) and tags the object with `ssot-safe-delete` tag.
-- **Sync Tagged Only**: Only load Nautobot data into DiffSync adapters that's been tagged with `ssot-synced-from-ipfabric` tag.
+- **Sync Tagged Only**: Only load Nautobot data into DiffSync adapters that's been tagged with `ssot_last_synchronized` tag.
 - **Dry run**: This will only report the difference between the source and destination without synchronization.
 - **Site Filter**: Filter the data loaded into DiffSync by a top level location of a specified Site.
 
@@ -118,7 +118,7 @@ If you would like to change the default status change value, ensure you provide 
 
 ![Safe Delete](../../images/ipfabric-safe-delete.png)
 
-An example object that's been modified by SSoT App and tagged as `ssot-safe-delete` and `ssot-synced-from-ipfabric`. Notice the Status and child object, IPAddress has also changed to Deprecated and, it's status changed and tagged as well.
+An example object that's been modified by SSoT App and tagged as `ssot-safe-delete` and `ssot_last_synchronized`. Notice the Status and child object, IPAddress has also changed to Deprecated and, it's status changed and tagged as well.
 
 ![Safe Delete Address](../../images/ipfabric-safe-delete-ipaddress.png)
 
--- a/docs/user/integrations/ipfabric.md
+++ b/docs/user/integrations/ipfabric.md
@@ -35,8 +35,8 @@ Now back to running the job. Let's click on **Sync Now**.
 There are several options available.
 
 - **Debug**: Enables more verbose logging that can be useful for troubleshooting synchronization issues.
-- **Safe Delete Mode**: Delete operations changes the object status to a predefined value (configurable via settings) and tags the object with `ssot-safe-delete` tag.
-- **Sync Tagged Only**: Only load Nautobot data into DiffSync adapters that's been tagged with `ssot_last_synchronized` tag.
+- **Safe Delete Mode**: Delete operations changes the object status to a predefined value (configurable via settings) and tags the object with `SSoT Safe Delete` Tag.
+- **Sync Tagged Only**: Only load Nautobot data into DiffSync adapters that has the `SSoT Synced from IPFabric` Tag.
 - **Dry run**: This will only report the difference between the source and destination without synchronization.
 - **Site Filter**: Filter the data loaded into DiffSync by a top level location of a specified Site.
 
@@ -104,7 +104,7 @@ Currently, this integration will provide the ability to sync the following IP Fa
 
 ## Safe Delete Mode
 
-By design, a Nautobot SSoT app using DiffSync will Create, Update or Delete when synchronizing two data sources. However, this may not always be what we want to happen with our Source of Truth (Nautobot). A job configuration option is available and enabled by default to prevent deleting objects from the database and instead, update the `Status` of said object alongside assigning a default tag, `ssot-safe-delete`. For example, if an additional snapshot is created from IPFabric, synchronized with Nautobot and, it just so happens that a device was unreachable, down for maintenance, etc., This doesn't `always` mean that our Source of Truth should delete this object, but we may need to bring attention to this matter. We let you decide what should happen. One thing to note is that some of the objects will auto recover from the changed status if a new job shows the object is present. However, currently, IP addresses and Interfaces will not auto-update to remove the `ssot-safe-delete` tag. The user is responsible for reviewing and updating accordingly. Safe delete tagging of objects works in an idempotent way. If an object has been tagged already, the custom field defining the last update will not be updated with a new sync date from IPFabric. So, if you re-run your sync job days apart and, you'd expect the date to change, but the object has been flagged as safe to delete; you will not see an updated date on the object custom field unless the status changed, in which case the tag (depending on the object) would be removed followed by updating the last date of sync.
+By design, a Nautobot SSoT app using DiffSync will Create, Update or Delete when synchronizing two data sources. However, this may not always be what we want to happen with our Source of Truth (Nautobot). A job configuration option is available and enabled by default to prevent deleting objects from the database and instead, update the `Status` of said object alongside assigning a default Tag, `SSoT Safe Delete`. For example, if an additional snapshot is created from IPFabric, synchronized with Nautobot and, it just so happens that a device was unreachable, down for maintenance, etc., This doesn't `always` mean that our Source of Truth should delete this object, but we may need to bring attention to this matter. We let you decide what should happen. One thing to note is that some of the objects will auto recover from the changed status if a new job shows the object is present. However, currently, IP addresses and Interfaces will not auto-update to remove the `SSoT Safe Delete` Tag. The user is responsible for reviewing and updating accordingly. Safe delete tagging of objects works in an idempotent way. If an object has been tagged already, the custom field defining the last update will not be updated with a new sync date from IPFabric. So, if you re-run your sync job days apart and, you'd expect the date to change, but the object has been flagged as safe to delete; you will not see an updated date on the object custom field unless the status changed, in which case the tag (depending on the object) would be removed followed by updating the last date of sync.
 
 The default status change of an object were to be `deleted` by SSoT DiffSync operations, will be specified below. These are the default transitions states, unless otherwise specified in the configuration options of the integration by a user.
 
@@ -112,13 +112,13 @@ The default status change of an object were to be `deleted` by SSoT DiffSync ope
 - IPAddresses -> Deprecated (Does not auto-delete tag upon recovery)
 - VLAN -> Deprecated (Auto deletes tag upon recovery)
 - Site -> Decommissioning (Auto deletes tag upon recovery)
-- Interfaces -> Tagged with `ssot-safe-delete` (Does not auto-delete tag upon recovery)
+- Interfaces -> Tagged with `SSoT Safe Delete` (Does not auto-delete Tag upon recovery)
 
 If you would like to change the default status change value, ensure you provide a valid status name available for the referenced object. Not all objects share the same `Status`.
 
 ![Safe Delete](../../images/ipfabric-safe-delete.png)
 
-An example object that's been modified by SSoT App and tagged as `ssot-safe-delete` and `ssot_last_synchronized`. Notice the Status and child object, IPAddress has also changed to Deprecated and, it's status changed and tagged as well.
+An example object that's been modified by SSoT App and tagged as `SSoT Safe Delete` and `SSoT Synced from IPFabric`. Notice the Status and child object, IPAddress has also changed to Deprecated and, it's status changed and tagged as well.
 
 ![Safe Delete Address](../../images/ipfabric-safe-delete-ipaddress.png)
 
--- a/nautobot_ssot/integrations/ipfabric/constants.py
+++ b/nautobot_ssot/integrations/ipfabric/constants.py
@@ -22,4 +22,4 @@ SAFE_DELETE_DEVICE_STATUS = CONFIG.get("ipfabric_safe_delete_device_status", "Of
 SAFE_DELETE_LOCATION_STATUS = CONFIG.get("ipfabric_safe_delete_location_status", "Decommissioning")
 SAFE_DELETE_VLAN_STATUS = CONFIG.get("ipfabric_safe_delete_vlan_status", "Deprecated")
 SAFE_DELETE_IPADDRESS_STATUS = CONFIG.get("ipfabric_safe_delete_ipaddress_status", "Deprecated")
-LAST_SYNCHRONIZED_CF_NAME = "last_synced_from_ssot"
+LAST_SYNCHRONIZED_CF_NAME = "last_synced_from_sor"
--- a/nautobot_ssot/integrations/ipfabric/constants.py
+++ b/nautobot_ssot/integrations/ipfabric/constants.py
@@ -22,3 +22,4 @@ SAFE_DELETE_DEVICE_STATUS = CONFIG.get("ipfabric_safe_delete_device_status", "Of
 SAFE_DELETE_LOCATION_STATUS = CONFIG.get("ipfabric_safe_delete_location_status", "Decommissioning")
 SAFE_DELETE_VLAN_STATUS = CONFIG.get("ipfabric_safe_delete_vlan_status", "Deprecated")
 SAFE_DELETE_IPADDRESS_STATUS = CONFIG.get("ipfabric_safe_delete_ipaddress_status", "Deprecated")
+LAST_SYNCHRONIZED_CF_NAME = "last_synced_from_ssot"
--- a/nautobot_ssot/integrations/ipfabric/diffsync/diffsync_models.py
+++ b/nautobot_ssot/integrations/ipfabric/diffsync/diffsync_models.py
@@ -79,7 +79,7 @@ class DiffSyncExtras(DiffSyncModel):
                     logger.warning(f"Tagging {nautobot_object} with `ssot-safe-delete`.")
                     update = True
             if update:
-                tonb_nbutils.tag_object(nautobot_object=nautobot_object, custom_field="ssot-synced-from-ipfabric")
+                tonb_nbutils.tag_object(nautobot_object=nautobot_object, custom_field="ssot_last_synchronized")
             else:
                 logger.warning(f"{nautobot_object} has previously been tagged with `ssot-safe-delete`. Skipping...")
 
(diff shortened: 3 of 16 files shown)


docs/user/integrations/ipfabric.md
```diff
@@ -118,7 +118,7 @@ If you would like to change the default status change value, ensure you provide
 
 ![Safe Delete](../../images/ipfabric-safe-delete.png)
 
-An example object that's been modified by SSoT App and tagged as `ssot-safe-delete` and `ssot-synced-from-ipfabric`. Notice the Status and child object, IPAddress has also changed to Deprecated and, it's status changed and tagged as well.
+An example object that's been modified by SSoT App and tagged as `ssot-safe-delete` and `ssot_last_synchronized`. Notice the Status and child object, IPAddress has also changed to Deprecated and, it's status changed and tagged as well.
```
glennmatthews, line 121: Does `ssot-safe-delete` need to be changed too? What about updating the screenshot?

nautobot_ssot/integrations/ipfabric/diffsync/diffsync_models.py
```diff
@@ -129,7 +129,7 @@ def update(self, attrs):
             device_tags = location.tags.filter(pk=safe_delete_tag.pk)
             if device_tags.exists():
                 location.tags.remove(safe_delete_tag)
-        tonb_nbutils.tag_object(nautobot_object=location, custom_field="ssot-synced-from-ipfabric")
+        tonb_nbutils.tag_object(nautobot_object=location, custom_field="ssot_last_synchronized")
```
glennmatthews, line 132: Seeing how many places this is needed, it might be appropriate to define a constant for `ssot_last_synchronized` somewhere?

nautobot_ssot/integrations/servicenow/diffsync/adapter_nautobot.py
```diff
@@ -127,7 +127,7 @@ def load(self):
 
     def tag_involved_objects(self, target):
         """Tag all objects that were successfully synced to the target."""
-        # The ssot-synced-to-servicenow tag *should* have been created automatically during plugin installation
+        # The servicenow_last_synchronized tag *should* have been created automatically during plugin installation
```
glennmatthews, line 130: Why isn't this also using the `ssot_last_synchronized` and `system_of_record` cfs?

nautobot_ssot/migrations/0007_remove_dashed_custom_fields.py
glennmatthews, line 1: perhaps `replace_dashed_custom_fields`?

nautobot_ssot/migrations/0007_remove_dashed_custom_fields.py
```diff
@@ -0,0 +1,50 @@
...
+            original_key = custom_field.key
+            updated_key = CF_KEY_CHANGE_MAP[key]
+            print(
+                f'CustomField instance "{custom_field.label}" key attribute "{original_key}" is being changed to "{updated_key}".'
+            )
+            custom_field.key = updated_key
+            custom_field.save()
+
+    for ct in ContentType.objects.filter(FeatureQuery("custom_fields").get_query()):
+        relevant_custom_fields = CustomField.objects.filter(content_types=ct)
```
glennmatthews, line 26: Perhaps narrow this down further with something like `.filter(content_types=ct, key__in=CF_KEY_CHANGE_MAP.values())` so that we are just iterating the models that these specific CFs apply to rather than all models that have *any* CF defined?

nautobot_ssot/migrations/0007_remove_dashed_custom_fields.py
```diff
@@ -0,0 +1,50 @@
...
+        if not relevant_custom_fields.exists():
+            continue
+        model = apps.get_model(ct.app_label, ct.model)
+        cf_list = []
+        for instance in model.objects.all():
+            new_custom_field_data = {}
+            for cf in relevant_custom_fields:
+                new_custom_field_data[cf.key] = instance._custom_field_data.pop(CF_KEY_CHANGE_MAP.get(cf.key), None)
+            instance._custom_field_data = new_custom_field_data
```
glennmatthews, line 35: This probably needs to be reworked since I think as written it's going to throw away any custom field value that *isn't* in CF_KEY_CHANGE_MAP?

glennmatthews (changes requested):

docs/user/integrations/ipfabric.md
jdrew82: I saw the `ssot-safe-delete` as a Tag not a CustomField so left it alone.

alhogan:
Why `ssot_last_synchronized` and `servicenow_last_synchronized` instead both being updated to `ssot_last_synchronized` or including the integration name in both?

alhogan:
Do we want to include the updates need to the Infoblox integration as well in this PR or is there another PR for that one?

nautobot_ssot/integrations/ipfabric/diffsync/diffsync_models.py
jdrew82: I'm not sure I follow. This is meant to be a date record of the last time the particular object was updated from the System of Record.

nautobot_ssot/migrations/0007_remove_dashed_custom_fields.py
jdrew82: Done.

nautobot_ssot/migrations/0007_remove_dashed_custom_fields.py
jdrew82: Good idea, done!

jdrew82:
> Do we want to include the updates need to the Infoblox integration as well in this PR or is there another PR for that one?

Good catch! I dunno how I missed that one before.

nautobot_ssot/migrations/0007_remove_dashed_custom_fields.py
jdrew82: AH, this was based off the other referenced migration.

docs/user/integrations/ipfabric.md
glennmatthews: Ah then maybe just a wording update and screenshot update? The screenshot shows both as tags?

nautobot_ssot/integrations/ipfabric/diffsync/diffsync_models.py
glennmatthews: I meant a constant like `LAST_SYNCHRONIZED_CF_NAME = "ssot_last_synchronized"` and use that rather than string literals all over the place.

nautobot_ssot/integrations/servicenow/diffsync/adapter_nautobot.py
glennmatthews: Hm, then maybe something like `last_synced_from_ssot` instead of `ssot_last_synchronized` would be clearer?

nautobot_ssot/migrations/0007_remove_dashed_custom_fields.py
glennmatthews: Yeah, the other migration was explicitly processing *all* CFs and storing them into `CF_KEY_TO_NAME`. Here we're just operating on a subset.

docs/user/integrations/ipfabric.md
jdrew82: There is a Tag on the objects `SSoT Synced from IPFabric` and also a CustomField identifying IPFabric as the System of Record. The Tag is used for filtering what's imported depending upon the Job form option selection. I've updated the docs to clarify what's specifically being spoken about.

nautobot_ssot/integrations/ipfabric/diffsync/diffsync_models.py
jdrew82: Got it, can do.

nautobot_ssot/integrations/servicenow/diffsync/adapter_nautobot.py
jdrew82: I was using `ssot_last_synchronized` as that's what I've used in other projects but that suggestion is slightly shorter and yes I can see it being clearer.

nautobot_ssot/migrations/0007_replace_dashed_custom_fields.py
```diff
@@ -0,0 +1,51 @@
...
+        relevant_custom_fields = CustomField.objects.filter(content_types=ct, key__in=CF_KEY_CHANGE_MAP.values())
+        if not relevant_custom_fields.exists():
+            continue
+        model = apps.get_model(ct.app_label, ct.model)
+        cf_list = []
+        for instance in model.objects.all():
+            new_custom_field_data = {}
+            for cf in relevant_custom_fields:
+                new_custom_field_data[cf.key] = instance._custom_field_data.pop(CF_KEY_CHANGE_MAP.get(cf.key), None)
+            instance._custom_field_data = new_custom_field_data
```
glennmatthews, line 36: Still looks to me like this will throw away every custom_field_data entry that isn't part of CF_KEY_CHANGE_MAP... perhaps try and test the below?

```suggestion
            new_custom_field_data = instance._custom_field_data.copy()
            for cf in relevant_custom_fields:
                if cf.key not in new_custom_field_data:
                    new_custom_field_data[cf.key] = new_custom_field_data.pop(CF_KEY_CHANGE_MAP.get(cf.key), None)
            instance._custom_field_data = new_custom_field_data
```

nautobot_ssot/integrations/ipfabric/signals.py
```diff
@@ -76,6 +76,7 @@ def nautobot_database_ready_callback(sender, *, apps, **kwargs):  # pylint: disa
     loc_type.content_types.add(ContentType.objects.get_for_model(apps.get_model("ipam", "Prefix")))
     loc_type.content_types.add(ContentType.objects.get_for_model(VLAN))
     synced_from_models = [Device, DeviceType, Interface, Manufacturer, Location, VLAN, Role, IPAddress]
-    create_custom_field("ssot-synced-from-ipfabric", "Last synced from IPFabric on", synced_from_models, apps=apps)
-    create_custom_field("ipfabric-site-id", "IPFabric Location ID", [Location], apps=apps, cf_type="type_text")
+    create_custom_field("system_of_record", "System of Record", synced_from_models, apps=apps, cf_type="type_text")
+    create_custom_field("last_synced_from_ssot", "Last sync from System of Record", synced_from_models, apps=apps)
```
glennmatthews, line 80: nit: perhaps `last_synced_from_sor` would be more self-consistent?

nautobot_ssot/integrations/ipfabric/constants.py
```diff
@@ -22,3 +22,4 @@
 SAFE_DELETE_LOCATION_STATUS = CONFIG.get("ipfabric_safe_delete_location_status", "Decommissioning")
 SAFE_DELETE_VLAN_STATUS = CONFIG.get("ipfabric_safe_delete_vlan_status", "Deprecated")
 SAFE_DELETE_IPADDRESS_STATUS = CONFIG.get("ipfabric_safe_delete_ipaddress_status", "Deprecated")
+LAST_SYNCHRONIZED_CF_NAME = "last_synced_from_ssot"
```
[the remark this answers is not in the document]
jdrew82: IMO I think it should be something that we should utilize across the integrations as it's been extremely helpful with client engagements being able to determine which SoR some object came from when you have multiple SSoTs operating together. This might be moot point though with some of the changes coming in core Nautobot though?

nautobot_ssot/integrations/ipfabric/signals.py
jdrew82: Done.

nautobot_ssot/migrations/0007_replace_dashed_custom_fields.py
jdrew82: Updated to match.

nautobot_ssot/migrations/0007_remove_dashed_custom_fields.py
jdrew82: Got it.

jdrew82:
Migration should now be updated to work accordingly:

- Look for any CustomFields that have a key with dashes that needs to be updated.
- Updates the key for those CustomFields to the new key and saves it.
- Looks through all ContentTypes that are related to those CustomFields and copies the value for the old CustomField to the new one.
- Deletes any of the old CustomFields that might remain.

nautobot_ssot/migrations/0007_replace_dashed_custom_fields.py
```diff
@@ -0,0 +1,53 @@
+from django.db import migrations
+from nautobot.extras.utils import FeatureQuery
+
+CF_KEY_CHANGE_MAP = {
+    "servicenow_last_synchronized": "ssot-synced-to-servicenow",
+    "last_synced_from_sor": "ssot-synced-from-ipfabric",
+    "ipfabric_site_id": "ipfabric-site-id",
+    "ssot_synced_to_infoblox": "ssot-synced-to-infoblox",
```
glennmatthews, line 8: Sorry to keep nitpicking, but would it make sense to change `servicenow_last_synchronized` to `ssot_synced_to_servicenow` to be consistent with `ssot_synced_to_infoblox`?

nautobot_ssot/migrations/0007_replace_dashed_custom_fields.py
```diff
@@ -0,0 +1,53 @@
...
+                print(
+                    f'   CustomField instance "{custom_field.label}" key attribute "{original_key}" is being changed to "{new_key}".'
+                )
+                custom_field.key = new_key
+                custom_field.save()
+
+    for ct in ContentType.objects.filter(FeatureQuery("custom_fields").get_query()):
+        model = apps.get_model(ct.app_label, ct.model)
+        cf_list = []
+        for instance in model.objects.all():
```
glennmatthews, line 29: Concerned about the performance overhead here of iterating through literally every instance of every model that supports any custom fields. Can we at least narrow this down to just the set of models that the old CFs from this app would exist for - looks like that'd be `[Device, DeviceType, Interface, Manufacturer, Location, VLAN, Role, IPAddress]`?

nautobot_ssot/migrations/0007_replace_dashed_custom_fields.py
```diff
@@ -0,0 +1,53 @@
...
+            for new_cf, old_cf in CF_KEY_CHANGE_MAP.items():
+                if old_cf in instance._custom_field_data and new_cf in instance._custom_field_data:
```
glennmatthews, line 31: Should this be?

```suggestion
                if old_cf in instance._custom_field_data and new_cf not in instance._custom_field_data:
```

?

jdrew82: Nope, I do this check to ensure that the old and new CustomField keys are in place for the copy to occur. The pop should remove the old one when it copies the data to the new CustomField key and then removes the CustomField at the end.

nautobot_ssot/migrations/0007_replace_dashed_custom_fields.py
jdrew82: Done.

nautobot_ssot/migrations/0007_replace_dashed_custom_fields.py
jdrew82: Done.

jdrew82:
I've also added a fix for the Device42 integration as it does create CustomFields that could potentially have spaces in the slug/key. I've corrected it to replace spaces and dashes with underscores.

jeffkala:
Any ETA for this? @jdrew82

jdrew82:
> Any ETA for this? @jdrew82

Just asked @glennmatthews for another review of the PR. After that we should be able to merge it and cut a new release.

nautobot_ssot/migrations/0007_replace_dashed_custom_fields.py
```diff
@@ -0,0 +1,58 @@
...
+        if not CustomField.objects.filter(key=new_key).exists():
+            for custom_field in CustomField.objects.filter(key=old_key):
+                original_key = custom_field.key
+                print(
+                    f'   CustomField instance "{custom_field.label}" key attribute "{original_key}" is being changed to "{new_key}".'
+                )
+                custom_field.key = new_key
+                custom_field.save()
+
+    for ct in ContentType.objects.filter(FeatureQuery("custom_fields").get_query()):
```
glennmatthews, line 29: Doesn't look like `ct` is actually used anywhere - unnecessary outer `for` loop?

nautobot_ssot/tests/device42/unit/test_utils_nautobot.py
```diff
@@ -153,13 +153,13 @@ def test_update_custom_fields_add_cf(self):
         test_site = Location.objects.create(
             name="Test", location_type=LocationType.objects.get_or_create(name="Site")[0], status=self.status_active
         )
-        self.assertEqual(len(test_site.get_custom_fields()), 3)
+        self.assertEqual(len(test_site.get_custom_fields()), 4)
         mock_cfs = {
             "Test Custom Field": {"key": "Test Custom Field", "value": None, "notes": None},
```
glennmatthews, line 158: Perhaps?

```suggestion
            "Test Custom Field": {"key": "Test_Custom_Field", "value": None, "notes": None},
```

nautobot_ssot/migrations/0007_replace_dashed_custom_fields.py
```diff
@@ -0,0 +1,58 @@
...
+    CustomField = apps.get_model("extras", "customfield")
+    ContentType = apps.get_model("contenttypes", "ContentType")
+
+    for new_key, old_key in CF_KEY_CHANGE_MAP.items():
+        if not CustomField.objects.filter(key=new_key).exists():
+            for custom_field in CustomField.objects.filter(key=old_key):
+                original_key = custom_field.key
+                print(
+                    f'   CustomField instance "{custom_field.label}" key attribute "{original_key}" is being changed to "{new_key}".'
+                )
```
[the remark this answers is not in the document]
jdrew82: Good point. Fixed.

nautobot_ssot/migrations/0007_replace_dashed_custom_fields.py
jdrew82: Good catch. You're right, we're not using that anymore so removed.

nautobot_ssot/tests/device42/unit/test_utils_nautobot.py
jdrew82: This part of the test is actually what the information from Device42 looks like and isn't the key for the CustomField like you might think. The name is admittedly a little confusing.

nautobot_ssot/integrations/ipfabric/utilities/nbutils.py
```diff
@@ -229,10 +230,10 @@ def _tag_object(nautobot_object):
...
         if hasattr(nautobot_object, "cf"):
-            # Ensure that the "ssot-synced-from-ipfabric" custom field is present
-            if not any(cfield for cfield in CustomField.objects.all() if cfield.key == "ssot-synced-from-ipfabric"):
-                custom_field_obj, _ = CustomField.objects.get(
-                    key="ssot-synced-from-ipfabric",
+            # Ensure that the LAST_SYNCHRONIZED_CF_NAME custom field is present
+            if not any(cfield for cfield in CustomField.objects.all() if cfield.key == LAST_SYNCHRONIZED_CF_NAME):
+                custom_field_obj = CustomField.objects.get(
+                    key=LAST_SYNCHRONIZED_CF_NAME,
                 )
```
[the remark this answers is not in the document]
jdrew82: Good question. That part of the code predates this PR and appears to come from the 2.0 update. I've gone ahead and made the suggested change though as I do agree it's more efficient.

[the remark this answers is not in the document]
jdrew82: I've removed it as it is no longer necessary with the CustomField being created via signal.

PR #281, opened 2023-11-27
Add Custom Database Parameter Loader

Renrut5:
Problem:
As it currently stands, trying to sync the `time_zone` field in `Location` objects from Nautobot does not appear to be supported.
- Loading objects from the database returns a `DstTzInfo` object
- `DstTzInfo` is not JSON serializable and raises an error when used, so cannot be used for SSoT
- Unable to determine existing way to convert the `DstTzInfo` object into a string which can be compared within SSoT
- Unable to override an individual field when loading from the Nautobot adapter

Solution:
This PR provides two solutions:
1. A way for any child class of `NautobotAdapter` to add additional methods to override the default loading of individual fields/parameters (excluding custom fields and relationships). It searches for a method named `load_param_{param_name}`
2. A method for handling the `time_zone` field by converting the `DstTzInfo` class into a string enabling default handling of the `time_zone` field. Any parameter or field without a specified method will be get default handling

### DIFF
--- a/docs/dev/jobs.md
+++ b/docs/dev/jobs.md
@@ -121,6 +121,7 @@ from nautobot_ssot.contrib import NautobotAdapter
 
 from your_ssot_app.models import DiffSyncDevice, DiffSyncPrefix, DiffSyncIPAddress
 
+
 class YourSSoTNautobotAdapter(NautobotAdapter):
     top_level = ("device", "prefix")
 
--- a/docs/dev/jobs.md
+++ b/docs/dev/jobs.md
@@ -121,7 +121,6 @@ from nautobot_ssot.contrib import NautobotAdapter
 
 from your_ssot_app.models import DiffSyncDevice, DiffSyncPrefix, DiffSyncIPAddress
 
-
 class YourSSoTNautobotAdapter(NautobotAdapter):
     top_level = ("device", "prefix")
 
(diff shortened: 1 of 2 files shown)


Renrut5:
@jdrew82 I just updated the initial comment with details

Kircheneer:
Couple thoughts:
- It would be cool to have a unit test for this behaviour
- The parameter may not necessarily be called `time_zone` - I have two separate approaches in mind to remedy this
    - EITHER: Don't provide any default implementation for any `load_param` function and provide the time zone example in the docs
    - OR: Use `typing.Annotate` to map this

I think for now I prefer the EITHER approach, what do you think?

Renrut5:
@Kircheneer 

> * The parameter may not necessarily be called `time_zone` - I have two separate approaches in mind to remedy this

This is in the default `NautobotAdapter` class, so the it should always be `time_zone` unless I'm missing something. Not sure how it would be mapped if done with that approach.

Kircheneer:
> @Kircheneer
> 
> > * The parameter may not necessarily be called `time_zone` - I have two separate approaches in mind to remedy this
> 
> This is in the default `NautobotAdapter` class, so the it should always be `time_zone` unless I'm missing something. Not sure how it would be mapped if done with that approach.

This depends on the model - for some existing core model it may be literally called `time_zone` but we need to take into account the possibility of future plugin models whose time zone fields may not literally be called `time_zone`

Renrut5:
@Kircheneer , I updated based on your suggestions. The default `time_zone` was removed and used in the documentation instead.

Kircheneer (approved):
LGTM, @jdrew82 what do you think?

docs/dev/jobs.md
```diff
@@ -121,7 +121,6 @@ from nautobot_ssot.contrib import NautobotAdapter
 
 from your_ssot_app.models import DiffSyncDevice, DiffSyncPrefix, DiffSyncIPAddress
```
jdrew82, line 123: Why was the extra line removed here? There are supposed to be two lines between imports and the top most function/class.

jdrew82 (changes requested):
Just one correction to make and then I think it's g2g.

Renrut5:
@jdrew82 my mistake, I missed that. Fixed and pushed.

jdrew82 (approved):
LGTM!

Issue #266, opened 2023-11-05
ARISTACV Loading data from Nautobot fails


gerardocastaldo:
### Environment
* Python version:  3.10
* Nautobot version:  2.0.3
* nautobot-ssot version:  2.0.0

### Observed Behavior
During the "Loading data from Nautobot" jobs i receive this error

```
Traceback (most recent call last):
  File "/opt/nautobot/.local/lib/python3.10/site-packages/nautobot_ssot/jobs/base.py", line 322, in run
    self.sync_data(memory_profiling)
  File "/opt/nautobot/.local/lib/python3.10/site-packages/nautobot_ssot/jobs/base.py", line 145, in sync_data
    self.load_target_adapter()
  File "/opt/nautobot/.local/lib/python3.10/site-packages/nautobot_ssot/integrations/aristacv/jobs.py", line 113, in load_target_adapter
    self.target_adapter.load()
  File "/opt/nautobot/.local/lib/python3.10/site-packages/nautobot_ssot/integrations/aristacv/diffsync/adapters/nautobot.py", line 140, in load
    self.load_ip_addresses()
  File "/opt/nautobot/.local/lib/python3.10/site-packages/nautobot_ssot/integrations/aristacv/diffsync/adapters/nautobot.py", line 92, in load_ip_addresses
    for ipaddr in OrmIPAddress.objects.filter(interface__device__device_type__manufacturer__name="Arista"):
  File "/usr/local/lib/python3.10/site-packages/django/db/models/manager.py", line 85, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/usr/local/lib/python3.10/site-packages/nautobot/ipam/querysets.py", line 450, in filter
    return super().filter(*args, **kwargs)
  File "/usr/local/lib/python3.10/site-packages/nautobot/core/models/querysets.py", line 86, in filter
    return super().filter(*args, **self.split_composite_key_into_kwargs(composite_key, **kwargs))
  File "/usr/local/lib/python3.10/site-packages/django/db/models/query.py", line 941, in filter
    return self._filter_or_exclude(False, args, kwargs)
  File "/usr/local/lib/python3.10/site-packages/django/db/models/query.py", line 961, in _filter_or_exclude
    clone._filter_or_exclude_inplace(negate, args, kwargs)
  File "/usr/local/lib/python3.10/site-packages/django/db/models/query.py", line 968, in _filter_or_exclude_inplace
[...]

PR #282, opened 2023-11-27
Fix CloudVision SSoT Jobs

jdrew82:
This PR includes the fix for the dry-run issues brought up in #266. Both Jobs were missing run() definition for 2.0 so Job options weren't being passed to the parent Job or other parts of the code for required functionality. Now with this in place the dry-run, debug, and memory profiling options should all work in the CloudVision SSoT integration.


Kircheneer (approved):
LGTM

PR #291, opened 2023-12-04
fixes bug with custom fields in _identifiers

### LINKED ISSUE #290: `NautobotModel.get_from_db()` doesn't work for custom fields and many to many relationships
### Environment
* Python version:  
* Nautobot version:  1/2
* nautobot-ssot version: 1.6/2.0

### Expected Behavior

When defining a model like this:

```
from typing import Annotated

from nautobot.tenancy.models import Tenant

from nautobot_ssot.contrib import NautobotModel, CustomFieldAnnotation

class TenantModel(NautobotModel):
    _model = Tenant
    _modelname = "tenant"
    _identifiers = ("name", "line_of_business",)

    name: str
    line_of_business: Annotated[str, CustomFieldAnnotation(name="Line of Business")]
```

I expect `TenantModel.get_from_db()` to still work. However, it will issue a query like `Tenant.objects.get(name="example", line_of_business="Shipping")` which will crash. Instead, we should use the mapping from the `CustomFieldAnnotation`.

### Observed Behavior
There is a crash, because the field `line_of_business` doesn't exist.

### Steps to Reproduce

Described above.

Kircheneer:
Fixes #290 for the custom field part


jdrew82:
Has this been confirmed to fix the issue with CustomFields where you need to pass the CF key and not the label? When I was trying to use this for a personal project I found that using the label simply created the key in `custom_field_data` dict and breaks GraphQL and doesn't display correctly in the UI.

jdrew82 (approved):
LGTM!

jdrew82:
Guess the test results answer my previous question. We need to revert the CustomField key to be `is_global` as `Is Global` is not valid for a CustomField key. This also goes for the CustomFieldAnnotation where name actually needs to be the key form of the CustomField.

Kircheneer:
See #350

PR #293, opened 2023-12-04
Contrib backports 1.6

Kircheneer:
Backports to the LTM branch for features/fixes needed in a client environment. Pending #291 as well before an LTM release.


Kircheneer:
Handled by #357

PR #299, opened 2023-12-07
Ensure Device ContentType on Site LocationType, Role, and Tags in CVP

jdrew82:
This should ensure that these objects can be used with Device object and address the bug discussed in #266.

### DIFF
--- a/nautobot_ssot/integrations/aristacv/utils/nautobot.py
+++ b/nautobot_ssot/integrations/aristacv/utils/nautobot.py
@@ -59,7 +59,7 @@ def verify_device_role_object(role_name, role_color):
     try:
         role_obj = Role.objects.get(name=role_name)
     except Role.DoesNotExist:
-        role_obj = Role(name=role_name, color=role_color)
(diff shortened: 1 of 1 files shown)


nautobot_ssot/integrations/aristacv/utils/nautobot.py
```diff
@@ -59,6 +60,7 @@ def verify_device_role_object(role_name, role_color):
         role_obj = Role.objects.get(name=role_name)
     except Role.DoesNotExist:
         role_obj = Role(name=role_name, color=role_color)
+        role_obj.content_types.add(ContentType.objects.get_for_model(Device))
         role_obj.validated_save()
```
Kircheneer, line 64:
```suggestion
        role_obj.validated_save()
        role_obj.content_types.add(ContentType.objects.get_for_model(Device))
```

I think we need to save before manipulating the many to many relationship

whitej6 (approved):
LGTM

Longer term we should have the `LocationType` configurable

Kircheneer (approved):

PR #303, opened 2023-12-12
Add single network view and IP Address type support

Dav-C:
- Add a new setting to optionally specify a single Network View to load from Infoblox
- Address the possibility of an extensible attribute containing multiple values
- Add support for Nautobot IP address types
- Move debug message `Creating IP Address {addr}` under a debug check to prevent unintended logging

### DIFF
--- a/nautobot_ssot/integrations/infoblox/diffsync/models/nautobot.py
+++ b/nautobot_ssot/integrations/infoblox/diffsync/models/nautobot.py
@@ -39,18 +39,27 @@ def process_ext_attrs(diffsync, obj: object, extattrs: dict):  # pylint: disable
                         f"supported by Nautobot. {err}"
                     )
             if attr.lower() == "vrf":
-                try:
-                    obj.vrfs.add(diffsync.vrf_map[attr_value])
-                except KeyError as err:
-                    diffsync.job.logger.warning(
-                        f"Unable to find VRF {attr_value} for {obj} found in Extensibility Attributes '{attr}'. {err}"
-                    )
-                except TypeError as err:
-                    diffsync.job.logger.warning(
-                        f"Cannot set vrf values {attr_value} for {obj}. Multiple vrfs are assigned "
-                        f"in Extensibility Attributes '{attr}', but multiple vrf assignments are not "
-                        f"supported by Nautobot. {err}"
-                    )
+                if type(attr_value) is list:
+                    for vrf in attr_value:
+                        try:
+                            obj.vrfs.add(diffsync.vrf_map[vrf])
+                        except KeyError as err:
+                            diffsync.job.logger.warning(
+                                f"Unable to find VRF {vrf} for {obj} found in Extensibility Attributes '{attr}'. {err}"
+                            )
+                else:
+                    try:
+                        obj.vrfs.add(diffsync.vrf_map[attr_value])
+                    except KeyError as err:
+                        diffsync.job.logger.warning(
+                            f"Unable to find VRF {attr_value} for {obj} found in Extensibility Attributes '{attr}'. {err}"
+                        )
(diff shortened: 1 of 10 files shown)


nautobot_ssot/integrations/infoblox/diffsync/models/nautobot.py
```diff
@@ -32,13 +32,25 @@ def process_ext_attrs(diffsync, obj: object, extattrs: dict):
...
+                    )
             if attr.lower() == "vrf":
                 try:
                     obj.vrfs.add(diffsync.vrf_map[attr_value])
                 except KeyError as err:
                     diffsync.job.logger.warning(
                         f"Unable to find VRF {attr_value} for {obj} found in Extensibility Attributes '{attr}'. {err}"
                     )
+                except TypeError as err:
+                    diffsync.job.logger.warning(
```
jdrew82, line 49: This is inaccurate for Nautobot 2.0. You can assign multiple VRFs to a Prefix through the `vrfs` attribute on the Prefix model. It's a M2M field so you need to use add() or set().

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "nautobot-ssot"
-version = "2.0.2"
+version = "2.0.3"
```
jdrew82, line 3: Please revert this. We increment the version when we do the release.

jdrew82 (changes requested):
Just a few things to address.

nautobot_ssot/integrations/infoblox/diffsync/models/nautobot.py
Dav-C: fixed, multiple vrfs can now be assigned.

pyproject.toml
Dav-C: Changed

jdrew82 (approved):
LGTM!

nautobot_ssot/integrations/infoblox/utils/client.py
```diff
@@ -791,7 +791,8 @@ def get_all_subnets(self, prefix: str = None):
             "_return_fields": "network,network_view,comment,extattrs,rir_organization,rir,vlans",
             "_max_results": 10000,
         }
-
+        if PLUGIN_CFG.get("NAUTOBOT_INFOBLOX_NETWORK_VIEW"):
```
qduk, line 794: Should the inclusion of the `network_view` be here as well? https://github.com/nautobot/nautobot-plugin-ssot/blob/2a0f24d0612cbe934629d666a4ef9113cfa9e71b/nautobot_ssot/integrations/infoblox/utils/client.py#L1278

Dav-C: Do you mean include the check for the Network View in the `if prefix` statement instead of just above it? There is check for the Newtwork View just before this.

qduk: Sorry, I should've been more clear. You added the check for the network view in `get_all_subnets` do we also need that check in `get_network_containers` so that when `get_network_containers` gets called we are only retrieving network containers within the specified network view?

Dav-C: There is a check for the Network View in these functions: `get_all_subnets`, `get_network_containers`, `get_child_network_containers` and `get_child_subnets_from_container`. I think the check you are referring to is already included on the new line 1288.

qduk: Ah yep! Okay, I think I was just looking too fast at the code. Thanks @Dav-C!

qduk (approved):
Thank you @Dav-C!

PR #309, opened 2023-12-19
Support using long interface names with IPFabric

jmcgill298:


nautobot_ssot/tests/ipfabric/test_ipfabric_adapter.py
```diff
@@ -83,3 +83,28 @@ def test_data_loading(self):
             self.assertTrue(hasattr(interface, "ip_address"))
             self.assertTrue(hasattr(interface, "subnet_mask"))
             self.assertTrue(hasattr(interface, "type"))
+
+    @patch("nautobot_ssot.integrations.ipfabric.diffsync.adapter_ipfabric.IP_FABRIC_ELONGATE_INTERFACE_NAME", True)
```
whitej6, line 87: Not requiring to swap out but would be good to patch the setting instead of the constant. Should be the same net effect. TBH how we do constants for plugin config kinda negates the lazy loading part of Django settings BUT that pattern is everywhere and we don't document or expect users to relying on lazy loading

nautobot_ssot/management/commands/elongate_interface_names.py
```diff
@@ -0,0 +1,50 @@
...
+
+        parser.add_argument(
+            "-l",
+            "--locations",
+            default=None,
+            help="DCIM.Location names to limit which devices have their interface names updated (comma separated).",
+        )
+
+    def handle(self, *args, **options):  # noqa: D102
+        if options.get("devices"):
```
whitej6, line 33: Thoughts on first filtering to just devices applicable for ipfabric (based on custom field but later will be solved with controller models) to prevent accidentally running the command for everything?

nautobot_ssot/management/commands/elongate_interface_names.py
```diff
@@ -0,0 +1,50 @@
...
+
+from netutils.interface import canonical_interface_name as netutils_elongate_interface_name
+
+
+class Command(BaseCommand):
+    """MGMT command to use netutils to update Nautobot Interface names to use long format."""
+
+    help = (
+        "Elongate DCIM.Interface names in Nautobot based off of netutils: "
+        "https://github.com/networktocode/netutils/blob/main/netutils/interface.py#L97"
```
whitej6, line 14: What about linking to the header for this in the docs versus a branch link with line number?

https://netutils.readthedocs.io/en/latest/dev/code_reference/interface/#netutils.interface.canonical_interface_name

nautobot_ssot/tests/ipfabric/test_ipfabric_adapter.py
jmcgill298: ya, I'm not a fan either, but following convention. As for patching the setting, as far as I know, I would have to patch all of `nautobot_ssot` settings, which seems a bit excessive; is there another solution?

nautobot_ssot/management/commands/elongate_interface_names.py
jmcgill298: Personally, I feel like Devices and Locations is a good starting point, and letting users provide feedback on what else is desired; but I can implement whatever.

nautobot_ssot/management/commands/elongate_interface_names.py
jmcgill298, line 14:
```suggestion
        "https://netutils.readthedocs.io/en/latest/dev/code_reference/interface/#netutils.interface.canonical_interface_name"
```

jmcgill298: ya, I didn't like that when I did it, but couldn't think of something better. Your suggestion makes better sense.

nautobot_ssot/tests/ipfabric/test_ipfabric_adapter.py
whitej6: Yes it's a patch of all plugin config but all tests should run against the default config so if you pass in just the one toggle that's all that changed.

nautobot_ssot/management/commands/elongate_interface_names.py
whitej6: @jdrew82 what are your thoughts? I'm leaning towards constrained scoping since the management command is meant for one integration in an app not any device object in Nautobot.

nautobot_ssot/tests/ipfabric/test_ipfabric_adapter.py
jdrew82: Currently tests are being impacted by the settings in PLUGINS_CONFIG so it might be necessary to patch the specific setting. We are planning on addressing that issue in the near future.

nautobot_ssot/management/commands/elongate_interface_names.py
jdrew82: I agree on the constrained scope. It should be specific to IPFabric only.

docs/admin/integrations/ipfabric_setup.md
```diff
@@ -57,6 +57,7 @@ PLUGINS_CONFIG = {
 | `ipfabric_safe_delete_location_status` | The status that is set for a Location when the `Safe Delete Mode` flag is set in the Job. | `Decommissioning` |
 | `ipfabric_safe_delete_vlan_status` | The status that is set for a VLAN when the `Safe Delete Mode` flag is set in the Job. | `Deprecated` |
 | `ipfabric_safe_delete_ipaddress_status` | The status that is set for an IP Address when the `Safe Delete Mode` flag is set in the Job. | `Deprecated` |
+| `ipfabric_elongate_interface_name` | Whether to attempt to elongate interface names as found in IP Fabric. | `False` |
```
itdependsnetworks, line 60: why not keep use `canonical_interface_name`? I don't love either name, but at least it's consistent.

nautobot_ssot/management/commands/elongate_interface_names.py
jvanderaa: Why would we constrain this to a single app? The management command is the management command. Allow it to be run.

nautobot_ssot/tests/ipfabric/test_ipfabric_adapter.py
jvanderaa: We should fix the tests to be in a state that do not have to have the settings changed.

jdrew82: I agree and it's on the todo list.

nautobot_ssot/management/commands/elongate_interface_names.py
jdrew82: Because the change being made is specific to one integration and we don't know how it'll impact the other ones.

nautobot_ssot/tests/ipfabric/test_ipfabric_adapter.py
whitej6: I disagree on this one, this is testing a setting change so it should be tested from the PLUGIN_CONFIG side. Intercepting the constant definition could mask an improperly defined plugin config in the NautobotAppConfig.

I am okay resolving this item, whatever we do in the future should test making the setting change in nautobot config and not patching elsewhere. Django lazy loading + default values for plugin config + patching the constant that pulls out a value from plugin config could be a false sense of coverage. I would vote this is a feature request into the nautobot testing framework.

jmcgill298: actually, this is testing whatever the value of the constant is since that is what is imported/used. There should be a separate set of tests for constants.py that tests the constants are set correctly

jdrew82 (approved):
LGTM!

PR #320, opened 2024-01-10
Store sets in diffsync

snaselj:
# Closes: NaN

When using `set()` values in `DiffSyncModel` values, SSoT job failed, due to json.dumps() not being able to serialize `set()` values.

Will be necessary to implement SSoT for NetBox Importer, as it's using `set`s for `content_types` fields values:

https://github.com/nautobot/nautobot-app-netbox-importer/pull/114

## What's Changed

- Added ability to store `set()` values in `Sync.diff` fields.


Kircheneer (approved):
LGTM

PR #325, opened 2024-01-16
Update Infoblox SSoT to allow for gathering of IPv6 Prefixes

qduk:
I'm adding some options to the SSoT sync to allow the load_prefixes function to gather IPv6 subnets and network containers ONLY if the infoblox_import_subnets is not set. We can certainly build this out more, but I wanted to get this PR open to see if this is even something we want to include. I'm not sure if the ipv6 endpoint exists only on newer versions of the WAPI, but I do know they exist on [2.12](https://ipam.illinois.edu/wapidoc/#objects) and 2.13. It appears that you need credentials to access the actual WAPI documentation.

I'm also not entirely sure I have the settings setup properly as I haven't had to work with the `constants.py` file yet. Any feedback appreciated!


jdrew82 (approved):
LGTM!

PR #326, opened 2024-01-16
Fix ipfabric create_location to use correct arg name

jmcgill298:


nautobot_ssot/tests/ipfabric/test_nbutils.py
```diff
@@ -23,7 +24,7 @@ class TestNautobotUtils(TestCase):
 
     def setUp(self):
         """Setup."""
-        reg_loctype = LocationType.objects.update_or_create(name="Region")[0]
+        reg_loctype = LocationType.objects.update_or_create(name="Site")[0]
```
jdrew82, line 27: It we're changing this to be a Site LocationType we should probably be changing the variable name accordingly.

jdrew82 (approved):
LGTM!

PR #327, opened 2024-01-18
BugFix: account for VLAN not having VLANGroup

jmcgill298:


jdrew82:
Is this for cases where a VLAN is added outside the Infoblox import? If I remember correctly, Infoblox associates all VLANs to a VLANView, which we import as VLANGroups.

jmcgill298:
> Is this for cases where a VLAN is added outside the Infoblox import? If I remember correctly, Infoblox associates all VLANs to a VLANView, which we import as VLANGroups.

TBH, I don't remember how I triggered this, but I did at some point hit an AttributeError for trying to access the `name` attribute when vlan.vlan_group was not a VLANGroup object.

jdrew82:
> > Is this for cases where a VLAN is added outside the Infoblox import? If I remember correctly, Infoblox associates all VLANs to a VLANView, which we import as VLANGroups.
> 
> TBH, I don't remember how I triggered this, but I did at some point hit an AttributeError for trying to access the `name` attribute when vlan.vlan_group was not a VLANGroup object.

Interesting. The only way I can see that happening is if the VLAN and VLANGroup didn't get associated for some reason or a VLAN was added manually without associating to a VLANGroup. EIther way it makes sense to account for this edge case.

jdrew82 (approved):
LGTM!

PR #333, opened 2024-01-19
Allow Exceptions To Fail Job

jdrew82:
This PR addresses the issue brought up in #329.

As we now need to allow Exceptions to be thrown to mark a Job as failed we need to remove the try/except block. The rollback also isn't a concern as the atomic transaction wrapper was removed in Nautobot 2.0.

### DIFF
--- a/nautobot_ssot/jobs/base.py
+++ b/nautobot_ssot/jobs/base.py
@@ -1,7 +1,6 @@
 """Base Job classes for sync workers."""
 from collections import namedtuple
 from datetime import datetime
-import traceback
 import tracemalloc
 from typing import Iterable, Optional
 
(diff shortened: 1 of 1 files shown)


Kircheneer:
Nice, I like this change. Makes everything a good bit simpler.

Kircheneer (approved):

ubuntu811:
I'm still torn if it would be better to log and re-raise the exception instead, because having the exception in the log helps to see whats going wrong. Functionally both versions are fine.

Kircheneer:
With how this is set up I _think_ that the exception should go right into the log. Is this not the case?

ubuntu811:
> With how this is set up I _think_ that the exception should go right into the log. Is this not the case?

At least in the SSOT job logs you don't. That's why I ended up changing my private debug version back to logging the stacktrace and re-raising the exception.

Kircheneer:
@jdrew82 in that case I like the approach better where we log the exception. Can you make this possible?

jdrew82:
> @jdrew82 in that case I like the approach better where we log the exception. Can you make this possible?

I need to test this but I'm fairly certain that an Exception being raised will be displayed in the Job logs, it just wasn't marking the Job as failed.

ubuntu811:
When are you guys expecting to merge this to develop?

jdrew82:
> When are you guys expecting to merge this to develop?

I still need to confirm this works as expected. I'll test it today and then we should be able to merge.

jdrew82:
Confirmed this works as expected so merging.

PR #339, opened 2024-01-25
Fix ACI LocationType Bug

jdrew82:
There's a bug in ACI where it's performing a `get_or_create` for LocationType of Site when trying to get a Location but this is wrong as `get_or_create` returns a tuple when we only want the first item from the tuple. As the Site LocationType is created in signals we should be able to just do a get for it.

This also fixes the calls that were using `get_identifiers()` which is unnecessary as that var is on the self variable.


chadell (approved):
LGTM

PR #341, opened 2024-01-27
Wrap IPFabric's database calls around try/except

jmcgill298:
Wrap the majority of database calls around a try/except to prevent job failure from individual object failure, and add logging about failures


alhogan-aa (approved):
LGTM! Thank you for your contribution!

alhogan (approved):
LGTM!

Issue #323, opened 2024-01-12
"KeyError: infoblox_network_view"


Bahu2016:
### Environment
* Python version:  3.11
* Nautobot version:  2.0.3
* nautobot-ssot version:  latest

"KeyError: infoblox_network_view"
https://github.com/nautobot/nautobot-app-ssot/blob/af53c81384ff171c222668666fe696fc27227529/nautobot_ssot/integrations/infoblox/constant.py#L15C23-L15C23
it looks like the constants were updated recently and we didn’t catch that; that should be a get on config

Temporary fix:
fix it temporarily by adding that key to your PLUGINS_CONFIG definition for nautobot_ssot
"infoblox_network_view": os.getenv("NAUTOBOT_SSOT_INFOBLOX_NETWORK_VIEW", ""),

Thank you JDonga who suggested fix from slack community

PR #348, opened 2024-02-01
Fix Infoblox Config Bug

jdrew82:
This PR should fix the issue detailed in #323. While I was fixing that the tests also turned up some issues with IPFabric integration that I've fixed.


qduk (approved):
LGTM

PR #360, opened 2024-02-20
Contrib Updates Part 1

Renrut5:
First step in suggested changes for contrib files. Goal is to make a number of changes to improve usability based on my experiences with contrib but without changing current functionality.


nautobot_ssot/contrib/utils.py
```diff
@@ -0,0 +1,73 @@
+"""Supporting classes for SSoT contib interfacing with Nautobot."""
```
Kircheneer, line 1:
```suggestion
"""Supporting classes for SSoT contrib interfacing with Nautobot."""
```

nautobot_ssot/contrib/utils.py
```diff
@@ -0,0 +1,73 @@
...
+
+from enum import Enum
+from dataclasses import dataclass
+
+
+class RelationshipSideEnum(Enum):
+    """This details which side of a custom relationship the model it's defined on is on."""
+
```
Kircheneer, line 9: Can we rename this file to `types.py`?

Kircheneer (changes requested):
I like it! Just two minor changes and a bit of CI and then we're good to go.

Renrut5:
@Kircheneer , got the requested changes made, once the pipeline finishes should be good to merge

jdrew82 (approved):
LGTM!

PR #362, opened 2024-02-21
use typing.get_type_hints everywhere in favor of __annotations__

Kircheneer:
I had an SSoT job break because at least in Python 3.11 `Klass.__annotations__` does not show annotations of parent classes, whereas in 3.8 it does. This PR changes that behaviour to be uniform across all Python versions.


jdrew82 (approved):
LGTM!

jdrew82 (approved):
LGTM!

PR #368, opened 2024-02-26
Import Optional from typing directly

jmcgill298:
Optional has always existed in the typing module, so there is no need to import from typing_extensions, which has only recently added it to the list of importable objects.


jdrew82:
LGTM!

jdrew82 (approved):
LGTM!

PR #371, opened 2024-02-27
change contrib.NautobotModel.get_from_db to use a PK

Kircheneer:
This drastically simplifies the method implementation and also makes the PK available for other uses where needed. It also is very cheap, because we load all objects already anyway.


jdrew82 (approved):
LGTM!

PR #374, opened 2024-02-29
improve error handling in contrib

Kircheneer:
This PR changes `nautobot_ssot.contrib.NautobotModel` such that when an error occurs in `create`, `update` or `delete` it raises the appropriate diffsync exception class so that it can be used properly with the `CONTINUE_ON_FAILURE` global diffsync flag: https://diffsync.readthedocs.io/en/latest/core_engine/index.html#supported-global-flags

Furthermore, error handling was added in a lot of places to make diagnosis of errors easier.


chadell (approved):
LGTM

PR #377, opened 2024-03-05
Allow foreign keys inside of many to many relationships.

### LINKED ISSUE #295: TypedDict fields should support lookups that span relationships
### Environment
* Nautobot version:  1.6.6
* nautobot-ssot version:  1.6.0

### Proposed Functionality
TypedDict is used to uniquely reference entities in many-to-many relationships. This works fine unless some of the entity fields necessary to make the reference unique represent an object.
TypedDict fields should support lookups that span relationships

### Use Case
It should be possible to create TypedDicts like these
```
class VLANDict(TypedDict):
    vid: int
    site__slug: Optional[str]
class IPAddressDict(TypedDict):
    host: str
    prefix_length: int
    vrf__name: Optional[str] 
```
but this produces errors like  `AttributeError: 'VLAN' object has no attribute 'site__slug'` and `AttributeError: 'IPAddress' object has no attribute 'vrf__name'`

### Proposed implementation
```
contrib.py	2023-12-01 16:15:14
@@ -203,7 +203,7 @@
         # TODO: Allow for filtering, i.e. not taking into account all the objects behind the relationship.
         for related_object in getattr(database_object, parameter_name).all():
             dictionary_representation = {
-                field_name: getattr(related_object, field_name) for field_name in inner_type.__annotations__
+                field_name: NautobotAdapter._handle_foreign_key(related_object, field_name) if "__" in field_name else getattr(related_object, field_name) for field_name in inner_type.__annotations__
             }
             # Only use those where there is a single field defined, all 'None's will not help us.
             if any(dictionary_representation.values()):

```

Kircheneer:
Closes #295

### DIFF
--- a/changes/377.fixed
+++ b/changes/377.fixed
@@ -1,0 +1,1 @@
+Allow foreign keys inside of many to many relationships.
--- a/nautobot_ssot/contrib/adapter.py
+++ b/nautobot_ssot/contrib/adapter.py
@@ -215,13 +215,28 @@ class NautobotAdapter(DiffSync):
             related_object = getattr(
                 association, "source" if annotation.side == RelationshipSideEnum.DESTINATION else "destination"
             )
-            dictionary_representation = {
-                field_name: getattr(related_object, field_name) for field_name in get_type_hints(inner_type)
-            }
+            dictionary_representation = self._handle_typed_dict(inner_type, related_object)
             # Only use those where there is a single field defined, all 'None's will not help us.
             if any(dictionary_representation.values()):
                 related_objects_list.append(dictionary_representation)
         return related_objects_list
+
+    @classmethod
+    def _handle_typed_dict(cls, inner_type, related_object):
+        """Handle a typed dict for many to many relationships.
+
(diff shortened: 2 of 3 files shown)


Kircheneer:
@gioccher can you share your [coauthor information](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors#required-co-author-information)? I'm trying to credit you for the fix, but `@Co-Authored-By: Claudio Beretta <gioccher@users.noreply.github.com>` doesn't seem to be working.

gioccher:
> @gioccher can you share your [coauthor information](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors#required-co-author-information)? I'm trying to credit you for the fix, but `@Co-Authored-By: Claudio Beretta <gioccher@users.noreply.github.com>` doesn't seem to be working.

oh thanks, but co-authorship is definitely not needed

PR #379, opened 2024-03-07
Add LCM plugin config. Fix 'infoblox_import_subnets` setting.

progala:
Updates to dev `nautobot_config.py`:

- Adds config for Nautobot Device Lifecycle Management plugin.
- Fixes logic in `infoblox_import_subnets` setting.


jdrew82:
So I'm a bit confused here. You're creating a list it looks like, but shouldn't that automatically be done if the value is a concatenated string with the split()?

progala:
Problem here is that the default value `""` will result in `"".split(",")`. This does not produce an empty list which breaks logic in https://github.com/nautobot/nautobot-app-ssot/blob/develop/nautobot_ssot/integrations/infoblox/diffsync/adapters/infoblox.py#L51

```
>>> "".split(",")
['']
```

jdrew82:
And you've got the same result with your suggestion:

```
>>> [x for x in "".split(",") if x]
[]
```

progala:
It's not the same. One is a 1-element list with an empty string `['']`, the other is a 0-element list `[]`. `bool([''])` evaluates to `True`, `bool([])` evaluates to `False`.

jdrew82:
Fair point. Looks like it should work right with a passed string as typically expected too so I'm fine with merging this. Can we also get the documentation updated to match?

jdrew82 (approved):
LGTM!

PR #390, opened 2024-03-13
use typing.get_args in contrib in favor of accessing __args__

Kircheneer:
This should provide a more consistent and bug-free experience as we are directly utilizing a standard library API.


chadell (approved):
LGTM

Issue #352, opened 2024-02-06
Custom 1-to-Many Relationships Error when defining Destination side


Renrut5:
### Environment
* Python version:  3.8
* Nautobot version:  2.1.2
* nautobot-ssot version:  2.2.0

### Expected Behavior
SSoT to create custom one to many relationships when defining the relationship from the destination side.

### Observed Behavior
An error occurs as the returned value from `_handle_custom_relationship_to_many_relationship` method is always a list of objects regardless if it is 1-to-many or many-to-many.

When `RelationshipSideEnum.DESTINATION` is used on 1-to-many relationships, a single object is expected, not a list of objects.

### Steps to Reproduce
1. Define 1-to-many custom relationships in SSoT and build relationships from destination side

Kircheneer:
Can you share some code how we can reproduce this?

Kircheneer:
Is this the same as #392 ?

chadell:
I have a fix, going to contribute it

Renrut5:
@chadell @Kircheneer , is this considered completed from the merge for #393 ?

Kircheneer:
IMO this can be closed, yes. Feel free to re-open in case you encounter any issues. Thanks!

PR #393, opened 2024-03-14
fix: handle custom 1-to-many relationships

chadell:
Will cover #352 

TODO:

- [x] Add tests
- [x] Raise proper exceptions and clean code

### DIFF
--- a/nautobot_ssot/contrib/model.py
+++ b/nautobot_ssot/contrib/model.py
@@ -14,6 +14,8 @@ from django.contrib.contenttypes.models import ContentType
 from django.core.exceptions import ValidationError, MultipleObjectsReturned
 from django.db.models import Model, ProtectedError
 from nautobot.extras.models import Relationship, RelationshipAssociation
+from nautobot.extras.choices import RelationshipTypeChoices
+
 from typing_extensions import get_type_hints
 from nautobot_ssot.contrib.types import (
     CustomFieldAnnotation,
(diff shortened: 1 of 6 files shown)


nautobot_ssot/contrib/model.py
```diff
@@ -13,6 +14,8 @@
 from django.core.exceptions import ValidationError, MultipleObjectsReturned
 from django.db.models import Model, ProtectedError
 from nautobot.extras.models import Relationship, RelationshipAssociation
+from nautobot.extras.choices import RelationshipTypeChoices
+
```
Kircheneer, line 18:
```suggestion
```

nautobot_ssot/contrib/model.py
```diff
@@ -168,10 +171,18 @@ def _handle_single_field(
             else:
                 related_object_content_type = relationship.destination_type
             related_model_class = related_object_content_type.model_class()
-            relationship_fields["custom_relationship_many_to_many_fields"][field] = {
-                "objects": [diffsync.get_from_orm_cache(parameters, related_model_class) for parameters in value],
-                "annotation": custom_relationship_annotation,
-            }
+
+            if relationship.type == RelationshipTypeChoices.TYPE_ONE_TO_MANY:
+                relationship_fields["custom_relationship_foreign_keys"][related_model_class.__name__] = {
```
Kircheneer, line 176:
```suggestion
                relationship_fields["custom_relationship_foreign_keys"][field] = {
```

I think even though this is not used we should keep using the field name on the Pydantic object as the key, what do you think?

Kircheneer (changes requested):
Two minor things, if you add tests it should be good to go.

PR #398, opened 2024-03-19
External Integrations

snaselj:
# Closes: NaN

## What's Changed

- Added external integration to example jobs.
- Added external integration to Arista Cloud Vision jobs.


nautobot_ssot/jobs/examples.py
```diff
@@ -378,6 +380,12 @@ def load(self):
 class ExampleDataSource(DataSource):
     """Sync Region and Site data from a remote Nautobot instance into the local Nautobot instance."""
 
+    source = ObjectVar(
```
snaselj, line 383: Is this the right way for other integrations as well?

@cmsirbu @jdrew82

snaselj:
Added ExternalIntegration to AristaCV jobs. Could you pls. check it? @cmsirbu @jdrew82

jdrew82:
> Added ExternalIntegration to AristaCV jobs. Could you pls. check it? @cmsirbu @jdrew82

I looked at what you'd done and good idea with the config information but I'm not sure how that'd work for the configuration information though as that's technically static in most cases and without the ExternalIntegration being defined at time of Job run I'd be curious how it'd show data. Have you tested that bit?

snaselj:
> > Added ExternalIntegration to AristaCV jobs. Could you pls. check it? @cmsirbu @jdrew82
> 
> I looked at what you'd done and good idea with the config information but I'm not sure how that'd work for the configuration information though as that's technically static in most cases and without the ExternalIntegration being defined at time of Job run I'd be curious how it'd show data. Have you tested that bit?

Without ExternalIntegration defined, this should work as before. If it's defined and specified as a job input argument, it takes precedence over PLUGIN_CONFIG values. Hope this is what was requested @cmsirbu 

Have tested without real job run, just to create an integration, and call the job with it. It reads the data as expected. Would be great if someone with access to Arista CVP could test this.

cmsirbu:
The general idea was to avoid making this a breaking change - which means that if someone upgrades to the next release of SSoT and uses the PLUGIN_CONFIG method, their setup will continue working without any change.

jdrew82:
> > > Added ExternalIntegration to AristaCV jobs. Could you pls. check it? @cmsirbu @jdrew82
> > 
> > 
> > I looked at what you'd done and good idea with the config information but I'm not sure how that'd work for the configuration information though as that's technically static in most cases and without the ExternalIntegration being defined at time of Job run I'd be curious how it'd show data. Have you tested that bit?
> 
> Without ExternalIntegration defined, this should work as before. If it's defined and specified as a job input argument, it takes precedence over PLUGIN_CONFIG values. Hope this is what was requested @cmsirbu
> 
> Have tested without real job run, just to create an integration, and call the job with it. It reads the data as expected. Would be great if someone with access to Arista CVP could test this.

I'm talking about the Config Information that shows on the DataSource/DataTarget pages. I'm pretty sure that's read at App load and isn't able to be dynamically updated by the ExternalIntegration selection.

snaselj:
> > > > Added ExternalIntegration to AristaCV jobs. Could you pls. check it? @cmsirbu @jdrew82
> > > 
> > > 
> > > I looked at what you'd done and good idea with the config information but I'm not sure how that'd work for the configuration information though as that's technically static in most cases and without the ExternalIntegration being defined at time of Job run I'd be curious how it'd show data. Have you tested that bit?
> > 
> > 
> > Without ExternalIntegration defined, this should work as before. If it's defined and specified as a job input argument, it takes precedence over PLUGIN_CONFIG values. Hope this is what was requested @cmsirbu
> > Have tested without real job run, just to create an integration, and call the job with it. It reads the data as expected. Would be great if someone with access to Arista CVP could test this.
> 
> I'm talking about the Config Information that shows on the DataSource/DataTarget pages. I'm pretty sure that's read at App load and isn't able to be dynamically updated by the ExternalIntegration selection.

I've refactored, how the configuration is gathered for CloudVision. There is a new optional key in `PLUGIN_CONFIG["nautobot_ssot"]["aristacv_external_integration_name"]` that can be used to specify integration to use.

Kircheneer:
Can you provide an invoke command for the script you mentioned in your PR description? I think this would aid development in the future

snaselj:
> Can you provide an invoke command for the script you mentioned in your PR description? I think this would aid development in the future

Would rather do this in a separate PR not to block this one.

nautobot_ssot/integrations/aristacv/utils/nautobot.py
```diff
@@ -26,20 +210,20 @@ def verify_site(site_name):
     try:
         site_obj = Location.objects.get(name=site_name, location_type=loc_type)
     except Location.DoesNotExist:
-        site_obj = Location(
+        status, _ = Status.objects.get_or_create(name="Staging")
```
jdrew82, line 213: We should probably be checking if this was created and adding ContentTypes to the Location object. Although it should exist since it's there by default it's possible a User removes it for some reason and breaks this part.

jdrew82 (changes requested):
Few things to address but this looks good in general.

nautobot_ssot/integrations/aristacv/utils/nautobot.py
snaselj: Could you pls. add a suggestion? Adding content type to the location type is on line 209.

jdrew82 (changes requested):

nautobot_ssot/integrations/aristacv/utils/nautobot.py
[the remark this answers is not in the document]
jdrew82: @snaselj this is what I meant

jdrew82 (approved):
LGTM!

Issue #367, opened 2024-02-26
Example jobs fail to run, expecting wrong data structure in source system


JonathanNathanson:
### Environment
* Python version:  3.9
* Nautobot version: 2.1.5
* nautobot-ssot version:  2.3.0

### Expected Behavior
For the example job "Example Data Source" to sync against the default parameters (https://demo.nautobot.com) without producing errors.

### Observed Behavior

The job fails to run with a return value of:

`{'exc_type': 'KeyError', 'exc_module': 'builtins', 'exc_message': ['label']}`

The traceback looks like this, indicating that the job is expecting a different data format to that which is being returned via the source system's API:

```
celery_worker-1    | [2024-02-23 11:11:57,456: ERROR/ForkPoolWorker-7] Task nautobot_ssot.jobs.examples.ExampleDataSource[342f7259-0c37-4c8c-879c-e15a6515503c] raised unexpected: KeyError('label')
celery_worker-1    | Traceback (most recent call last):
celery_worker-1    |   File "/usr/local/lib/python3.9/site-packages/celery/app/trace.py", line 477, in trace_task
celery_worker-1    |     R = retval = fun(*args, **kwargs)
celery_worker-1    |   File "/usr/local/lib/python3.9/site-packages/nautobot/extras/jobs.py", line 153, in __call__
celery_worker-1    |     return self.run(*args, **deserialized_kwargs)
celery_worker-1    |   File "/opt/nautobot/.local/lib/python3.9/site-packages/nautobot_ssot/jobs/examples.py", line 570, in run
celery_worker-1    |     super().run(dryrun, memory_profiling, *args, **kwargs)
celery_worker-1    |   File "/opt/nautobot/.local/lib/python3.9/site-packages/nautobot_ssot/jobs/base.py", line 316, in run
celery_worker-1    |     self.sync_data(memory_profiling)
celery_worker-1    |   File "/opt/nautobot/.local/lib/python3.9/site-packages/nautobot_ssot/jobs/base.py", line 135, in sync_data
celery_worker-1    |     self.load_source_adapter()
celery_worker-1    |   File "/opt/nautobot/.local/lib/python3.9/site-packages/nautobot_ssot/jobs/examples.py", line 575, in load_source_adapter
celery_worker-1    |     self.source_adapter.load()
celery_worker-1    |   File "/opt/nautobot/.local/lib/python3.9/site-packages/nautobot_ssot/jobs/examples.py", line 439, in load
[...]

PR #405, opened 2024-03-25
Fix example jobs

jdrew82:
This PR addresses the issues in #367 and updates the example Jobs to support handling Roles, Platforms, Manufacturers, DeviceTypes, and Devices.

### DIFF
--- a/nautobot_ssot/jobs/examples.py
+++ b/nautobot_ssot/jobs/examples.py
@@ -475,7 +475,7 @@ class NautobotRemote(DiffSync):
                 self.add(devicetype)
                 manufacturer.add_child(devicetype)
             except ObjectNotFound:
-                self.logger.debug(f"Unable to find Manufacturer {device_type['manufacturer']['name']}")
+                self.job.logger.debug(f"Unable to find Manufacturer {device_type['manufacturer']['name']}")
 
         for platform in self._get_api_data("api/dcim/platforms/?depth=1"):
             platform = self.platform(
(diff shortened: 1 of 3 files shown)


nautobot_ssot/jobs/examples.py
```diff
@@ -94,14 +126,100 @@ class TenantModel(NautobotModel):
...
+    _attributes = ("part_number", "u_height", "is_full_depth")
+
+    model: str
+    manufacturer__name: str
+    part_number: str
+    u_height: int
+    is_full_depth: bool
+
+    # Not in _attributes or _identifiers, hence not included in diff calculations
+    pk: Optional[UUID]
```
Kircheneer, line 152: With the newest iteration of contrib, you get these out of the box, no need to explicitly annotate them - pk should be available on all `NautobotModel`s

Kircheneer (changes requested):
Just one minor thing that I'd like to get fixed in case people use these as examples (huh 😄). Thanks for taking care of this!

nautobot_ssot/jobs/examples.py
jdrew82: That's what I thought but it appeared to be used when communicating to the remote Nautobot instance.

Kircheneer: Ahhh, yeah got it

Kircheneer (approved):
LGTM, merge if you want :)

PR #413, opened 2024-04-02
Fix IPFabric technology calls to us `all`

jmcgill298:
Previously these were using `fetch`, which does not collect all results. Switching this to use `all` will get all relevant data points.


alhogan (approved):
LGTM

PR #414, opened 2024-04-03
Improve IPFabric matching of InterfaceTypes

jmcgill298:
Adds regex matching of interface names for when IPFabric does not provide a media_type.


jmcgill298:
The failed test case is unrelated and fixed in #413

alhogan (approved):

jdrew82 (approved):
LGTM!

PR #421, opened 2024-04-12
opens prometheus-client dependency range and removes direkt drf-spect…

Kircheneer:
…acular dependency.

This should fix the upstream CI that's currently failing: https://github.com/nautobot/nautobot-app-ssot/actions/runs/8641403014/job/23690873680


chadell (approved):
LGTM

glennmatthews (approved):

PR #435, opened 2024-04-18
Fix DLM App RuntimeError

jdrew82:
This PR is to catch the RuntimeError Exception being thrown if DLM App isn't enabled in the PLUGINS list.


nautobot_ssot/integrations/aristacv/diffsync/models/nautobot.py
```diff
@@ -35,7 +36,7 @@
     from nautobot_device_lifecycle_mgmt.models import SoftwareLCM
 
     LIFECYCLE_MGMT = True
-except ImportError:
+except (ImportError, RuntimeError):
     print("Device Lifecycle app isn't installed so will revert to CustomField for OS version.")
```
gsnider2195, line 40: Maybe we should have another `except RuntimeError:` block that logs a warning? Something like:
```py
logger.warn("nautobot-device-lifecycle-mgmt is installed but not enabled. Did you forget to add it to your settings.PLUGINS?")
logger.info("Device Lifecycle app isn't installed so will revert to CustomField for OS version.")
```

gsnider2195 (approved):

nautobot_ssot/integrations/aristacv/diffsync/models/nautobot.py
jdrew82: Done!

PR #436, opened 2024-04-23
Device42 Unit Tests

jdrew82:
This PR is to add unit tests to the Device42 integration.

### DIFF
--- a/nautobot_ssot/tables.py
+++ b/nautobot_ssot/tables.py
@@ -36,7 +36,7 @@ DRY_RUN_LABEL = """
 MESSAGE_SPAN = """<span class="message">{% if record.message %}{{ record.message }}{% else %}—{% endif %}</span>"""
 
 
-class DashboardTable(BaseTable):
+class DashboardTable(BaseTable):  # pylint: disable=nb-sub-class-name
     """Abbreviated version of SyncTable, for use with the dashboard."""
 
     start_time = DateTimeColumn(linkify=True, short=True)
@@ -49,7 +49,7 @@ class DashboardTable(BaseTable):
         """Metaclass attributes of DashboardTable."""
 
         model = Sync
-        fields = ["source", "target", "start_time", "status", "dry_run"]
+        fields = ["source", "target", "start_time", "status", "dry_run"]  # pylint: disable=nb-use-fields-all
         order_by = ["-start_time"]
 
 
@@ -105,7 +105,7 @@ class SyncTable(BaseTable):
         """Metaclass attributes of SyncTable."""
 
         model = Sync
-        fields = (
+        fields = (  # pylint: disable=nb-use-fields-all
             "pk",
             "source",
             "target",
@@ -138,7 +138,7 @@ class SyncTable(BaseTable):
         order_by = ("-start_time",)
 
 
-class SyncTableSingleSourceOrTarget(SyncTable):
+class SyncTableSingleSourceOrTarget(SyncTable):  # pylint: disable=nb-no-model-found
     """Subclass of SyncTable with fewer default columns."""
 
     class Meta(SyncTable.Meta):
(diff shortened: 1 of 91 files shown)


nautobot_ssot/tables.py
```diff
@@ -104,7 +105,7 @@ class Meta(BaseTable.Meta):
         """Metaclass attributes of SyncTable."""
 
         model = Sync
-        fields = (
+        fields = (  # pylint: disable=nb-use-fields-all
```
Kircheneer, line 108: Any reason why we can't use `__all__` here and in the next two instances?

nautobot_ssot/tests/device42/unit/test_models_nautobot_ipam.py
```diff
@@ -0,0 +1,511 @@
...
+from nautobot.extras.models import Role, Status
+from nautobot.ipam.models import IPAddress, IPAddressToInterface, Namespace, Prefix, VLAN, VRF
+from nautobot_ssot.integrations.device42.diffsync.models.nautobot import ipam
+
+
+class TestNautobotVRFGroup(TransactionTestCase):
+    """Test the NautobotVRFGroup class."""
+
+    def setUp(self):
+        self.diffsync = DiffSync()
```
Kircheneer, line 18: What do you think about using the adapter directly here instead of the bare `DiffSync`?

nautobot_ssot/tests/device42/unit/test_models_nautobot_ipam.py
```diff
@@ -0,0 +1,511 @@
...
+class TestNautobotSubnet(TransactionTestCase):
+    """Test the NautobotSubnet class."""
+
+    def setUp(self):
+        super().setUp()
+        self.status_active = Status.objects.get(name="Active")
+        self.test_ns = Namespace.objects.get_or_create(name="Test")[0]
+        self.test_vrf = VRF.objects.get_or_create(name="Test", namespace=self.test_ns)[0]
+        self.prefix = Prefix.objects.create(prefix="10.0.0.0/24", namespace=self.test_ns, status=self.status_active)
+        self.diffsync = DiffSync()
```
Kircheneer, line 106: This one's similar to the above comment

Kircheneer (changes requested):

nautobot_ssot/tables.py
jdrew82: I believe this pylint rule is wrong. I've tried to use `fields = "__all__"` and also `fields = ["__all__"]` and both return errors or the pylint error. Using the list version just returns the pylint error and trying to use just the string throws a TypeError:

`TypeError: DashboardTable.fields = '__all__' (type str), but type must be one of (tuple, list, set)`.

nautobot_ssot/tests/device42/unit/test_models_nautobot_ipam.py
jdrew82: Using the direct adapter requires a lot more mocking and setup that isn't used. As I'm just testing the CRUD functions of the models directly I didn't think using the Device42Adapter would be necessary.

nautobot_ssot/tests/device42/unit/test_models_nautobot_ipam.py
jdrew82: Same answer.

nautobot_ssot/tables.py
glennmatthews: yes - see https://github.com/nautobot/pylint-nautobot/pull/83. Not sure when the next release will include this fix.

jdrew82: > yes - see [nautobot/pylint-nautobot#83](https://github.com/nautobot/pylint-nautobot/pull/83). Not sure when the next release will include this fix.

Thanks for confirming that the rule is broken @glennmatthews!

Kircheneer (approved):

PR #438, opened 2024-04-25
[Infoblox] Replace `requests.request` with `requests.Session` in the Infoblox API client.

progala:
Changed the Infoblox `utils.client` to make API calls using `requests.Session` instead of `requests.request`.

This change should result in 1.5 - 3x performance improvement when loading data from Infoblox.


jdrew82 (approved):
LGTM!

Issue #443, opened 2024-04-30
Loading the IPAddress if Object Already Exists


Bahu2016:
### Environment
* Python version:  3.11
* Nautobot version:  2.1.8
* nautobot-ssot version:  2.6.1

It should skip or update if object exists

Warning Duplicate prefix found: x.x.x.x/x. Duplicate prefixes are not supported and only the first occurence will be included in the sync

[2024-04-30 02:00:30, 275: ERROR/ForkPoolWorker-1] Task nautobot_ssot.integrations.infoblox.jobs.InfobloxDataSource[333b343-34343-34343-34343] raised unexpected: ObjectAlreadyExists('Object x.x.x.x__x.x.x.x/x__x already present, ipaddress "x.x.x.x__x.x.x.x/x__x"

### Steps to Reproduce
1. sync the subnet if already exists
2.
3.

Issue #177, opened 2022-11-30
Clean Up Logging


jdrew82:
### Environment
* Python version:  N/A
* Nautobot version:  N/A
* nautobot-ssot-infoblox version:  0.7.4

### Expected Behavior
Logging of information to be done in a reasonable manner. Ie, keep informational logging to just necessary information so the user knows where in the process the Job is and trim out unnecessary logging and only necessary debug logging.

### Observed Behavior
Logging isn't consistent and missing information that could aide the user in knowing what is happening.

PR #444, opened 2024-04-30
Fix Duplicate IPAddress Loading in Infoblox

jdrew82:
This PR is to address the bug mentioned in #443. It also addresses the logging issue in #177.


qduk (approved):
LGTM!

jdrew82:
Confirmed working by community member so merging.

PR #446, opened 2024-05-02
[WIP] DNA Center Integration

jdrew82:
This PR is to add a SSoT integration for Cisco DNA Center.


nautobot_ssot/integrations/dna_center/constants.py
```diff
@@ -0,0 +1,33 @@
+"""Defines various constants used throughout the App."""
+
+from django.conf import settings
+
+# Import config vars from nautobot_config.py
+PLUGIN_CFG = settings.PLUGINS_CONFIG["nautobot_ssot"]
+
+DNAC_PLATFORM_MAPPER = {
```
itdependsnetworks, line 8: Can we add to netutils and use that here?

nautobot_ssot/integrations/dna_center/jobs.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+name = "DNA Center SSoT"  # pylint: disable=invalid-name
+
+
+class DnaCenterDataSource(DataSource):  # pylint: disable=too-many-instance-attributes
+    """DNA Center SSoT Data Source."""
+
+    dnac = ObjectVar(
+        model=ExternalIntegration,
```
chadell, line 22: awesome!
what about using the Controller model instead? (and accessing its ExternalIntegration)

nautobot_ssot/integrations/dna_center/constants.py
jdrew82: Sure, I can look into getting this added to netutils. When I made this I didn't think it was something worth putting in netutils but I can see the value if we add it to the network_driver mapping.

PR #448, opened 2024-05-04
Fixed link from list view to filtered sync log view by changing filte…

itdependsnetworks:
…r query to `sync` from overview.


jdrew82 (approved):
LGTM!

PR #451, opened 2024-05-06
[WIP] Add DNA Center Integration

jdrew82:
This PR is to add the relevant parts for a DNA Center integration. Still has some bits added before it's ready for full review.


jdrew82:
@chadell So we don't forget about the comment in the other PR:

https://github.com/nautobot/nautobot-app-ssot/pull/446#discussion_r1588691530

How would you like to address supporting the Controller model and/or the ExternalIntegration model as one requires 2.1 and the other 2.2?

Issue #455, opened 2024-05-13
CI Failing on Wrong BIG-IP Netmiko Platform


jdrew82:
### Environment
* Python version:  3.11
* Nautobot version:  2.2.3
* nautobot-ssot version:  2.6.1

### Expected Behavior
CI to pass for test_get_netmiko_platform_5_bigip unit test.

### Observed Behavior
Test fails with AssertionError:

```
======================================================================
FAIL: test_get_netmiko_platform_5_bigip (nautobot_ssot.tests.device42.unit.test_utils_device42.TestUtilsDevice42.test_get_netmiko_platform_5_bigip)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/local/lib/python3.11/site-packages/parameterized/parameterized.py", line 533, in standalone_func
    return func(*(a + p.args), **p.kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/source/nautobot_ssot/tests/device42/unit/test_utils_device42.py", line 174, in test_get_netmiko_platform
    self.assertEqual(device42.get_netmiko_platform(sent), received)
AssertionError: 'bigip_f5' != 'f5_tmsh'
- bigip_f5
+ f5_tmsh
```

### Steps to Reproduce
1. Build dev environment for SSoT.
2. Run `invoke tests`.
3. Look at results.

jdrew82:
I believe this is related to a change in the PYATS_LIB_MAPPER mapping in netutils.

PR #456, opened 2024-05-13
Fix CI for Device42 Test

jdrew82:
This PR is to fix the issue discussed in #455.


jdrew82:
> ~Should this not be changed to the correct string instead of removing?~ Nevermind, I see that Netmiko does not support F5/BigIP.

My first attempt was to correct it but for some reason CI wasn't picking it up. Technically the other tests already validate functionality so I don't think this one in particular is really required.

gsnider2195 (approved):

PR #469, opened 2024-06-13
[WIP] Update Example Jobs

jdrew82:
This PR is to update the example Jobs to include more of the standard object types most want to import with SSoT including:

- Interfaces
- IPAddress
- Prefix
- Namespace

Also added support for Locations on Prefixes.


nautobot_ssot/jobs/examples.py
```diff
@@ -99,6 +106,39 @@ class RoleModel(NautobotModel):
...
+    _modelname = "status"
+    _identifiers = ("name",)
+    _attributes = ("content_types", "color")
+
+    name: str
+    color: str
+    content_types: List[ContentTypeDict] = []
+
+    # Not in _attributes or _identifiers, hence not included in diff calculations
+    pk: Optional[UUID]
```
Kircheneer, line 123: This doesn't need to be explicitly defined, its already in `NautobotModel`

nautobot_ssot/jobs/examples.py
```diff
@@ -107,19 +147,48 @@ class PrefixModel(NautobotModel):
...
 
     # Data type declarations for all identifiers and attributes
     network: str
+    namespace__name: str
     prefix_length: int
     tenant__name: Optional[str]
     status__name: str
     description: str
 
+    locations: Optional[List[LocationDict]] = []
```
Kircheneer, line 160:
```suggestion
    locations: List[LocationDict] = []
```

I think if you default to an empty list it isn't necessary to accept `None` as well, right?

nautobot_ssot/jobs/examples.py
jdrew82: Alright, will remove. I feel like those were there prior to my updates but no worries.

nautobot_ssot/jobs/examples.py
jdrew82: Fair point, it does seem to be a [] by default in the API response so will remove Optional.

nautobot_ssot/jobs/examples.py
Kircheneer: Yes they were - but since you added some I thought I'd mention it

jdrew82: Fair enough. Was just replicating the others. I've removed it from all the models though.

nautobot_ssot/jobs/examples.py
```diff
@@ -79,9 +87,6 @@ class LocationModel(NautobotModel):
     tenant__name: Optional[str]
     description: str
 
-    # Not in _attributes or _identifiers, hence not included in diff calculations
-    pk: Optional[UUID]
```
qduk, line 83: Are you removing this for a particular reason?  I only ask because I know I've questioned how/where to use a non diff'd attribute in my models and having one in the example might be helpful?

qduk: Oh I just saw your conversation with Leo.

qduk (approved):

nautobot_ssot/jobs/examples.py
jdrew82: Agreed on that front. I think the documentation update that @Kircheneer is doing will illustrate that bit too.

PR #481, opened 2024-07-12
Fix issue with markdown dependency

collin-wicker:
Nautobot would not build with 'Markdown = "!=3.3.6"'. This issue was resolved on my machine with the included code.


pyproject.toml
```diff
@@ -32,7 +32,7 @@ python = ">=3.8,<3.12"
 nautobot = "^2.1.0"
 diffsync = "^1.6.0"
 Jinja2 = { version = ">=2.11.3", optional = true }
-Markdown = "!=3.3.5"
+Markdown = ">=3.3.7,<=3.4.0"
```
glennmatthews, line 35: This change will likely prevent installation against newer versions of Nautobot - for example Nautobot 2.2.7 requires `Markdown~3.5.2`.

collin-wicker: Would 'Markdown = ">=3.3.7"' suffice? I just put 'Markdown = ">=3.3.7,<=3.4.0"' because that's what my shell gave me and it works in my environment.

Kircheneer: In my opinion going `>=3.3.7` is good, then if we discover any versions that don't work we can explicitly exclude them.

jdrew82:
I'm going to close this out as it doesn't make sense anymore with Markdown being unbound in pyproject.toml now.

collin-wicker:
> I'm going to close this out as it doesn't make sense anymore with Markdown being unbound in pyproject.toml now.

That makes sense. Thanks. I was wondering about that.

PR #503, opened 2024-08-12
Pagination in Infoblox client

qduk:
The limit set on the `get_all_subnets` of 10,000 was too low in one of my use cases. I implemented pagination for that call. I can incorporate it into the other calls `get_child_subnets_from_container` and maybe `get_vlans` (since the max result is so large). Before I do that, I just wanted to make sure everyone was okay with the way I implemented it and their respective tests.


jdrew82:
Good work! I had difficulty figuring that out in the past so I'm glad you were able to.

jdrew82 (approved):
LGTM!

Issue #504, opened 2024-08-12
Pagination for Infoblox Client


qduk:
### Environment
* Nautobot version:   2.2.6
* nautobot-ssot version:   2.6.1

### Proposed Functionality
Add pagination to Infoblox client so we can 
1. Lower the max_result if wanted/needed
2. Ensure we grab all info from infoblox. 
3. 

### Use Case
I may have over 10k prefixes in my Infoblox. Currently, the subnet call has a `max_result` of 10k.

Issue #497, opened 2024-08-05
`test_create_ip_validation_error_on_first_try` failed due to missing `model` argument for `Status.objects.get_for_model(self, model)`


HanlinMiao:
Test failure in upstream CI:https://github.com/nautobot/nautobot-app-ssot/actions/runs/10225216697/job/28293965021

https://github.com/nautobot/nautobot-app-ssot/blob/develop/nautobot_ssot/tests/ipfabric/test_nbutils.py#L444 and every invocation of `mock_status()` should be changed to `mock_status(IPAddress)` instead.

Issue #491, opened 2024-07-26
[ACI Integration] Improve sync behaviour


mzbroch:
### Environment
* Nautobot version:  
* nautobot-ssot version:  

### Proposed Functionality

1. Set Nautobot namespace to Global for ACI management tenant
2. Set Nautobot tenant for IP, Prefix to ACI BridgeDomain's tenant if found
3. Add tag for Multisite

Issue #488, opened 2024-07-25
Infoblox Setup docs tables aren't rendering properly


qduk:
### Environment
* Python version:  3.12
* Nautobot version:  2.2.4
* nautobot-ssot version:  2.7.0

### Expected Behavior
Infoblox doc tables to render as tables.

### Observed Behavior
Infoblox docs tables render as text

### Steps to Reproduce
1. Look at the docs [here](https://docs.nautobot.com/projects/ssot/en/latest/admin/integrations/infoblox_setup/#prerequisites)
2. Build the docs locally and look at the same page.

PR #510, opened 2024-08-21
Release v2.8.0

jdrew82:
### Added

- [#504](https://github.com/nautobot/nautobot-app-ssot/issues/504) - Added pagination to the `get_all_subnets` Infoblox client call.

### Documentation

- [#488](https://github.com/nautobot/nautobot-app-ssot/issues/488) - Fixed issue with Infoblox setup docs.

### Fixed

- [#491](https://github.com/nautobot/nautobot-app-ssot/issues/491) - Fixed tenant names and introduced tag for multisite.
- [#497](https://github.com/nautobot/nautobot-app-ssot/issues/497) - Fixed IPFabric test failures under Django 4.2.


nautobot_ssot/integrations/aci/signals.py
```diff
@@ -38,6 +38,10 @@ def aci_create_tag(apps, **kwargs):
         name=PLUGIN_CFG.get("tag_down"),
         color=PLUGIN_CFG.get("tag_down_color"),
     )
+    tag.objects.update_or_create(
```
smk4664, line 41: Best practice for update_or_create is to use the `defaults` dictionary key for keys that are not part of the uniqueness.  If someone changed the color in the UI, this would look for the Tag object with a name of "ACI_MULTISITE" and a color of "03a9f4" which would not be found, so it will try to create the Tag instead, which will result in an error as the name is not unique. This in turn would cause post_upgrade to fail.

smk4664 (approved):
A followup issue should be created for the update_or_create calls in signals.py to use `defaults= {"key": "value"}`.

jdrew82:
> A followup issue should be created for the update_or_create calls in signals.py to use `defaults= {"key": "value"}`. 

Absolutely! I thought we'd fixed those but I'll make sure we go back after 3.0 and fix em all.

Issue #507, opened 2024-08-20
Example Data Target Job - AttributeError: 'StringVar' object has no attribute 'startswith'


nathanielfernandez:
### Environment
* Python version:  3.11
* Nautobot version:  2.1.5
* nautobot-ssot version:  2.2.0

Thread in #nautobot: https://networktocode.slack.com/archives/C01NWPK6WHL/p1724054957478579

I wanted to try the example data target job that is provided by default with the nautobot-app-ssot.
I tried to supply the URL and token for target instances like the demo instance and our own nautobot instances. Made sure the proper url with protocol (https://) is provided.

I got an error related to the URL.

`AttributeError: 'StringVar' object has no attribute 'startswith'`

```
  File "/opt/nautobot/lib64/python3.11/site-packages/nautobot_ssot/jobs/examples.py", line 405, in __init__
    if not url.startswith("http"):
           ^^^^^^^^^^^^^^
AttributeError: 'StringVar' object has no attribute 'startswith'
```

I thought there was nothing wrong with the URL I provided. So for troubleshooting, I tried to play with the nautobot-ssot files and removed the if condition to bypass it. I still got an error with the URL.

`requests.exceptions.MissingSchema: Invalid URL '<nautobot.extras.jobs.StringVar object at 0x7fd2b96355d0>/api/dcim/location-types/': No scheme supplied. Perhaps you meant https://<nautobot.extras.jobs.StringVar object at 0x7fd2b96355d0>/api/dcim/location-types/?`

Asked for help in slack, and Justin Drew advised to open an issue.

![image](https://github.com/user-attachments/assets/b991caef-80bc-4115-b6a5-14c19b265cb6)

### Steps to Reproduce
1. Make sure the example jobs are not hidden, and the data target job is also enabled.
2. Run the Example Data Target job, provide the proper URL and token.
3. It would error right away about the URL

Renrut5:
@jdrew82 it looks like this one is completed, in #519 but not closed.

PR #519, opened 2024-08-23
Fix DataTarget Example Job

jdrew82:
This PR closes out the issue in #507.

### DIFF
--- a/changes/507.fixed
+++ b/changes/507.fixed
@@ -1,1 +1,1 @@
-Fix DataTarget example Job to include run() function for using ExternalIntegration or supplied URL and token.
+Fixed DataTarget example Job to include run() function for using ExternalIntegration or supplied URL and token.
--- a/changes/507.fixed
+++ b/changes/507.fixed
@@ -1,0 +1,1 @@
+Fix DataTarget example Job to include run() function for using ExternalIntegration or supplied URL and token.
--- a/nautobot_ssot/jobs/examples.py
+++ b/nautobot_ssot/jobs/examples.py
@@ -2,17 +2,21 @@
 
 # Skip colon check for multiple statements on one line.
 # flake8: noqa: E701
-
+# pylint: disable=too-many-lines
 try:
     from typing_extensions import TypedDict  # Python<3.9
 except ImportError:
     from typing import TypedDict  # Python>=3.9
 
-from typing import Optional, Mapping, List
+from typing import List, Mapping, Optional
+
+import requests
+from diffsync import Adapter
+from diffsync.enum import DiffSyncFlags
+from diffsync.exceptions import ObjectNotFound
 from django.contrib.contenttypes.models import ContentType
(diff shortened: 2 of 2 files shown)


changes/507.fixed
```diff
@@ -0,0 +1 @@
+Fix DataTarget example Job to include run() function for using ExternalIntegration or supplied URL and token.
```
gsnider2195, line 1:
```suggestion
Fixed DataTarget example Job to include run() function for using ExternalIntegration or supplied URL and token.
```

nautobot_ssot/jobs/examples.py
```diff
@@ -943,6 +947,38 @@ def data_mappings(cls):
...
+        memory_profiling,
+        target,
+        target_url,
+        target_token,
+        *args,
+        **kwargs,
+    ):
+        """Run sync."""
+        self.dryrun = dryrun
+        self.memory_profiling = memory_profiling
```
gsnider2195, line 962: As long as we're using celery it's not safe to store any information on the Job instance

jdrew82: This has been required for all SSoT Apps as we need to save the value and pass it to super().

gsnider2195: It doesn't have to be fixed here but since celery only instantiates a task once it's dangerous to store anything on the instance. That is probably something that should be followed up on

nautobot_ssot/jobs/examples.py
```diff
@@ -943,6 +947,38 @@ def data_mappings(cls):
...
+                secrets_group = target.secrets_group
+                self.target_token = secrets_group.get_secret_value(
+                    access_type=SecretsGroupAccessTypeChoices.TYPE_HTTP,
+                    secret_type=SecretsGroupSecretTypeChoices.TYPE_TOKEN,
+                )
+            else:
+                self.target_url = target_url
+                self.target_token = target_token
+        except Exception as error:
+            # TBD: Why are these exceptions swallowed?
```
gsnider2195, line 976: Is there a follow-up issue for this note?

nautobot_ssot/jobs/examples.py
gsnider2195: Also related to the secrets being stored on the instance below, credentials could be leaked between jobs using this pattern.

gsnider2195 (approved):

nautobot_ssot/jobs/examples.py
jdrew82: This appears to be from @snaselj but I'm not seeing one in a quick search. I've opened #521 to track it.

nautobot_ssot/jobs/examples.py
jdrew82: @gsnider2195 I'll follow up with you about this.

PR #526, opened 2024-08-28
Update the ExampleDataSource job to improve memory utilization

gsnider2195:
This changes the `NautobotRemote` adapter to use a generator to process large lists of data instead of building a large list and returning it.

This also updates the logging output to be more human-readable.

## Before

image

## After

image

Also confirmed with `docker stats` that the container memory utilization only increased by ~110MiB after running the job


jdrew82 (approved):
LGTM!

Issue #528, opened 2024-08-29
PROTECT Method Breaks New Log Cleanup Job


cmong-nextech:
### Environment
* Python version: 3.11
* Nautobot version:  2.3.1
* nautobot-ssot version:  2.7.0

### Expected Behavior
`Log Cleanup` job runs without issue.

### Observed Behavior
Job fails and references the `SSOT FK Sync.job_result`

### Steps to Reproduce
1. Utilize the SSOT plugin to sync something to nautobot, creating a job log in the process.
2. Run the `Log Cleanup` job introduced into Nautobot in version 2.3.1
3. Witness the following error:
```
{
    "exc_message": [
        "Cannot delete some instances of model 'JobResult' because they are referenced through protected foreign keys: 'Sync.job_result'.",
        [
            {
                "__nautobot_type__": "nautobot_ssot.models.Sync",
                "display": "Sync Network Data From Network → Nautobot, 2024-07-31 23:43",
                "id": "b9a07bc5-1640-43b7-b701-b6f947102016"
            },
```

Kircheneer:
We could change the foreign key from `Sync` to cascade when the `JobResult` is deleted and put a warning into the docs about this. This would in my mind be the most straight-forward behavior. @jdrew82, @glennmatthews any thoughts?

glennmatthews:
Sounds appropriate to me.

jdrew82:
> We could change the foreign key from `Sync` to cascade when the `JobResult` is deleted and put a warning into the docs about this. This would in my mind be the most straight-forward behavior. @jdrew82, @glennmatthews any thoughts?

I agree that this makes the most sense.

PR #532, opened 2024-09-06
Change PROTECT to CASCADE for Sync

jdrew82:
This PR closed out #528.

I also ran across a few other minor bugs that I fixed here when trying to create the migrations.

- The _MIN_NAUTOBOT_VERSION check wasn't being done right and 2.1 wasn't working due to ACI requiring 2.2.
- DNA Center was missing from list so I've added that.
- I also had to adjust the logger in the Job dunder init because the utils file has a Controller import.


glennmatthews (approved):
lgtm

Issue #521, opened 2024-08-23
Determine Reasoning for Exception Handling in Example Job


jdrew82:
### Environment
* Python version:  3.8
* Nautobot version:  2.3.0
* nautobot-ssot version:  3.0.0

It appears that there's a try/except block capturing all Exceptions in the DataSource Job [here](https://github.com/nautobot/nautobot-app-ssot/blob/8d007faba8dabeda4a23aaa9c210e6485b870149/nautobot_ssot/jobs/examples.py#L871). We need to investigate to see why this is and most likely remove and change to only capture expected Exceptions and allow others to populate.

Kircheneer:
Looks like it should be `nautobot.extras.exceptions.SecretError`

PR #534, opened 2024-09-06
Fix Generalized Exceptions in Example Jobs

jdrew82:
This PR is to close out #521. It replaces the generalized Exceptions with SecretError and raises a custom Exception if SecretsGroup isn't found on ExternalIntegration.

### DIFF
--- a/nautobot_ssot/jobs/examples.py
+++ b/nautobot_ssot/jobs/examples.py
@@ -37,8 +37,6 @@ name = "SSoT Examples"  # pylint: disable=invalid-name
 class MissingSecretsGroupException(Exception):
     """Custom Exception in case SecretsGroup is not found on ExternalIntegration."""
 
-    pass
-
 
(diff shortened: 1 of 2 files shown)


nautobot_ssot/jobs/examples.py
```diff
@@ -866,7 +878,7 @@ def run(  # pylint: disable=too-many-arguments, arguments-differ
             else:
                 self.source_url = source_url
                 self.source_token = source_token
-        except Exception as error:
+        except SecretError as error:
             # TBD: Why are these exceptions swallowed?
```
smk4664, line 882: Should this comment still be here?

smk4664 (approved):
Just one comment, but otherwise this looks to resolve the attached issue.

nautobot_ssot/jobs/examples.py
jdrew82: Oops, good catch! I've removed.

PR #549, opened 2024-09-16
Update variable used in Dryrun job decision

qduk:
This changes the variable looked at for dryrun to be `self.sync.dry_run` instead of `self.dryrun`.

I can pull out the format changes if need be.

### DIFF
--- a/changes/548.fixed
+++ b/changes/548.fixed
@@ -1,1 +1,1 @@
-Fixed SSoT jobs not respecting DryRun variable
+Fixed SSoT jobs not respecting DryRun variable.
--- a/changes/548.fixed
+++ b/changes/548.fixed
@@ -1,0 +1,1 @@
+Fixed SSoT jobs not respecting DryRun variable
--- a/nautobot_ssot/tests/test_jobs.py
+++ b/nautobot_ssot/tests/test_jobs.py
@@ -93,19 +93,37 @@ class BaseJobTestCase(TransactionTestCase):
         self.assertTrue(self.job.sync.dry_run)
         self.assertEqual(self.job.job_result, self.job.sync.job_result)
 
-    @patch.object(DataSyncBaseJob, "execute_sync")
-    def test_job_dryrun_false(self, mock_execute_sync):
-        """Test the job is ran in dryrun mode."""
-        self.job.run(dryrun=False, memory_profiling=False)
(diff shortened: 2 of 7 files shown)


nautobot_ssot/tests/test_jobs.py
```diff
@@ -113,9 +139,6 @@ def test_calculate_diff_fail_diff_save_too_large(self):
         self.job.logger.warning = Mock()
         self.job.source_adapter.diff_to().dict.return_value = {}
         self.job.calculate_diff()
-        self.job.logger.warning.assert_any_call(
```
chadell, line 116: why is this no longer valid?

chadell (approved):
LGTM

changes/548.fixed
```diff
@@ -0,0 +1 @@
+Fixed SSoT jobs not respecting DryRun variable
```
gsnider2195, line 1:
```suggestion
Fixed SSoT jobs not respecting DryRun variable.
```

gsnider2195:
Are you still using black for formatting? It looks like your autoformatter has defaulted to 100 character line-width which is creating a bunch of noise. There's a bug in the `ruff-format` ci check that causes it to always pass but ruff formatting is actually failing if you look at https://github.com/nautobot/nautobot-app-ssot/actions/runs/10890864505/job/30220509988?pr=549

nautobot_ssot/tests/test_jobs.py
qduk: Thanks for catching that Christian!

qduk:
@gsnider2195  I've gone and fixed ruff. I had a unittest failing as well but I've gone and resolved that.

gsnider2195:
> @gsnider2195 I've gone and fixed ruff. I had a unittest failing as well but I've gone and resolved that.

Ruff is still complaining about one file:

> Would reformat: nautobot_ssot/jobs/base.py

qduk:
> > @gsnider2195 I've gone and fixed ruff. I had a unittest failing as well but I've gone and resolved that.
> 
> Ruff is still complaining about one file:
> 
> > Would reformat: nautobot_ssot/jobs/base.py

Now it's "ruffed". Thanks Gary!

jdrew82 (approved):
LGTM!

jdrew82 (approved):
LGTM!

PR #550, opened 2024-09-18
fixed invoke check-migrations in ltm

gsnider2195:
This fixes the `invoke check-migrations` command so that upstream testing doesn't fail on this step anymore.


jdrew82 (approved):
LGTM!

Issue #558, opened 2024-10-02
 Potential KeyError in SSOT Integration with ACI (nautobot_ssot/integrations/aci/diffsync/models/nautobot.py)


TheHack42:
Hello,

I'm encountering a potential issue in the SSOT integration with ACI, specifically in the following code:
https://github.com/nautobot/nautobot-app-ssot/blob/f381da2b5ecdbc75af8b688ca499c8bff46405c0/nautobot_ssot/integrations/aci/diffsync/models/nautobot.py#L452

**Problem:**
The code is trying to access "vrf" as an attribute from the attrs dictionary.
However, "vrf" is not an attribute in attrs; it is actually an identifier of the DiffSyncModel. Therefore, "vrf" should be accessed from the ids dictionary instead, like so: `ids["vrf"]`

If we continue accessing "vrf" via attrs["vrf"], a KeyError will be raised when this code attempts to log the warning.

**Suggested Fix:**
We should update the logger line to properly access "vrf" from ids:
`adapter.job.logger.warning(f"Tenant {attrs['vrf_tenant']} not found for VRF {ids['vrf']}")`.

Please apply the necessary fix.

Thanks

jdrew82:
Good catch! I'll get a PR in to fix this. BTW, in the future we do accept PRs from the community.

PR #559, opened 2024-10-02
Fix VRF Attribute in ACI Prefix.create()

jdrew82:
This PR closes issue mentioned in #558.


gsnider2195 (approved):
Just a minor nitpick and it's something we should fix globally instead of just here. This shouldn't be an fstring since it's a logger but should use `%s`

PR #562, opened 2024-10-02
IP Fabric Bug - Network has host bits set

jjeff07:
Fixes: https://github.com/nautobot/nautobot-app-ssot/issues/561

Set `strict=False` for parsing IP Networks using `ipaddress.ip_network`


jdrew82 (approved):
LGTM!

PR #566, opened 2024-10-09
Allow for customization of NautobotAdapter loading on a per model basis

### LINKED ISSUE #400: Allow for customization of `NautobotAdapter` loading on a per-model basis
### Environment
* Nautobot version:  
* nautobot-ssot version:  

### Use Case

If I need some special filtering logic for a certain model, I want a hook where I can place that.

### Implementation Idea

Pseudo code:

```python
def load(self):
    for model_name in self.top_level:
        if hasattr(self, f"load_{model_name}"):
            getattr(self, f"load_{model_name}")()
        else:
             # Normal code
             pass
```

Renrut5:
Closes #400

### DIFF
--- a/changes/400.added
+++ b/changes/400.added
@@ -1,0 +1,1 @@
+Add ability to customize `NautobotAdapter` loading on a per-model basis.
--- a/nautobot_ssot/contrib/adapter.py
+++ b/nautobot_ssot/contrib/adapter.py
@@ -168,6 +168,10 @@ class NautobotAdapter(Adapter):
             raise ValueError("'top_level' needs to be set on the class.")
 
         for model_name in self.top_level:
+            if hasattr(self, f"load_{model_name}"):
+                getattr(self, f"load_{model_name}")()
+                continue
+
             diffsync_model = self._get_diffsync_class(model_name)
 
             # This function directly mutates the diffsync store, i.e. it will create and load the objects
--- a/docs/dev/jobs.md
+++ b/docs/dev/jobs.md
@@ -135,7 +135,34 @@ class YourSSoTNautobotAdapter(NautobotAdapter):
 
 The `load` function is already implemented on this adapter and will automatically and recursively traverse any children relationships for you, provided the models are [defined correctly](../user/modeling.md).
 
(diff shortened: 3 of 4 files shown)


changes/400.added
```diff
@@ -0,0 +1 @@
+Add ability to customize `NautobotAdapter` loading on a per-model basis.
```
chadell, line 1: isn't this file 566?

nautobot_ssot/contrib/adapter.py
```diff
@@ -168,6 +168,10 @@ def load(self):
             raise ValueError("'top_level' needs to be set on the class.")
 
         for model_name in self.top_level:
+            if hasattr(self, f"load_{model_name}"):
```
chadell, line 171: we should document this behavior for developers... but why not just removing the model from the top_level if you don't want to load it?

changes/400.added
Renrut5: Fragment files use issue number:
https://docs.nautobot.com/projects/core/en/stable/development/core/?h=changelog#creating-changelog-fragments

nautobot_ssot/contrib/adapter.py
Renrut5: @chadell The point is we still want to load the model, but be able to customize the functionality when loading on a per-model basis. 

I'll get started on the docs.

Renrut5: @chadell looking into it, it doesn't appear there's any docs yet for using contrib. Because of that, I think it would be best to save the docs portion for a separate issue/PR.

jdrew82: @Renrut5 The docs for contrib are under Modeling in the User Guide section.

jdrew82:
I'm a bit confused about the purpose behind this PR. You already control what's loaded via the top_level and parent/child relationships. Can you elaborate on what the use case is for this?

Renrut5:
@jdrew82 I'll yield that question to @Kircheneer since he's the one who made the issue to begin with.

Kircheneer:
As per comment in #400 I will close this as this functionality already exists.

PR #586, opened 2024-10-24
Fix Bootstrap for DLM 3.0 and Correct Docs

### LINKED ISSUE #585: Upgrading to DLM 3.0 Throws LookupError
### Environment
* Python version:  3.10.
* Nautobot version:  2.3.8
* nautobot-ssot version:  3.2.0

### Expected Behavior
Expected for Bootstrap integration to load without issue when upgrading to DLM 3.0.

### Observed Behavior
Observed that a LookupError is thrown due to the SoftwareLCM class being removed:

```
  File "/usr/local/lib/python3.10/site-packages/nautobot_ssot/integrations/bootstrap/signals.py", line 53, in nautobot_database_ready_callback
    SoftwareLCM = apps.get_model("nautobot_device_lifecycle_mgmt", "SoftwareLCM")
  File "/usr/local/lib/python3.10/site-packages/django/apps/registry.py", line 213, in get_model
    return app_config.get_model(model_name, require_ready=require_ready)
  File "/usr/local/lib/python3.10/site-packages/django/apps/config.py", line 237, in get_model
    raise LookupError(
LookupError: App 'nautobot_device_lifecycle_mgmt' doesn't have a 'SoftwareLCM' model.
```

### Steps to Reproduce
1. Update project instance to use next3.0 branch for DLM.
2. Enable Bootstrap integration.
3. Attempt startup of Nautobot and see LookupError.

jdrew82:
This PR closes #585 and also updates the documentation on the Bootstrap integration to clarify installation and settings.


nautobot_ssot/integrations/bootstrap/signals.py
```diff
@@ -83,13 +78,21 @@ def nautobot_database_ready_callback(sender, *, apps, **kwargs):  # pylint: disa
...
-                "software": SoftwareLCM,
-                "software_image": SoftwareImageLCM,
-                "validated_software": ValidatedSoftwareLCM,
-            }
-        )
+        try:
+            SoftwareLCM = apps.get_model("nautobot_device_lifecycle_mgmt", "SoftwareLCM")
+            signal_to_model_mapping["software"] = SoftwareLCM
+        except LookupError as err:
+            print(f"Unable to find SoftwareLCM model from Device Lifecycle Management App. {err}")
```
itdependsnetworks, line 85: In my mind we should either.

- Rely on pyproject.toml to ensure we side step this issue by pinning / setting the proper versions of dependencies
- Support both new and old versions

The print to me would not be easily seen, and as far as I can tell, not let the customer know the issue is exactly.

jdrew82: @itdependsnetworks The problem is this isn't a direct dependency. The Bootstrap integration supports both Golden Config and Device Lifecycle Management. The problem is that with DLM's change to get rid of SoftwareLCM and SoftwareImageLCM we need to account for that with this integration as it was blowing up my testing of the DLM update.

docs/admin/integrations/bootstrap_setup.md
```diff
@@ -1,29 +1,28 @@
...
...tforward. Assuming that you're copying the example settings below, the `bootstrap_nautobot_environment_branch` setting will be loaded from the environment variable `NAUTOBOT_BOOTSTRAP_SSOT_ENVIRONMENT_BRANCH`, or default to develop. The `bootstrap_models_to_sync` setting defines which models/objects you want to have the App sync to Nautobot. There are a couple of caveats to this functionality. For example, for DynamicGroup objects to sync, the filter criteria need to already exist in Nautobot. So, if you are going to have groups that are filtered on Platforms, Locations, etc, make sure not to include DynamicGroup objects in your Bootstrap Data until those items exist in Nautobot. The same goes for Golden Config-related Git Repositories. It should go without saying, but the [Golden Config App](https://github.com/nautobot/nautobot-app-golden-config) must be installed, for the backup, intended, and template `provided_contents` items to be available options in your Bootstrap Data. This also goes for support of the `SoftwareLCM, SoftwareImageLCM, and ValidatedSoftware` models from the [Device Lifecycle Management app](https://github.com/nautobot/nautobot-app-device-lifecycle-mgmt) with.
```
bile0026, line 17: So, if you are going to have groups that are filtered on Platforms, Locations, etc, make sure not to include DynamicGroup objects in your Bootstrap Data until those items exist in Nautobot.

Might make sense to add:
So, if you are going to have groups that are filtered on Platforms, Locations, etc, make sure not to include DynamicGroup objects in your Bootstrap Data until those items exist in Nautobot. If these items are also being synced from Bootstrap in your source YAML and they will be created in the proper order.

bile0026 (commented):
One small edit to docs but otherwise looks good to me.

docs/admin/integrations/bootstrap_setup.md
jdrew82: @bile0026 Slightly tweaked it to use Bootstrap Data to be clear of the source of the configuration.

bile0026 (approved):
LGTM

PR #589, opened 2024-10-25
Re-add Hostname Mapping Feature to DNA Center

### LINKED ISSUE #588: DNA Center Missing Hostname Mapping Feature
### Environment
* Python version:  3.10
* Nautobot version:  2.3.8
* nautobot-ssot version:  3.2.0

### Expected Behavior
Expected to be able to use hostnames to assign Role to Devices.

### Observed Behavior
Found that the hostname_mapping feature from original app was broken and not updated to be in the Job form.

### Steps to Reproduce
1. Install Nautobot and SSoT.
2. Enable DNA Center integration.
3. Enable Job and look at form.

jdrew82:
This PR closes #588 by adding back in the hostname mapping functionality that the original had but has been updated to have the setting be used through the Job form. I've also gone ahead and added support for the SoftwareVersion model and added handling for cases where SoftwareLCM isn't available.


nautobot_ssot/integrations/dna_center/diffsync/models/nautobot.py
```diff
@@ -23,12 +23,19 @@
...
+    from nautobot_device_lifecycle_mgmt import SoftwareLCM  # noqa: F401
 
     LIFECYCLE_MGMT = True
 except ImportError:
     LIFECYCLE_MGMT = False
 
+try:
+    from nautobot.dcim.models import SoftwareImageFile, SoftwareVersion  # noqa: F401
+
+    SOFTWARE_VERSION_FOUND = True
```
Kircheneer, line 35: Should this variable be called something like `SOFTWARE_VERSION_FOUND_IN_CORE` or something? Because given `LIFECYCLE_MGMT` it would be there up until some version so it may be good to differentiate these.

nautobot_ssot/integrations/dna_center/diffsync/models/nautobot.py
```diff
@@ -214,10 +221,20 @@ def create(cls, adapter, ids, attrs):
         if attrs.get("tenant"):
             new_device.tenant_id = adapter.tenant_map[attrs["tenant"]]
         if attrs.get("version"):
-            new_device.custom_field_data.update({"os_version": attrs["version"]})
```
Kircheneer, line 217: Do I understand correctly that this effectively removes compatibility with having the software version come from the plugin rather than Nautobot? Not a problem, just something I'd like to call out a bit more explicitly in 588.removed

nautobot_ssot/integrations/dna_center/diffsync/models/nautobot.py
jdrew82: Sure, updated.

nautobot_ssot/integrations/dna_center/diffsync/models/nautobot.py
jdrew82: This was to remove the `OS Version` CustomField as it isn't needed any longer. I thought I was clear in the changelog snippet about it being removed but happy to repharse if you can think of a clearer way of stating it.

Kircheneer (approved):

PR #594, opened 2024-11-05
Various Bugfixes

jdrew82:
- Fixes bug where multiple locations assigned to a Prefix would cause Meraki integration to fail.
- Added ability to Meraki and DNA Center integrations to allow renaming of imported locations using `name` key in location map option.
- Fixed DNA Center loading incorrect names for buildings and floors in Nautobot adapter.
- Added missing code owners for DNA Center, Meraki, and Itential integrations.
- Fixed KeyError being thrown when port is missing from uplink_settings dict in Meraki integration.
- Fixed bug in Bootstrap integration that occurs when loading ValidatedSoftware in a system where SoftwareLCM doesn't exist.


bile0026 (approved):
LGTM

PR #599, opened 2024-11-14
Add Citrix ADM Integration

jdrew82:
This PR Is to add an integration for Citrix ADM.


pyproject.toml
```diff
@@ -49,6 +49,7 @@ python-magic = { version = ">=0.4.15", optional = true }
 pytz = { version = ">=2019.3", optional = true }
 requests = { version = ">=2.21.0", optional = true }
 requests-oauthlib = { version = ">=1.3.0", optional = true }
+urllib3 = { version = "^2.2.3", optional = true }
```
Kircheneer, line 52: Can we go to >= 2.2.3 or is there a known incompatibility?

Kircheneer: Also, requests already requires urllib3: https://github.com/psf/requests/blob/main/setup.py#L39

Do we use anything directly from urllib3?

nautobot_ssot/integrations/dna_center/utils/dna_center.py
```diff
@@ -198,25 +198,6 @@ def get_port_status(port_info: dict):
             status = "Planned"
         return status
 
-    @staticmethod
-    def parse_hostname_for_role(hostname_map: List[Tuple[str, str]], device_hostname: str):
```
Kircheneer, line 202: Can we get a quick change log entry for this change?

pyproject.toml
jdrew82: @Kircheneer It's used for disabling SSL verification warnings. As for the version used, I just added with `poetry add urllib3` so it was poetry that decided on that format.

nautobot_ssot/integrations/dna_center/utils/dna_center.py
jdrew82: Done!

pyproject.toml
Kircheneer: I know that this is the poetry default, but I don't like it, unless there is a reason to believe we can't use urllib 3.

jdrew82: Updated to >=2.2.3.

Kircheneer (approved):

Issue #600, opened 2024-11-15
Add Slurpit Support


lpconsulting321:
### Environment
* Nautobot version:  v2+
* nautobot-ssot version: v3.2.0+

### Proposed Functionality
Add the ability to sync data from network discovery using slurpit.io. Ability to sync device, sites, IP's and more. 

### Use Case
Slurpit users would like an easy way to be able to sync data discovered from Slurpit into Nautobot to further develop their source of truth.

chadell:
Isn't https://slurpit.io/nautobot-plugin/ working well for you?

jdrew82:
@lpconsulting321 Isn't this already covered by the App that Slurpit made? https://slurpit.io/nautobot-plugin/

pietos:
We worked on this ssot app implementation since we got a couple requests for it. It's more flexible for other data models then the once that are in the plugin for the moment.

lpconsulting321:
Closing ticket as this has been merged.

PR #601, opened 2024-11-15
Slurpit Support

lpconsulting321:
Implements: #600 

Added support for slurpit to sync locations, device types, devices, interfaces, inventory items, prefixes, vlans, vrf's and IP's. 

Let me know if there are any issues.


jdrew82 (approved):
LGTM! Good work and awesome to see an async example with the contrib pattern!

jdrew82:
@lpconsulting321 One thing I just realized I forgot about. Can you update the CODEOWNERS in the github folder with who is responsible for the integration? Would it be just you or are there others?

lpconsulting321:
@jdrew82 Yeah. I wanted to use the built in Nautobot adapter as much as possible. Possible improvements for that in future would be to allow the ability allow `get_queryset` to pull based on tags/cf's I overcome this by just overiding the functions. Other than that it was perfect.

lpconsulting321:
@jdrew82 

`/nautobot_ssot/integrations/slurpit/ @lpconsulting321 @pietos @nautobot/plugin-ssot` is this okay or shall i remove the last user?

jdrew82:
> @jdrew82
> 
> `/nautobot_ssot/integrations/slurpit/ @lpconsulting321 @pietos @nautobot/plugin-ssot` is this okay or shall i remove the last user?

You can leave the plugin-ssot group as a fallback just in case it's something we can easily address.

lpconsulting321:
Do you know what the issue might be on the pylint?

jdrew82 (approved):
LGTM!

lpconsulting321:
@jdrew82 Looks like unit tests are failing for non slurpit integrations. What do you do in these situations?

development/development.env
```diff
@@ -110,3 +110,6 @@ IPFABRIC_SSL_VERIFY="True"
 IPFABRIC_TIMEOUT=15
 
 NAUTOBOT_SSOT_ENABLE_ITENTIAL="True"
+
+NAUTOBOT_SSOT_ENABLE_SLURPIT="True"
```
jdrew82, line 114: This should be False.

jdrew82: @lpconsulting321 That is most likely being caused due to your integration being enabled. Making this change should resolve the issue.

lpconsulting321: <3

lpconsulting321:
Ahh 3.8. The Slurpit SDK does not support 3.8 at the moment. We will make sure our SDK supports 3.8 and get this passing.

jdrew82:
@lpconsulting321 Nautobot 2.4 will be removing support for Python 3.8 so you might want to wait for that to be released and we can set that as minimum for this integration?

jdrew82:
Oh, also, before I forget, please make sure you update the docs in the README to include you integration in the list and the acknowledgements.

Issue #634, opened 2024-12-13
Service Now Integration - SourceAdapter load Location with site filter


michalis1:
### Environment
* Python version:  3.11.4
* Nautobot version:  2.3.12
* nautobot-ssot version:  3.3.0

### Expected Behavior
When you run the Nautobot ⟹ ServiceNow job with the site filter set, you expect the job to sync only the specified locations to ServiceNow.

### Observed Behavior
It seems that the job loads only the parent locations, but not the specified location itself - [Load Location with site filter](https://github.com/nautobot/nautobot-app-ssot/blob/d27068b74935ba2f3c8ee69817f3e8ad079def50/nautobot_ssot/integrations/servicenow/diffsync/adapter_nautobot.py#L60).

### Steps to Reproduce
1. Go to the Nautobot ⟹ ServiceNow job 
2. Select a location in the Site Filter
3. Run the Job

PR #637, opened 2024-12-17
Service Now Integration - Fixes issue #634

michalis1:
### Fixes issue 634

When running the Nautobot ⟹ ServiceNow job with a site filter set, you expect the job to sync only the specified locations to ServiceNow. 
However, it appears that the job currently loads only the parent locations, not the specified location itself.
This is fixed by initializing the locations list with the user-selected location and then loading the parent locations of that selection.

### DIFF
--- a/changes/634.fixed
+++ b/changes/634.fixed
@@ -1,4 +1,1 @@
-ServiceNow Integration
-When running the Nautobot ⟹ ServiceNow job with a site filter set, you expect the job to sync only the specified locations to ServiceNow. 
-However, it appears that the job currently loads only the parent locations, not the specified location itself.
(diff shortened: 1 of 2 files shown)


changes/634.fixed
```diff
@@ -0,0 +1,4 @@
+ServiceNow Integration
```
jdrew82, line 1: Please rephrase this to be a bit more concise. This is the changelog fragment that will be added to the release notes when we cut the next release. This should be a single line or two at most.

jdrew82 (changes requested):

changes/634.fixed
michalis1: I’ve made the change. I hope it’s better now.

jdrew82 (approved):
LGTM!

Issue #641, opened 2025-01-06
Import issue with IPAM models when DLM app not installed


bile0026:
### Environment
* Python version:  3.11.4
* Nautobot version:  2.3.15
* nautobot-ssot version:  3.3.0

### Expected Behavior
Plugin should work without the nautobot_device_lifecycle app installed.

### Observed Behavior
Imports for the following models are failing if DLM plugin is not installed becuase they are within the `if` block checking for whether the app is installed and they should not be. 
BootstrapNamespace
BootstrapRiR
BootstrapVLANGroup
BootstrapVLAN
BootstrapVRF
BootstrapPrefix

### Steps to Reproduce
1. Run Bootstrap to Nautobot job without the DLM app installed in the Nautobot environment

PR #642, opened 2025-01-06
Bootstrap: fix imports for ipam models

bile0026:
Fixes: #641 

Fix incorrectly nested classes for IPAM models in the bootstrap models.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -83,6 +83,7 @@ jobs:
       matrix:
         python-version: ["3.11"]
         nautobot-version: ["2.2.1"]
+        poetry-version: ["1.8.4"]
     env:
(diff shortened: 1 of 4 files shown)


glennmatthews:
I think the right way to pin the poetry version is something like:
```yaml
      - name: "Setup environment"
        uses: "networktocode/gh-action-setup-poetry-environment@v6"
        with:
          poetry-version: "1.8.4"
```

bile0026:
> I think the right way to pin the poetry version is something like:
> 
> ```yaml
>       - name: "Setup environment"
>         uses: "networktocode/gh-action-setup-poetry-environment@v6"
>         with:
>           poetry-version: "1.8.4"
> ```

This would be a separate "job"?

PR #649, opened 2025-01-09
Fix Citrix ADM Removing Used SoftwareVersions

### LINKED ISSUE #648: Citrix ADM Removes SoftwareVersions Used with ValidatedSoftware
### Environment
* Python version:  3.10
* Nautobot version:  2.3.12
* nautobot-ssot version:  3.3.0

### Expected Behavior
I expect that a Citrix SoftwareVersion that's in use with ValidatedSoftware to be left and not deleted accidentally.

### Observed Behavior
The SoftwareVersion that is defined with a ValidatedSoftware but not assigned to a Device is being deleted.

### Steps to Reproduce
1. Create SoftwareVersion for citrix.adc.
2. Create ValidatedSoftware with newly created SoftwareVersion. Do NOT assign SoftwareVersion to a Device.
3. Run Citrix ADM SSoT integration.

jdrew82:
This PR closes #648 and fixes the issue with deleting a SoftwareVersion that is used with a ValidatedSoftware. I found that this same issue was also possible with Meraki integration so replicated the check there.

### DIFF
--- a/changes/648.fixed
+++ b/changes/648.fixed
@@ -1,1 +1,2 @@
 Fixed Citrix ADM deleting SoftwareVersion in use with ValidatedSoftware.
+Fixed Meraki deleting SoftwareVersion in use with ValidatedSoftware.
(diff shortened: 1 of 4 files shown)


bile0026 (approved):
LGTM

Issue #669, opened 2025-01-17
ACI loading requires tag on ACI agnostic objects resulting in duplication exceptions


kingfetty:
### Environment
* Python version:  3.11
* Nautobot version:  2.3.15
* nautobot-ssot version:  3.4.0

### Expected Behavior
ACI SSoT will be able to sync with an ACI fabric even if objects such as DeviceType, InterfaceTemplate, and Prefix already exist within Nautobot. 

### Observed Behavior
If DeviceType or InterfaceTemplate already exist within the Nautobot system prior to ACI sync and do not have the proper tags applied the sync will fail because the sync tries to create an object that already exists with the same primary key. 

### Steps to Reproduce
1. Configure a Nautobot instance to sync to an ACI fabric
2. Create a DeviceType for some of the switch models that will exist inside the ACI fabric
3. Attempt to sync with the ACI fabric
4. Note the sync fails because it tries to create a DeviceType that already exists.  This is because the query against DeviceType excludes objects that do not have the proper tag applied.  Device Types will likely exist inside and outside the ACI fabric so the tag filter should not be applied to the query.

kingfetty:
This issue will persist due to the fact that DeviceTypes non-existent in ACI would end up being removed from Nautobot.

PR #671, opened 2025-01-17
Exclude tag requirements when loading DeviceType and InterfaceTemplate

kingfetty:
# Closes: #669 

## What's Changed
Remove tag filter when loading DeviceType and InterfaceTemplate into diffsync model.

## To Do

- [X] Added change log fragment(s) (for more information see [the documentation]

### DIFF
--- a/changes/669.fixed
+++ b/changes/669.fixed
@@ -1,0 +1,1 @@
+Excluded tag requirements when loading DeviceType, InterfaceTemplate
--- a/nautobot_ssot/integrations/aci/diffsync/adapters/nautobot.py
+++ b/nautobot_ssot/integrations/aci/diffsync/adapters/nautobot.py
@@ -123,8 +123,7 @@ class NautobotAdapter(Adapter):
 
     def load_devicetypes(self):
         """Method to load Device Types from Nautobot."""
-        _tag = Tag.objects.get(name=PLUGIN_CFG.get("tag"))
-        for nbdevicetype in DeviceType.objects.filter(tags=_tag):
+        for nbdevicetype in DeviceType.objects.all():
             _devicetype = self.device_type(
                 model=nbdevicetype.model,
                 part_nbr=nbdevicetype.part_number,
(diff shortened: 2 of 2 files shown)


jdrew82:
I'm not entirely sure this is a bug or an enhancement. This is changing the default behavior which makes me inclined to say that this should most likely be an optional setting for users. Can you refactor this to work from a setting in the PLUGINS_CONFIG for the App? The default behavior should be to use the Tag to filter the loaded DeviceTypes and InterfaceTemplates and only load all if the setting is True.

kingfetty:
> I'm not entirely sure this is a bug or an enhancement. This is changing the default behavior which makes me inclined to say that this should most likely be an optional setting for users. Can you refactor this to work from a setting in the PLUGINS_CONFIG for the App? The default behavior should be to use the Tag to filter the loaded DeviceTypes and InterfaceTemplates and only load all if the setting is True.

I will work on refactoring as requested.  I still believe it to be incorrect to filter DeviceType at all because it's something that will never be specific to ACI.  The purpose behind the filtering for tags as it was explained to me is for integrations that are not ACI to not step on each other.  It is highly unlikely that you will have DeviceTypes that are just created for ACI and no other purpose.

pato23arg:
> > I'm not entirely sure this is a bug or an enhancement. This is changing the default behavior which makes me inclined to say that this should most likely be an optional setting for users. Can you refactor this to work from a setting in the PLUGINS_CONFIG for the App? The default behavior should be to use the Tag to filter the loaded DeviceTypes and InterfaceTemplates and only load all if the setting is True.
> 
> I will work on refactoring as requested. I still believe it to be incorrect to filter DeviceType at all because it's something that will never be specific to ACI. The purpose behind the filtering for tags as it was explained to me is for integrations that are not ACI to not step on each other. It is highly unlikely that you will have DeviceTypes that are just created for ACI and no other purpose.

@kingfetty without a filter/data ownership tag, the ACI SSOT will be "data owner" of all DeviceTypes. That means that ACI SSOT will have to declare every DeviceType in Nautobot. For example, if you declare a DeviceType manually, ACI SSOT will delete it during the next sync.
While for some users such behavior might be desired, most users want to still be able to define DeviceTypes manually or have others SSOTs able to define them. I agree with @jdrew82 that this should be an optional setting, but default behavior should remain the same.

kingfetty:
> @kingfetty without a filter/data ownership tag, the ACI SSOT will be "data owner" of all DeviceTypes. That means that ACI SSOT will have to declare every DeviceType in Nautobot. For example, if you declare a DeviceType manually, ACI SSOT will delete it during the next sync. While for some users such behavior might be desired, most users want to still be able to define DeviceTypes manually or have others SSOTs able to define them. I agree with @jdrew82 that this could be an optional setting, but default behavior should remain the same.

Thank you for the clarification here, and a very good observation.  I will withdraw this and kill the issue.  I'll have to resort to manually tagging the DeviceTypes in my Nautobot Instance to get around this issue then.

PR #703, opened 2025-02-20
adds catching of ValueError on validated_save in contrib

Kircheneer:
Had this in an environment where the MTU for an interface was empty, it didn't raise a validation error but rather a value error.


chadell (approved):
LGTM

Issue #726, opened 2025-03-13
Infoblox Integration - Namespace/View Mapping


michalis1:
### Environment
* Nautobot version:  2.4
* nautobot-ssot version:  3.5.0

### Proposed Functionality

Currently, the integration supports a [one-to-one mapping](https://github.com/nautobot/nautobot-app-ssot/blob/6bf86de74e7d6248cdaf9571aa62c960ef679d19/nautobot_ssot/integrations/infoblox/utils/diffsync.py#L153) between Infoblox network views and Nautobot namespaces. Specifically, the Infoblox network view “default” corresponds to the Nautobot namespace “Global”. This mapping is currently hardcoded. Allowing users the flexibility to edit and adjust this mapping based on their requirements would be beneficial. Additionally, there may be a need to add more mappings depending on the environment.

### Use Case

In environments where a Nautobot instance synchronizes data from multiple IPAM sources, it’s possible to encounter overlapping or duplicate prefixes. To manage this, utilizing different namespaces for each data source could be helpful. This approach ensures that prefixes from each source remain distinct and organized within their respective namespaces.
So, if your use case requires that prefixes from Infoblox’s ‘default’ network view be synchronized to a namespace other than ‘Global’ in Nautobot, this default behavior may not suffice. This enhancement would provide greater flexibility in managing data synchronization across multiple IPAM sources.

PR #728, opened 2025-03-14
SSoT Infoblox integration - New setting for network view to namespace mapping

michalis1:
# Closes: #726

## What's Changed

Create a new setting in the SSoT Infoblox integration that allows users to define mappings between Infoblox network views and Nautobot namespaces.

I added a new attribute, `infoblox_network_view_to_namespace_map`, to the `SSOTInfobloxConfig` class. Since this introduces a new column in an existing database table, I avoided setting default values. Instead, I assign the default values in signals.py. I’m not sure if this follows best practices—please let me know if there’s a better way to handle it.


michalis1:
I removed the initialization of the setting from `signals.py` and set a default value in the **Nautobot model**. Based on my conversation with Justin and Glen on Slack, and considering [this](https://github.com/nautobot/cookiecutter-nautobot-app/issues/185), I disabled `new-db-field-with-default`.

Issue #729, opened 2025-03-17
Location doesn’t get synced from Infoblox extensibility attributes.


michalis1:
### Environment
* Python version:  3.11
* Nautobot version:  2.3.16
* nautobot-ssot version:  3.5.0

### Expected Behavior
The location of a prefix should be synced from the “site” or “location” extensibility attribute of an Infoblox network.

### Observed Behavior
The location of the prefix doesn’t get updated, even though the sync logs show that the extensibility attributes have been set correctly on the Infoblox network.

PR #730, opened 2025-03-17
Fix the synchronization of the prefix location from Infoblox extensibility attributes. (#729)

michalis1:
# Closes: #729

## What's Changed

The location of a prefix should be synced from the “site” or “location” extensibility attribute of an Infoblox network.

After reviewing the code, I identified that we were setting [location_id ](https://github.com/michalis1/nautobot-app-ssot/blob/6bf86de74e7d6248cdaf9571aa62c960ef679d19/nautobot_ssot/integrations/infoblox/diffsync/models/nautobot.py#L50) instead of the location of a prefix.

### DIFF
--- a/changes/729.fixed
+++ b/changes/729.fixed
@@ -1,0 +1,1 @@
+Fixed the synchronization of the prefix location from Infoblox extensibility attributes.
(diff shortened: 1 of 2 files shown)


jdrew82 (approved):
LGTM!

PR #731, opened 2025-03-17
Update Aruba/Cisco Platform logic

nopg:
Allows better differentiation between Aruba OS types, as well as add support for (EOL) 8500-WLC AireOS platform type.

Used fantastic chart created here to help with model types: https://github.com/networktocode/netutils/pull/588#issuecomment-2672972706


jdrew82 (approved):
LGTM!

PR #744, opened 2025-03-30
SolarWinds integration: Add 'wireless' to device_type checks

nopg:
Cisco 8540 WLC's were getting `WS-` prepended to their device_type, causing them to be incorrectly categorized as ios, not aireos. This resolves that issue.

Adding the one extra 'if' check caused Pylint to complain with `R0916: Too many boolean expressions in if statement`, so I refactored that logic a tad as well.

Added one extra test case for the `standardize_device_type` function for this case.

### DIFF
--- a/changes/744.added
+++ b/changes/744.added
@@ -1,0 +1,1 @@
+Added extra check when determiming device_type for 'wireless' in Cisco model names.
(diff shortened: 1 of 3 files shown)


jdrew82 (approved):
LGTM!

PR #745, opened 2025-03-30
Bootstrap integration: add enabled option for scheduled_jobs

nopg:
Added `enabled` parameter for scheduled_jobs, in Boostrap integration.

I noticed that disabled Scheduled Jobs do not appear in the list view in the UI, due to [this](https://github.com/nautobot/nautobot/blob/cb9bacbed04e7b8849cd24bd28094b632e093ca2/nautobot/extras/views.py#L1956) in Nautobot core. I'd love to see if that can be changed, but in the meantime I also added a warning to the docs for anyone using this parameter.

### DIFF
--- a/changes/745.added
+++ b/changes/745.added
@@ -1,0 +1,1 @@
+Added enabled option for ScheduledJob in Bootstrap SSOT.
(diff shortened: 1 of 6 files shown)


jdrew82 (approved):
LGTM!

Issue #750, opened 2025-04-01
ServiceNow Instance does not load from DB config due to default settings value


justinbrink:
### Environment
* Python version:  3.11
* Nautobot version:  2.3.1
* nautobot-ssot version:  3.5.1a0 (current develop version, but any are impacted)

### Expected Behavior
Install SSoT App
Enable ServiceNow integration
Configure ServiceNow `Instance` in DB config
Have ServiceNow instance usable for syncing

### Observed Behavior
No serviceNow instance is available unless it is configured in the nautobot_config.py or environment variables. This is due to the default config containing a servicenow_instance key with a `""` value that is always loaded over the config value.

This appears to have been missed in testing because the settings were overridden and the defaults were removed, which makes the existing code work.
```
    @override_settings(PLUGINS_CONFIG={})
    @mock.patch.dict(os.environ, {"SNOW_USERNAME": "someuser", "SNOW_PASSWORD": "notsosecret"})
    def test_config_information_db(self):
        ...
```

### Steps to Reproduce
1. Clone latest develop branch and run local dev environment
2. Add a ServiceNow instance in the UI config form
3. Open the ORM and run...
```
from django.conf import settings
from nautobot_ssot.integrations.servicenow.models import SSOTServiceNowConfig

db_config = SSOTServiceNowConfig.load()
settings_config = settings.PLUGINS_CONFIG.get("nautobot_ssot", {})
settings_config.get("servicenow_instance", db_config.servicenow_instance)
# Instance will be returned blank due to defaults
settings_config.get("servicenow_instance", "") or db_config.servicenow_instance
# Instance will return the settings instance first if provided, or DB config instance if not provided or blank
```

PR #751, opened 2025-04-01
Updated ServiceNow instance config loading

justinbrink:
# Closes: #750

## What's Changed

Updated the ServiceNow integration config loading to allow instance from the DB config. Previously it was being overridden by the default app settings in __init__.py

Tests were also updated to make sure this is captured the same way in the future.

### DIFF
--- a/changes/750.fixed
+++ b/changes/750.fixed
@@ -1,0 +1,1 @@
+Fixed ServiceNow instance configuration loading.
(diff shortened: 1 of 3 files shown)


justinbrink:
CI was passing on this with the exception of the cryptography/cloudvision issue.

jdrew82 (approved):
LGTM!

PR #769, opened 2025-04-07
Bumping Device Lifecycle Mgmt to 3.0.0

mattmiller87:


jdrew82:
I'm going to close this out as I'm working on a fix that addresses the various issues that come from DLM 3.0 still having SoftwareLCM but SoftwareVersion should be used.

Issue #632, opened 2024-12-13
Leverage the `Metadata` feature to allow advanced data governance


chadell:
### Environment
* Nautobot version:  2.3.x
* nautobot-ssot version:  

### Proposed Functionality

Incorporate basic metadata for data imported using the [metadata feature](https://docs.nautobot.com/projects/core/en/stable/user-guide/platform-functionality/objectmetadata/?h=metadata#developer-usage)

### Use Case

This could allow better data governance in Nautobot, and allow for advanced data clean-up functionalities understanding which data souce takes more precedence

From the example:

```py
... omitted code ...

# Define the different metadata sources, in this case ServiceNow
choice_service_now = MetadataChoice(
    metadata_type=source_cmdb_type,
    value="SNOW"
)

... omitted code ...

# For every object/fields, assign the  source
source_cmdb_metadata = ObjectMetadata(
    metadata_type=source_cmdb_type,
    assigned_object=device,
    scoped_fields=["name", "location", "primary_ip4", "device_type"]  # Only these fields are sourced from the upstream CMDB
)
source_cmdb_metadata.value = choice_service_now.value 
```

Kircheneer:
Related: https://github.com/nautobot/nautobot-app-ssot/issues/577

Kircheneer:
The [documentation](https://docs.nautobot.com/projects/core/en/stable/user-guide/platform-functionality/objectmetadata/) already hints at a clear use case, which is the one that you have already outlined @chadell.

The following are things are can provide out of the box through inference:

- Which system of record is this data originally sourced from? (through the name of the integration, including fields through their names)
- When was this data last synced from that system into Nautobot?

I believe we should make these toggle-able, potentially think about even making the names of the `MetadataType` configurable.

chadell:
yup, that's the point. I think it should be part of the framework to easily use it :)

Kircheneer:
Not sure I want to close this quite yet, we haven't incorporated anything in the documentation, and only one of our integrations implement this feature, also there is nothing in contrib.

PR #776, opened 2025-04-10
Enhance contrib to support object metadata.

michalis1:
# Closes: #632

## What's Changed

- Create a MetadataType of type datetime per integration called "Last sync from job.Meta.data_source" - we can do this in the job in the sync_data method by checking if the Nautobot adapter is a subclass of or exactly the nautobot_ssot.contrib.NautobotAdapter class
  - Also check the setting explained further down

- On model.create and model.update, create or update the timestamp value in the metadata
  - Also check the settings explained further down

- In nautobot_config.py under PLUGINS_CONFIG.nautobot_ssot.enable_metadata_for there is a list of Job class names for which metadata should be used

## To Do

- [x] Unit, Integration Tests

### DIFF
--- a/nautobot_ssot/contrib/adapter.py
+++ b/nautobot_ssot/contrib/adapter.py
@@ -12,7 +12,7 @@ from diffsync.exceptions import ObjectCrudException
 from django.contrib.contenttypes.models import ContentType
 from django.db.models import Model
 from nautobot.extras.choices import RelationshipTypeChoices
-from nautobot.extras.models import Relationship, RelationshipAssociation
+from nautobot.extras.models import MetadataType, Relationship, RelationshipAssociation
 from typing_extensions import get_type_hints
 
 from nautobot_ssot.contrib.types import (
@@ -50,6 +50,7 @@ class NautobotAdapter(Adapter):
         super().__init__(*args, **kwargs)
         self.job = job
         self.sync = sync
+        self.metadata_type = None
         self.invalidate_cache()
 
     def invalidate_cache(self, zero_out_hits=True):
@@ -388,3 +389,26 @@ class NautobotAdapter(Adapter):
             if lookups[-1] in ["app_label", "model"]:
                 return getattr(ContentType.objects.get_for_model(related_object), lookups[-1])
         return None
+
+    def get_or_create_metadatatype(self):
+        """Retrieve or create a MetadataType object to track the last sync time of this SSoT job."""
+        # MetadataType name will be extracted from the Data Source name.
+        print(dir(self.job.__class__))
+        metadata_type__name = self.job.__class__.Meta.data_source
+
+        # Create a MetadataType of type datetime
+        metadata_type, created = MetadataType.objects.get_or_create(
+            name=f"Last sync from {metadata_type__name}",
+            defaults={
+                "data_type": "datetime",
+                "description": f"Timestamp of the last sync from the Data source {metadata_type__name}",
+            },
+        )
+
+        # Use select_related for content_types since they'll be used during model CRUD operations.
+        if not created:
+            # Refetch with select_related if needed
+            metadata_type = MetadataType.objects.prefetch_related("content_types").get(pk=metadata_type.pk)
+
+        # Define the metadata type on the adapter so that can be used on the models crud operations
+        self.metadata_type = metadata_type
--- a/nautobot_ssot/contrib/model.py
+++ b/nautobot_ssot/contrib/model.py
@@ -4,6 +4,7 @@
 # Diffsync relies on underscore-prefixed attributes quite heavily, which is why we disable this here.
 
 from collections import defaultdict
+from datetime import datetime
 from typing import ClassVar, Optional
 from uuid import UUID
 
@@ -14,6 +15,7 @@ from django.core.exceptions import MultipleObjectsReturned, ValidationError
 from django.db.models import Model, ProtectedError
 from nautobot.extras.choices import RelationshipTypeChoices
 from nautobot.extras.models import Relationship, RelationshipAssociation
+from nautobot.extras.models.metadata import ObjectMetadata
 from typing_extensions import get_type_hints
 
 from nautobot_ssot.contrib.types import (
@@ -105,6 +107,12 @@ class NautobotModel(DiffSyncModel):
             cls._update_obj_with_parameters(obj, parameters, adapter)
         except ObjectCrudException as error:
             raise ObjectNotCreated(error) from error
+
+        if adapter.metadata_type:
+            try:
+                cls._update_obj_metadata(obj, parameters, adapter)
+            except ObjectCrudException as error:
+                raise ObjectNotCreated(error) from error
 
         return super().create(adapter, ids, attrs)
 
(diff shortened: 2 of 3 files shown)


nautobot_ssot/contrib/adapter.py
```diff
@@ -388,3 +390,25 @@ def _handle_foreign_key(database_object, parameter_name):
...
+        metadata_type, created = MetadataType.objects.get_or_create(
+            name=f"Last sync from {metadata_type__name}",
+            defaults={
+                "data_type": "datetime",
+                "description": f"Timestamp of the last sync from the Data source {metadata_type__name}",
+            },
+        )
+
+        # Use prefetch_related for content_types since they'll be used during model CRUD operations.
+        if not created:
```
Kircheneer, line 410: Why did you decide against attaching the content types right here?

nautobot_ssot/contrib/model.py
```diff
@@ -245,6 +253,10 @@ def _update_obj_with_parameters(cls, obj, parameters, adapter):
         # Set many-to-many fields after saving.
         cls._set_many_to_many_fields(relationship_fields["many_to_many_fields"], obj)
 
+    @classmethod
+    def _update_obj_with_metadata(cls, obj, parameters, adapter):
```
Kircheneer, line 257: What is the purpose of this method?

nautobot_ssot/contrib/model.py
```diff
@@ -388,3 +400,22 @@ def _lookup_and_set_foreign_keys(cls, foreign_keys, obj, adapter):
...
             setattr(obj, field_name, related_object)
+
+    @classmethod
+    def _update_obj_metadata(cls, obj, parameters, adapter):
+        """Update a given Nautobot ORM object with the required object metadata."""
+        # Check if the objects Content Type is included in Metadata Type Content Types
+        obj_content_type = (obj._meta.app_label, obj._meta.model_name)
+        metadata_type_content_types = adapter.metadata_type.content_types.values_list("app_label", "model")
+        if obj_content_type not in metadata_type_content_types:
+            adapter.metadata_type.content_types.add(ContentType.objects.get_for_model(obj._meta.model))
```
Kircheneer, line 411: I guess this is code we would omit if we chose to attach the content types directly to the model in the creation process

nautobot_ssot/contrib/model.py
```diff
@@ -388,3 +400,22 @@ def _lookup_and_set_foreign_keys(cls, foreign_keys, obj, adapter):
...
+
+        # Get the scope_fields from the DiffSync Model
+        obj_metadata_scope_fields = list({parameter.split("__", maxsplit=1)[0] for parameter in parameters.keys()})
+
+        # Create the Object Metadata for the obj
+        obj_metadata = ObjectMetadata(
```
Kircheneer, line 417: Don't we have to to update or create here? What happens if the metadata already exists?

nautobot_ssot/contrib/model.py
```diff
@@ -388,3 +400,22 @@ def _lookup_and_set_foreign_keys(cls, foreign_keys, obj, adapter):
...
+    def _update_obj_metadata(cls, obj, parameters, adapter):
+        """Update a given Nautobot ORM object with the required object metadata."""
+        # Check if the objects Content Type is included in Metadata Type Content Types
+        obj_content_type = (obj._meta.app_label, obj._meta.model_name)
+        metadata_type_content_types = adapter.metadata_type.content_types.values_list("app_label", "model")
+        if obj_content_type not in metadata_type_content_types:
+            adapter.metadata_type.content_types.add(ContentType.objects.get_for_model(obj._meta.model))
+
+        # Get the scope_fields from the DiffSync Model
+        obj_metadata_scope_fields = list({parameter.split("__", maxsplit=1)[0] for parameter in parameters.keys()})
```
Kircheneer, line 414: Ideally we would compute this once somewhere, like pull the scoped fields off of the models at the beginning of a run, possibly with a cached property on the models themselves?

nautobot_ssot/jobs/base.py
```diff
@@ -179,7 +181,12 @@ def record_memory_trace(step: str):
         )
         if memory_profiling:
             record_memory_trace("target_load")
-
+        # Check if the adapter is an instance of NautobotAdapter to determine if it's a contrib implementation,
+        # in which case we should create the required MetadataType object.
+        if self.__class__.__name__ in settings.PLUGINS_CONFIG.get("nautobot_ssot", {}).get(
+            "enable_metadata_for", []
+        ) and isinstance(self.target_adapter, NautobotAdapter):
+            self.target_adapter.get_or_create_metadatatype()
```
Kircheneer, line 189: I see now the difficulty in making this a property, maybe you can still make it work but the current approach is still okay for me.

nautobot_ssot/contrib/adapter.py
michalis1: Because when the MetadataType is first created, there are no ContentType objects attached to it yet  since ContentTypes are populated during model CRUD operations.
However, I think I should attach them anyway to avoid an extra database lookup, even if it’s only needed during the first run of the Job. For all subsequent runs, the MetadataType will already exist.

nautobot_ssot/contrib/model.py
michalis1: This method is called after `_update_obj_with_parameters()` in `create()` method.

nautobot_ssot/contrib/model.py
michalis1: You're absolutely right!!

nautobot_ssot/contrib/model.py
michalis1: The `_update_obj_metadata` method will be used both during the creation and the update of an object. If the content_type is already included in metadata_type_content_types, no update will occur.

That’s why I chose to represent ContentType as a tuple (app_label, model_name) — to avoid multiple database lookups by using ContentType.objects.get_for_model() repeatedly.

nautobot_ssot/contrib/model.py
michalis1: I calculated them in the adapter using a new attribute called metadata_scope_fields.

nautobot_ssot/contrib/model.py
michalis1: I set the content_types for the MetadataType in the adapter.

nautobot_ssot/contrib/adapter.py
michalis1: I ended up attaching them right here 🙂

nautobot_ssot/contrib/model.py
Kircheneer: I am asking because its implementation is simply `pass`

nautobot_ssot/contrib/adapter.py
cdtomkins, line 1: Tested well in my environment, except:

Currently, child objects (e.g. interfaces on devices) can cause a `KeyError`:
```
Exception has occurred: KeyError       (note: full exception trace is shown but execution is paused at: create)
<class '<SNIP>.diffsync.models.base.InterfaceModel'>
  File "<SNIP>/nautobot_ssot/contrib/model.py", line 410, in _update_obj_metadata
    obj_metadata_scope_fields = adapter.metadata_scope_fields[cls]
                                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^
```
Adding `interface` to `top_level` clears the results in `ObjectAlreadyExists` exceptions due to duplicated imports, it's not really a solution.

As discussed this will need resolving, thanks!

nautobot_ssot/contrib/model.py
michalis1: Yep you're right. I deleted this method!!

nautobot_ssot/contrib/adapter.py
michalis1: Fixed!

cdtomkins (approved):
Looks good and worked for me in a lab env.

Kircheneer (approved):

Issue #708, opened 2025-02-24
SSOT - Catalyst Center (DNA Center) sync improvement


PiotrOrman:
### Environment
* Nautobot version:  2.4.2
* nautobot-ssot version:  3.5.0

### Proposed Functionality
Ability to correctly sync devices under the same building name in DNAC.

### Use Case
   Catalyst Center (Dna Center) Allow by design use the same building name located in different Areas. We can have Area name which correspond to for example city name . Area can have two buildings called Plant and Office. This scenario can be multiple by Areas number. There is also Scenario where Buildings can have multiple floors named exactly the same in multiple building (Like Floor 1 or Floor 2)
SSOT should check Area name to which builiding belongs and assign network devices to correct location in Nautobot. Currently Nautobot is checking only Area \ Building \ Floor name independently and it is assigning all devices to first location name which match the one from DNAC.

jdrew82:
Just to be clear, currently the uniqueness for Areas are the Area name and the associated parent Area if there is one. For Buildings it's the Building name and Area so as long as the Area name is different the Building name can be the same. For Floors we automatically add the Building name to ensure uniqueness while allowing for repeat names like you described.

PiotrOrman:
Hello

This is exactlyh my scenarion. Differetn Area name but the same building name. 

![Image](https://github.com/user-attachments/assets/0a871485-681c-4686-8d46-e2cc6c1acb27)

And all devices from multpile PLant-Business Buildings in DNAC assigned just to one location in Nautobot

![Image](https://github.com/user-attachments/assets/56a7c55a-28b3-47a0-9453-73fff0a82224)

jdrew82:
@PiotrOrman I see the problem now. That should be an easy enough fix.

PiotrOrman:
OK i checked your patch, and look like Bulidng name issue is solved now, however the same scenario may happen on floor name. So if there is the same floor name all de devices are assinged to one floor locationin nautbot.

PS. not sure how you created patch branch but it still has issue with location. DNA Center Location Structure Incorrect #687

jdrew82:
@PiotrOrman thBks for confirming that the building issue is fixed and confirming the floors were an issue. I had a feeling on that. I'll get a fix for the floors. Can you clarify your last comment? Are you saying that the location structure still isn't matching what it is in DNAC? It could be ordering of Areas being imported if you have more levels than my instance has.

PiotrOrman:
[...]

PR #777, opened 2025-04-16
Few DNAC Tweaks

jdrew82:
This PR includes a few fixes:

- Corrects location names for Device Building, parent Area, and Location to use dnac_location_map so it uses updated names if using the location_map function from the Job form.
- It reverts the change made in 724 of adding mask_length as an identifier to IPAddress as that was incorrectly done. There should only be one IPAddress object of a certain host address.
- I also added additional unit tests confirming end to end functionality of the Job with a multi-level location structure where locations use repeated naming. There's also confirmation of all Locations being created as expected.
 
# Closes: #708

## What's Changed
This PR includes 

## To Do
Still need to do some further testing of user's location config and without Global being imported.

- [X] Unit, Integration Tests


tasks.py
```diff
@@ -283,7 +283,7 @@ def lock(context, check=False, constrain_nautobot_ver=False, constrain_python_ve
         except UnexpectedExit:
             print("Unable to add Nautobot dependency with version constraint, falling back to git branch.")
             command = (
-                f"poetry add --lock git+https://github.com/nautobot/nautobot.git#{context.nautobot_ssot.nautobot_ver}"
+                f"poetry add --lock git+https://github.com/nautobot/nautobot.git#v{context.nautobot_ssot.nautobot_ver}"
```
smk4664, line 286: Just want to call out that this is not meant to work for the tagged versions such as `v2.0.0` instead this was done to allow installing from `develop` or `next` which do not start with the `v`. I think we should add more details on the error message as the error is very misleading.

jdrew82: This was done just as a test to see if it'd at least get me a bit further.

bile0026 (approved):
LGTM!

Issue #784, opened 2025-04-22
_handle_typed_dict cannot be overwritten


julmanglano:
### Environment
* Nautobot version:  2.4.0
* nautobot-ssot version:  3.6.0

### Proposed Functionality
We have _handle_to_many_relationship method that is defined as a staticmethod. However, it uses _handle_typed_dict directly from its own class NautobotAdapter.

Therefore, an adapter would inherit NautobotAdapter cannot overwrite this method because it would still be the one from its parent that is called.

### Use Case
Why would I need to inherit it ?

Pursuing my quest to be able to target custom field of foreign keys in the NautobotAdapter, I need to rewrite _handle_typed_dict to look for custom field when the annotation  is there.
Because today, TypedDict object cannot look for custom fields in anyway.
Already discussed with @Kircheneer

PR #785, opened 2025-04-22
Overwrite _handle_typed_dict

julmanglano:
# Closes: #784 

## What's Changed

We have _handle_to_many_relationship method that is defined as a staticmethod. However, it uses _handle_typed_dict directly from its own class NautobotAdapter.

Therefore, an adapter would inherit NautobotAdapter cannot overwrite this method because it would still be the one from its parent that is called.

### DIFF
--- a/changes/784.fixed
+++ b/changes/784.fixed
@@ -1,0 +1,1 @@
+Change _handle_to_many_relationship from static method to instance method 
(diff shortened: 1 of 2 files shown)


jdrew82 (approved):
LGTM. @Kircheneer you see any issues with this change?

PR #788, opened 2025-04-22
Fixes #787: Add SKIP_UNMATCHED_DST to slurpit sync

### LINKED ISSUE #787: Slurpit job automatically deletes unrelated objects
### Expected Behavior
Sync into Nautobot from Slurpit without deleting items on Nautobot that are not present in Slurpit. As there may be multiple slurpit instances for a network or data relating to other areas of the business. 

### Observed Behavior
Objects are deleted if they do not match in slurpit.

lpconsulting321:
# Closes: #787 

## What's Changed
add `SKIP_UNMATCHED_DST` to slurpit sync job so that it does not delete existing objects. 

## To Do

- [ ] Unit, Integration Tests

### DIFF
--- a/changes/787.fixed
+++ b/changes/787.fixed
@@ -1,0 +1,1 @@
+Add `SKIP_UNMATCHED_DST` to Slurpit sync.
--- a/nautobot_ssot/integrations/slurpit/jobs.py
+++ b/nautobot_ssot/integrations/slurpit/jobs.py
@@ -2,6 +2,7 @@
 """Slurpit DataSource job class."""
 
 import slurpit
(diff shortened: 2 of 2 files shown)


jdrew82:
You sure you want to set this at the Job level and not on individual models or allow the user to specify what things are import only?

lpconsulting321:
Yes job level. We have a customer using Nautobot and this is causing devices, interfaces and more to be deleted from Nautobot that are not present in Slurpit. This should solve their issues doing import only. Once we have feedback we could implement a more elegant solution or toggle for this setting.

jdrew82 (approved):
LGTM!

PR #792, opened 2025-05-01
DRAFT: Contrib Overhaul

Renrut5:
Overhaul of contrib code, particularly `NautobotAdapter` and `NautobotModel`, to make the code more SOLID compliant.


Renrut5:
@Kircheneer adding this PR and you as a reviewer for visibility as I work on this. This has been something I've been wanting to do for a while.

nautobot_ssot/contrib/cache.py
Renrut5, line 1: Moved cache to it's own, separate interface and basic implementation.

Having the cache functionality integrated into the NautobotAdapter class violates the Single-Responsibility Principle (SRP) and adds additional complexity into the class.

Separating the cache and injecting it into the Adapter class satisfies the SRP and easier to maintain. Additionally, this leaves the cache class open for re-use in custom implementations and for extension where needed.

nautobot_ssot/contrib/exceptions.py
Renrut5, line 1: Custom exceptions for use in the caching classes.

nautobot_ssot/contrib/model.py
Renrut5, line 1: Adapter class had a method which combined two of the fields from a model class. 

```python
class NautobotAdapter(Adapter):
   ...
    @staticmethod
    def _get_parameter_names(diffsync_model):
        """Ignore the differences between identifiers and attributes, because at this point they don't matter to us."""
        return list(diffsync_model._identifiers) + list(diffsync_model._attributes)  # pylint: disable=protected-access
```

All data acted upon in this method is only from the model class and there is no action on data from the `NautobotAdapter` class. Additionally, the data acted upon are protected attributes and direct access to those attributes should generally be avoided outside of the class itself. 

Since this functionality is strictly upon the model data, the method should be moved to the model class as a public method.

nautobot_ssot/contrib/adapter.py
Renrut5, line 1: This is where most of the changes are currently.

Caching class
Moved to a separate file and implementation, then injected into the NautobotAdapter class. More info in the caching file comment.

Update Handle Single Parameter Method
The def _handle_single_parameter method was a more complex than it needed to be. I split it up into different sections:

New method (def _get_parameter_meta) to determine the parameter type as identified in a new options class (class ParameterType).
Then dynamically calls the appropriate method to load the parameter based on the value of the parameter type which returns the value of that particular parameter.
Additionally, these are updated to return values versus updating a variable that was passed by reference. Passing a variable by reference through numerous methods adds additional, unneeded complexity while making the other methods more easily testable.

Update load single object method
The def _load_single_object method passes a parameters variable by reference, as mentioned in the previous point. I changed this to instead using the called methods (originally _handle_single_parameter) to get a value and assigning it in this method vs passing by reference. This makes for cleaner, more easily readable code than passing by reference and making it more maintainable/extendable.

Issue #457, opened 2024-05-20
Custom relationships items ordering is not ignored when calculating changes


gioccher:
This is an issue in ssot:contrib CustomRelationshipAnnotation with custom relationships items ordering

### Environment
* Python version:  3.9.19
* Nautobot version:  2.2.3
* nautobot-ssot version:  2.6.1

### Expected Behavior
No changes are detected after running the sync a 2nd time

### Observed Behavior
The same set of changes is detected running the sync multiple times. The only difference is the order of items in the custom relationship.

### Steps to Reproduce
1. define a many-to-many relationship and a ssot job that syncs from external to nautobot
2. add multiple elements to the relationship in the external system
3. run the ssot sync. This populates the custom relationship with items, but *sometimes* in a different order from what's defined in the external system
4. run the ssot sync again. This second sync should not detect any changes, but *sometimes* you'll see the same list of detected changes in the Data Sync tab, and ssot attempting to apply the update again in the Sync Logs tab

![image](https://github.com/nautobot/nautobot-app-ssot/assets/921538/15df8e75-49fc-4cd7-91c3-6f02d1324eea)

![image](https://github.com/nautobot/nautobot-app-ssot/assets/921538/80f14e7c-dba8-4eae-939c-33b730013523)

Kircheneer:
Interesting point. There is https://github.com/networktocode/diffsync/issues/25 which would address this, until then, if you need a quick fix for this, I recommend you do something like

```
class YourAdapter(NautobotAdapter):
    ...

    def load(...):
        return_value = super.load(...)
        for obj in self.get_all("modelname"):
            obj.field = sorted(obj.field, key=lambda o: o.trunk_groups)
            self.update(obj)
        return return_value
```

This should sort your values to be consistent - you need to change this such that it matches your other adapter, or change the other adapter, too.

gioccher:
Thanks, the quick fix works
```
class YourAdapter(NautobotAdapter):
    ...
    def load(self):
        super().load()

        for obj in self.get_all(self.interface):
            obj.trunk_groups = sorted(obj.trunk_groups, key=operator.itemgetter("name"))
            self.update(obj)
```

PR #806, opened 2025-05-09
Fix gRPC Exception in CVP Integration

jdrew82:
# Closes: #457

## What's Changed
This PR is to try to fix the gRPC error being seen when getting devices in the CVP integration.

## To Do
Still need to finish testing and confirm functionality.

- [X] Added change log fragment(s) (for more information see [the documentation](https://docs.nautobot.com/projects/core/en/stable/development/#creating-changelog-fragments))
- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot_ssot/integrations/aristacv/utils/cloudvision.py
+++ b/nautobot_ssot/integrations/aristacv/utils/cloudvision.py
@@ -242,7 +242,7 @@ class CloudvisionApi:  # pylint: disable=too-many-instance-attributes, too-many-
         return (self.decode_batch(nb) for nb in res)
 
 
-def get_devices(client, import_active: bool):
+def get_devices(client, logger, import_active: bool):
     """Get devices from CloudVision inventory."""
     device_stub = services.DeviceServiceStub(client)
     if import_active:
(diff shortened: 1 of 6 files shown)


nautobot_ssot/integrations/aristacv/utils/cloudvision.py
```diff
@@ -253,17 +253,20 @@ def get_devices(client, import_active: bool):
...
+                "device_id": resp.value.key.device_id.value,
+                "hostname": resp.value.hostname.value,
+                "fqdn": resp.value.fqdn.value,
+                "sw_ver": resp.value.software_version.value,
+                "model": resp.value.model_name.value,
+                "status": "Active" if resp.value.streaming_status == 2 else "Offline",
+                "system_mac_address": resp.value.system_mac_address.value,
+            }
+            devices.append(device)
+    except grpc.RpcError as err:
```
chadell, line 268: what is raising the `grpc.RpcError`? I thought it could be `device_stub.GetAll(req)` instead of the response value access...

jdrew82: @chadell I thought the same thing but when I put the block around that it didn't actually catch it. It seems to be only when you're iterating through the messages, or I guess viewing the stream, that the error is thrown. In my particular test case it was an authentication error though I'm not sure what the other opened issue is having the error from.

qduk (approved):

Issue #859, opened 2025-06-20
Infoblox SSoT - cache misses due to the caching between sync runs


progala:
### Environment
* Python version:  3.11.11
* Nautobot version:  2.4.6
* nautobot-ssot version:  3.8.1

### Expected Behavior

Subsequent runs of Infoblox -> Nautobot syncs complete successfully.

### Observed Behavior

First Infoblox -> Nautobot sync succeeds, second one fails.

### Steps to Reproduce
1. Synchronize a prefix with IPs from Infoblox to Nautobot using the SSoT job
2. Delete from Nautobot one of the previously synchronized IP addresses
3. Attempt to run the Infoblox -> Nautobot SSoT job with the same parameters
4. See job fail with exception "IPAddress matching query does not exist."

progala:
This is caused by class attributes used for caching Nautobot objects loaded in the adapter and created during the sync. Values assigned to these attributes persist between the SSoT job runs. This means that if one of the Nautobot objects is deleted but its UUID still exists in the cache we can get a cache miss.

PR #861, opened 2025-06-20
Infoblox integration - Nautobot adapter cache fix

progala:
# Closes: #859

## What's Changed

Adds logic to clear the cache when initializing the Nautobot adapter in the Infoblox integration. This cache may persist between sync runs and lead to bugs caused by cache misses.


jdrew82 (approved):
LGTM!

Issue #867, opened 2025-06-25
ServiceNow - Swallow duplicate exceptions and produce error report


msheiny:
### Environment
* Nautobot version:  `2.4.6`
* nautobot-ssot version:  `3.6.0`

### Proposed Functionality

The current functionality today when dealing with a table that includes a parent element (`product_model` for instance) is that the code will throw an exception if a duplicate object is found in ServiceNow. The user is just shown the raw exception and it's pretty jarring to grok what to do next. It's not clear why these exceptions aren't just logged and swallowed because that's the behavior on other tables such as `core_company` and `location`.

I'm proposing we swallow these exceptions, log them, and produce a report for users at the end where they can download the results of ALL duplicates so they can go rectify them.

### Use Case

You are onboarding to nautobot, not familiar with the quirks of the nautobot ecosystem, you run this servicenow SSOT application. Run sync. The job gets to the `product_model` table where you have 500 duplicates. The current logic will throw an exception on the first duplicate encountered. You fix that duplicate. New exception on the next one.

Instead imagine that you run the sync attempt, you get warnings in the log lines and then at the end of the syncronization attempt - you are given a number of files that include the duplicate elements in each table that need to go be rectified in ServiceNow.

PR #871, opened 2025-06-26
SNOW - Gracefully handle duplicates, repair table_query feature

msheiny:
# Closes: #867 

## What's Changed

* Passes through the `table_query` attribute from the servicenow `mappings.yaml`. This is called out in the comments - `Only load companies that are flagged as manufacturer: true, as those correspond to Nautobot Manufacturer records` but in implementation this value was not being referenced anywhere so it was a noOp. I repaired the connection and tested locally to see that it properly only brings in company names that are listed as manufacturers.
* Added file output to give user an indication of all duplicates encountered in each servicenow table.
* Behavior is now to swallow, log overview of how many duplicates encountered, and provide file output outlining what duplicates were encountered.

## To Do

- [ ] Unit, Integration Tests

### DIFF
--- a/changes/867.fixed
+++ b/changes/867.fixed
@@ -1,0 +1,3 @@
+- Gracefully swallow ServiceNow exceptions
+- Adds ServiceNow duplicate file reports
+- Fixes ServiceNow comparison filters to only compare against company names with Manufacturer set to True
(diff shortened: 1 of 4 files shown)


jdrew82 (approved):
LGTM!

PR #880, opened 2025-07-08
bootstrap integration: validatedSW, auto-add missing SW

nopg:
When adding ValidatedSoftware, if the software version specified is missing, it will automatically add the software as well.


jdrew82 (approved):
LGTM!

bile0026:
@jdrew82  Can we unmerge this? I'm not sure we want to set the precedent on this.

Issue #904, opened 2025-07-12
Bootstrap fails with missing Software Version


nopg:
### Environment
* Python version:  3.11.4
* Nautobot version:  2.4.10
* nautobot-ssot version:  3.8.2a0

A log that the software version is missing, but the Job continues running or fails more gracefully.

Failed with exception:
```
{
    "exc_message": [
        "SoftwareVersion matching query does not exist."
    ],
    "exc_module": "nautobot.dcim.models.devices",
    "exc_type": "SoftwareVersion.DoesNotExist"
}
```

### Steps to Reproduce
1. Clone repo as-is
2. Modify a Validated Software to have a missing Software Version, i.e:
```
validated_software:
  - software: "arista_eos - 4.25.10Missing"
 ```
3. Run the Bootstrap Job

Full Traceback:
```
Traceback (most recent call last):
  File "/usr/local/lib/python3.11/site-packages/celery/app/trace.py", line 477, in trace_task
    R = retval = fun(*args, **kwargs)
                 ^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/celery/app/trace.py", line 760, in __protected_call__
    return self.run(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/nautobot/extras/jobs.py", line 1286, in run_job
    result = job(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/nautobot/extras/jobs.py", line 163, in __call__
    return self.run(*args, **deserialized_kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/source/nautobot_ssot/integrations/bootstrap/jobs.py", line 121, in run
    super().run(dryrun=self.dryrun, memory_profiling=self.memory_profiling, *args, **kwargs)
  File "/source/nautobot_ssot/jobs/base.py", line 367, in run
    self.sync_data(memory_profiling)
  File "/source/nautobot_ssot/jobs/base.py", line 209, in sync_data
    self.execute_sync()
  File "/source/nautobot_ssot/jobs/base.py", line 100, in execute_sync
    self.source_adapter.sync_to(self.target_adapter, flags=self.diffsync_flags)
[...]

PR #905, opened 2025-07-12
fix: bootstrap validated_software missing version

nopg:
# Closes: #904

## What's Changed
Adds a Warning log message that Validated Software is unable to find the SoftwareVersion it points to thus unable to be created, and handles the Exceptions so the Job can continue to run.


bile0026 (approved):
LGTM!

jdrew82 (approved):
LGTM!

Issue #881, opened 2025-07-08
Bootstrap fails with missing Secret


nopg:
### Environment
* Python version:  3.11.4
* Nautobot version:  2.4.10
* nautobot-ssot version:  3.8.2a0

### Expected Behavior
Logs that the secret is missing, but the Job continues or at least fails gracefully.

### Observed Behavior
Failed with exception:
```
{     "exc_message": [         "cannot access local variable '_orm_secret' (redacted) it is not associated with a value"     ],     "exc_module": "builtins",     "exc_type": "UnboundLocalError" }
```

### Steps to Reproduce
1. Create a Secrets Group in global-settings.yml
2. Point to a non-existing Secret
3. Run Bootstrap Sync job

Full Traceback:
```
Traceback (most recent call last):
  File "/usr/local/lib/python3.12/site-packages/nautobot_ssot/integrations/bootstrap/diffsync/models/nautobot.py", line 1960, in update
    _update_group.secrets.get(name=_secret["name"])
  File "/usr/local/lib/python3.12/site-packages/django/db/models/manager.py", line 87, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/django/db/models/query.py", line 637, in get
    raise self.model.DoesNotExist(
nautobot.extras.models.secrets.Secret.DoesNotExist: Secret (redacted) query does not exist.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib/python3.12/site-packages/celery/app/trace.py", line 477, in trace_task
    R = retval = fun(*args, **kwargs)
                 ^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/celery/app/trace.py", line 760, in __protected_call__
    return self.run(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/nautobot/extras/jobs.py", line 1286, in run_job
    result = job(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/nautobot/extras/jobs.py", line 163, in __call__
    return self.run(*args, **deserialized_kwargs)
[...]

PR #906, opened 2025-07-13
fix: bootstrap missing Secret in SecretsGroup

nopg:
# Closes: #881

## What's Changed
All related only to Bootstrap integration:
- Fixes early return when creating SecretsGroups
- Fixes exception occurring if Secret within a SecretGroup getting created does not actually exist within Nautobot.
- Fixes issue with incorrect SecretsGroupAssociation field being used.

## To Do
Since Secrets are not required in order to create a SecretGroup, and looking at the original error and assuming intent from that, I believe it's OK to still create the Secret Group just without the missing Secret, but logging as warning to ensure the end-user is able to notice this issue.

### DIFF
--- a/changes/881.fixed
+++ b/changes/881.fixed
@@ -1,0 +1,1 @@
+Fixed exception caused by missing secret value when creating a SecretsGroup with Bootstrap.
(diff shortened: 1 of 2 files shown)


bile0026 (approved):
LGTM!

jdrew82 (approved):
LGTM!

Issue #877, opened 2025-07-02
Add Abstract Base Classes for Contrib adapter and models


Renrut5:
### Proposed Functionality

Create Abstract Base Classes for `NautobotAdapter` and `NautobotModel` in `contrib.

Goals are to:

- Define expected relationships between models and adapters
- Enable ability to create a custom adapter or model and use the other from contrib and reduce coupling in other libraries

PR #909, opened 2025-07-16
Add abstract base classes for contrib adapter and models

Renrut5:
Closes: #877

Added base classes for `NautobotAdapter` and `NautobotModel` to identify and model the expected interactions between the two. The reasoning is because other libraries have specific expectations of how one or the other might work and should have an "interface" style class to base their development on.

Currently only adding the classes without implementing in existing adapter or model to allow for additional changes and other feedback before full implementation.


nautobot_ssot/contrib/base.py
```diff
@@ -0,0 +1,47 @@
+"""Abstract base classes for primary contrib module classes.
+
+These base classes are meant to define the expected interactions between the various classes
+within contrib to allow for additional customization and extension based on individual needs.
+
+Todo:
+  - Add abstract & common methods
+  - Update references in adapter and model
+"""
```
Kircheneer, line 9: I like the idea. I would like to see some real examples of customization that is made possible by this pattern and how it is not possible before.

Kircheneer (changes requested):

Issue #878, opened 2025-07-03
vSphere Integration can't sync IPv6


yeyep:
### Environment
* Python version:  3.12
* Nautobot version:  2.4.10
* nautobot-ssot version:  3.9.0

### Expected Behavior

vSphere Integration handles IPv4 and IPv6 during job sync. 

### Observed Behavior

IPv4 is handled correctly.
```
Loading IP Addresses {'mac_address': '00:50:56:8b:0e:6e', 'ip': {'ip_addresses': [{'ip_address': '10.73.199.30', 'prefix_length': 24, 'state': 'PREFERRED'}, {'ip_address':  'fe80::250:56ff:fe8b:e6e', 'prefix_length': 64, 'state': 'UNKNOWN'}]}, 'nic': '4000'}
```

This is the payload that the extenstion trys to handle last before the error occurs:
```
Loading IP Addresses {'mac_address': '00:50:56:8b:a9:f0', 'ip': {'ip_addresses': [{'ip_address': 'fde1:53ba:e9a0:cccc:20c:29ff:fed5:de4b', 'prefix_length': 64, 'state': 'UNKNOWN'}, {'ip_address': 'fe80::250:56ff:fe8b:a9f0', 'prefix_length': 64, 'state': 'UNKNOWN'}]}, 'nic': '4001'}
```
This contains a IPv6.

With this payload we get the following error output from the job.
```
{"exc_message": ["Parameter must be an integer between 0 and 32."],"exc_module": "builtins","exc_type": "ValueError"}
```

The error message in the celery job.
```
  File "/opt/nautobot/.local/lib/python3.12/site-packages/nautobot_ssot/integrations/vsphere/diffsync/adapters/adapter_vsphere.py", line 237, in load_vm_interfaces
    ipv4_addresses, ipv6_addresses = self.load_ip_addresses(
                                     ^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/nautobot/.local/lib/python3.12/site-packages/nautobot_ssot/integrations/vsphere/diffsync/adapters/adapter_vsphere.py", line 157, in load_ip_addresses
    netmask = cidr_to_netmask(ip_address["prefix_length"])
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/netutils/ip.py", line 433, in cidr_to_netmask
[...]

PR #918, opened 2025-07-22
Update vSphere integration to handle IPv6 properly

qduk:
# Closes: #878

## What's Changed

The vSphere integration was not properly accounting for IPv6 prefix calculation when it deduced a prefix from a given IP address. I've gone and updated the logic so that it can properly create an IPv6 prefix for a given IPv6 address. 

## To Do

- [X] Unit, Integration Tests

### DIFF
--- a/nautobot_ssot/integrations/vsphere/diffsync/adapters/adapter_vsphere.py
+++ b/nautobot_ssot/integrations/vsphere/diffsync/adapters/adapter_vsphere.py
@@ -6,7 +6,7 @@ import ipaddress
 from typing import List
 
 from diffsync import Adapter
-from netutils.ip import cidr_to_netmask
+from netutils.ip import cidr_to_netmask, cidr_to_netmaskv6
 
 from nautobot_ssot.integrations.vsphere.diffsync.models.vsphere import (
     ClusterGroupModel,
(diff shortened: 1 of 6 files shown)


tasks.py
```diff
@@ -918,6 +919,8 @@ def unittest(  # noqa: PLR0913
         command += f" -k='{pattern}'"
     if verbose:
         command += " --verbosity 2"
+    if pdb:
```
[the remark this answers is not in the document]
qduk: Good point. Let me move it over.

nautobot_ssot/integrations/vsphere/diffsync/adapters/adapter_vsphere.py
```diff
@@ -18,22 +18,28 @@
 )
 
 
-def deduce_network_from_ip(ip: ipaddress.IPv4Address, subnet_mask: str):
-    """Figure out the network of a given IP and its subnet mask."""
-    ip_int = int(ip)
-    subnet_int = int(ipaddress.IPv4Address(subnet_mask))
+def deduce_network_from_ip(ip: str, subnet_mask: str) -> str:
```
jdrew82, line 21: Isn't this available through netutils? If it's not, it probably should be a function there.

jdrew82 (approved):
Just a small comment. Looks good otherwise.

nautobot_ssot/integrations/vsphere/diffsync/adapters/adapter_vsphere.py
qduk: So, in digging deeper, it looks like that functionality is provided by IPAddress library. I refactored to use that instead.

PR #923, opened 2025-07-25
Remove expired A and PTR records left when HTTP 404 is seen

### LINKED ISSUE #922: Expired A and PTR records left when HTTP 404 is seen
### Environment
* Python version:  3.11
* Nautobot version:  2.4.13
* nautobot-ssot version:  3.9.1

Continues from [#596](https://github.com/nautobot/nautobot-app-ssot/issues/596) and #919 

Previously the job would fail on HTTP 404 for A and PTR records when seen (unhandled exception). This was changed to log the event instead of failing in #919.

The issue is that there still may be "stale" records that need removed in Nautobot?

### Expected Behavior
After logging the event, it should attempt to remove the stale object (if it exists)

### Observed Behavior
Currently we are just logging the event

Reproducing steps are the same as #596

MeganerdDev:
# Closes: #922 

## What's Changed

We will attempt to remove "stale" records when we see the HTTP 404 for A and PTR records

## To Do

- [ ] Unit, Integration Tests

### DIFF
--- a/changes/922.fixed
+++ b/changes/922.fixed
@@ -1,0 +1,1 @@
+Remove stale A and PTR records when HTTP 404 is seen.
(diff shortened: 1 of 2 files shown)


jdrew82 (changes requested):
@MeganerdDev can you address the linter and ensure tests are passing and we have coverage for these new bits? Also, I think this should be behind a setting for the integration so that users choose this behavior. We don't want to surprise them if data goes missing unexpectedly.

Issue #926, opened 2025-07-28
Issue with metadata_type when contrib models are used without the contrib adapter.


michalis1:
### Environment
* Python version:  3.11.4
* Nautobot version:  2.4.12
* nautobot-ssot version:  3.9.1

### Expected Behavior
Issue with `metadata_type` when contrib models are used without the corresponding contrib adapter.

### Observed Behavior
When using contrib models without the contrib adapter, the `metadata_type` attribute is missing from the adapter instance during model [creation](https://github.com/nautobot/nautobot-app-ssot/blob/a20baa131ab1c30d51d24f8cc8ea43fc92cf3e2b/nautobot_ssot/contrib/model.py#L113) or [update](https://github.com/nautobot/nautobot-app-ssot/blob/a20baa131ab1c30d51d24f8cc8ea43fc92cf3e2b/nautobot_ssot/contrib/model.py#L79).

### Steps to Reproduce
1. Run a sync job using contrib models without the contrib adapter, as in the case of nautobot-device-onboarding app.

PR #927, opened 2025-07-28
Fix issue with metadata_type when contrib models are used without the contrib adapter.

michalis1:
# Closes: #926

## What's Changed

When using contrib models without the contrib adapter, the `metadata_type` attribute is missing from the adapter instance during model [creation](https://github.com/nautobot/nautobot-app-ssot/blob/a20baa131ab1c30d51d24f8cc8ea43fc92cf3e2b/nautobot_ssot/contrib/model.py#L113) or [update](https://github.com/nautobot/nautobot-app-ssot/blob/a20baa131ab1c30d51d24f8cc8ea43fc92cf3e2b/nautobot_ssot/contrib/model.py#L79). To fix this, I first check whether the `metadata_type` attribute exists on the adapter instance, and then evaluate its value if present.

### DIFF
--- a/changes/926.fixed
+++ b/changes/926.fixed
@@ -1,0 +1,1 @@
+Fix issue with metadata_type when contrib models are used without the contrib adapter.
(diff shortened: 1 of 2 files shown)


joewesch (approved):
Thanks for the quick fix!

Issue #723, opened 2025-03-11
Nautobot Adapter fails to load into DiffSync model CustomFieldAnnotation if only "key" is set


qduk:
### Environment
* Python version:  
* Nautobot version:  2.3.1
* nautobot-ssot version:  3.5.1

### Expected Behavior
The Nautobot adapter to load the values for CustomFields into the diffsync model.

### Observed Behavior
Nautobot Adapter fails to load the custom field values into the diffsync model.

### Steps to Reproduce
1. Create diffsync model with CustomFieldAnnotation only setting the `key` field on the Annotation
2. Run SSoT in an attempt to load objects from Nautobot
3.

qduk:
It is due to [this line](https://github.com/nautobot/nautobot-app-ssot/blob/a65810b8a6d7f0e926fbd1586762fe2d592a890a/nautobot_ssot/contrib/adapter.py#L94). I think we could solve this a couple of ways. My current fix is to just check for either `key` or `name` but I think we could change some logic in the [Annotation](https://github.com/nautobot/nautobot-app-ssot/blob/a65810b8a6d7f0e926fbd1586762fe2d592a890a/nautobot_ssot/contrib/types.py#L50)

cdtomkins:
Still experiencing this issue as of June 2025 with the latest `develop` branch. `__post_init__()` method still only handles setting key from name but not the reverse.

I can see there is already a WIP PR for this.

cdtomkins:
The fix I'm working with in my dev environment in `_handle_single_parameter()` is simply:
```
                # if metadata.name in database_object.cf:
                    # parameters[parameter_name] = database_object.cf[metadata.key]

                field_key = metadata.key or metadata.name
                if field_key in database_object.cf:
                    parameters[parameter_name] = database_object.cf[field_key]
```

cdtomkins:
I'm going to go ahead and PR to fix this as the previous PR seems to have stopped moving forward.

Given this intent from the docstring:

```
    Note that for backwards compatibility purposes it is also possible to use `CustomFieldAnnotation.name` instead of
    `CustomFieldAnnotation.key`.
```
I will PR this *(closer to stated intent)*:
```
                field_key = metadata.key or metadata.name
                if field_key in database_object.cf:
                    parameters[parameter_name] = database_object.cf[field_key]
```
Not this *(closer to current functionality)*:
```
                field_key = metadata.name or metadata.key
                if field_key in database_object.cf:
[...]

PR #929, opened 2025-07-30
Fix _handle_single_parameter metadata `key` vs `name` inconsistency

cdtomkins:
# Closes: #723 

## What's Changed
Adjusted _handle_single_parameter to correctly implement the intent from the `CustomFieldAnnotation` docstring:
```
    Note that for backwards compatibility purposes it is also possible to use `CustomFieldAnnotation.name` instead of
    `CustomFieldAnnotation.key`.
```

## To Do
- [x] Unit, Integration Tests

# BEFORE FIX
Screenshot 2025-08-13 111337

# AFTER FIX
Screenshot 2025-08-13 114040

### DIFF
--- a/changes/723.fixed
+++ b/changes/723.fixed
@@ -1,0 +1,1 @@
+Fixed `_handle_single_parameter` metadata key vs name inconsistency.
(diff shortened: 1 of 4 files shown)


cdtomkins:
Need to add a new UT if possible.

PR #938, opened 2025-08-14
Add Metadata to Bootstrap Integration

nopg:
## What's Changed

This PR adds support for the [Object Metadata](https://docs.nautobot.com/projects/core/en/stable/user-guide/platform-functionality/objectmetadata/) feature that was added in Nautobot 2.3 to the Bootstrap integration. The exact Objects and `scoped_fields` can be found in `constants.py`


nopg:
Updated as requested. Used `app_label.model_name` as the keys for the scoped fields, and created a more 'global' metadata utils that other integrations will be able to use. Used a new file instead of the existing utils.py, due to some import/bootup issues caused by the new django/other imports, and the `utils.py` file being imported elsewhere before all apps were 'ready'.

I believe there needs to be some cleanup for DNA Center Metadata though as we are switching the type from DATE to DATETIME. You cannot modify the type after creation, so any existing Metadata would need to be removed first (or some better way on how to handle this gracefully, or just delete any type DATE that are found?)

jdrew82 (approved):
LGTM!

Issue #596, opened 2024-11-08
Infoblox -> Nautobot job fatally fails on 404 not found


MidnightRamenX:
### Environment
* Python version:  Python 3.9.18
* Nautobot version:  2.3.10
* nautobot-ssot version:  2.8.1

### Expected Behavior
Infoblox -> Nautobot job should gracefully handle a 404 not found on an import job.

### Observed Behavior
Fatal failure with the following error and stack trace.
```
 Value	
{
    "exc_message": [
        "404 Client Error: Not Found for url: https://<infoblox_dn>/wapi/v2.12/record:ptr/ZG5<obscured_ref>ZWR1:<ip_addr>.in-addr.arpa/default?_return_fields=name%2Cptrdname%2Cipv4addr%2Cipv6addr%2Cview%2Ccomment. {'Error': 'AdmConDataNotFoundError: Reference record:ptr/ZG5<obscured_ref>ZWR1:<ip_arr>.in-addr.arpa/default not found', 'code': 'Client.Ibap.Data.NotFound', 'text': 'Reference record:ptr/ZG5<obscured_ref>ZWR1:<ip_arr>.in-addr.arpa/default not found'}"
    ],
    "exc_module": "requests.exceptions",
    "exc_type": "HTTPError"
}
```
```
Traceback (most recent call last):
  File "/opt/nautobot/lib64/python3.9/site-packages/nautobot_ssot/integrations/infoblox/utils/client.py", line 194, in _request
    resp.raise_for_status()
  File "/opt/nautobot/lib64/python3.9/site-packages/requests/models.py", line 1024, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 404 Client Error: Not Found for url: https://<infoblox_dn>/wapi/v2.12/record:ptr/ZG5<obscured_ref>ZWR1:<ip_arr>.in-addr.arpa/default?_return_fields=name%2Cptrdname%2Cipv4addr%2Cipv6addr%2Cview%2Ccomment

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/opt/nautobot/lib64/python3.9/site-packages/celery/app/trace.py", line 477, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/opt/nautobot/lib64/python3.9/site-packages/celery/app/trace.py", line 760, in __protected_call__
    return self.run(*args, **kwargs)
[...]

PR #944, opened 2025-08-25
Patch 596 handleinfoblox404

fpocai:
# Closes: #596 

## What's Changed

This bug was marked "Closed" but is still an ongoing issue.  The calls to logger were using kwarg `message=` when the class was expecting the default kwarg `msg=`.  

I've changed the two places where this is configured in the Infoblox diffsync adapter.


jdrew82 (approved):
LGTM!

PR #946, opened 2025-08-27
Remove expired Host records left when HTTP 404 is seen

### LINKED ISSUE #922: Expired A and PTR records left when HTTP 404 is seen

### LINKED ISSUE #945: Infoblox to Nautobot sync may fail with 404 if a Host Record is removed while the sync is running.
### Environment
* Python version:  3.12.11
* Nautobot version:  2.4.16
* nautobot-ssot version:  3.9.3a0

### Expected Behavior
Synchronization skips the error and continues, cleans up any stale records.

### Observed Behavior
Synchronization fails due to the 404 error.

### Steps to Reproduce
1. Create Host Record on Infoblox
2. Start Sync job.
3. After networks scan, but before Hosts scan, delete the host record.

This is *very* similar to #922 , however, on #923, MeganerdDev only addressed A and PTR records.  This applies their change to Host records.

fpocai:
# Closes: #945

## What's Changed

Apply MeganerdDev's solution from #923 (Fixes #922) to `_load_dns_host_record_for_ip()`.  922 only addresses A and PTR records, but this issue also occurs with Host records.

## To Do

- [ ] Unit, Integration Tests

### DIFF
--- a/nautobot_ssot/integrations/infoblox/diffsync/adapters/infoblox.py
+++ b/nautobot_ssot/integrations/infoblox/diffsync/adapters/infoblox.py
@@ -307,7 +307,31 @@ class InfobloxAdapter(Adapter):
             ip_record (object): Parent IP Address record
             namespace (str): Namespace of this record
         """
-        host_record = self.conn.get_host_record_by_ref(ref)
+        try:
+            host_record = self.conn.get_host_record_by_ref(ref)
+        except requests.exceptions.HTTPError as exc:
+            if exc.response.status_code == 404:
+                self.job.logger.warning(f"Host record {ref} not found, likely dynamic and expired.")
+                # remove any existing stale Host record from Nautobot
+                try:
+                    stale=self.dnshostrecord(
+                        address=ip_record.address,
+                        prefix=ip_record.prefix,
+                        prefix_length=ip_record.prefix_length,
+                        namespace=namespace,
+                        dns_name="",
+                        ip_addr_type=ip_record.ip_addr_type,
+                        description="",
+                        status=ip_record.status,
+                        ext_attrs=record_ext_attrs,
+                        ref=ref,                          
+                    )
+                    self.remove(stale)
+                except Exception:  # noqa: S110
+                    pass
+                return
+            raise                    
+                        
         record_ext_attrs = get_ext_attr_dict(
             extattrs=host_record.get("extattrs", {}), excluded_attrs=self.excluded_attrs
         )
--- a/changes/945.fixed
+++ b/changes/945.fixed
@@ -1,0 +1,1 @@
+Apply MeganerdDev's fix from #923 (Fixes #922) to Host Record lookups.


nautobot_ssot/integrations/infoblox/diffsync/adapters/infoblox.py
```diff
@@ -307,7 +307,31 @@ def _load_dns_host_record_for_ip(self, ref: str, ip_record: object, namespace: s
             ip_record (object): Parent IP Address record
             namespace (str): Namespace of this record
         """
-        host_record = self.conn.get_host_record_by_ref(ref)
+        try:
+            host_record = self.conn.get_host_record_by_ref(ref)
+        except requests.exceptions.HTTPError as exc:
+            if exc.response.status_code == 404:
```
jdrew82, line 313: I'll say same thing to you as I did @MeganerdDev, this needs to be behind a setting for the integration where the user explicitly opts into this behavior. Following principle of least astonishment we don't want the users surprised by records disappearing that they don't expect.

nautobot_ssot/integrations/infoblox/diffsync/adapters/infoblox.py
```diff
@@ -307,7 +307,31 @@ def _load_dns_host_record_for_ip(self, ref: str, ip_record: object, namespace: s
...
+                        namespace=namespace,
+                        dns_name="",
+                        ip_addr_type=ip_record.ip_addr_type,
+                        description="",
+                        status=ip_record.status,
+                        ext_attrs=record_ext_attrs,
+                        ref=ref,                          
+                    )
+                    self.remove(stale)
+                except Exception:  # noqa: S110
```
jdrew82, line 330: This is a bit too broad catch. Can we be more specific on the Exception being caught?

nautobot_ssot/integrations/infoblox/diffsync/adapters/infoblox.py
jdrew82: Also, do you think you could write a unit test to cover this?

jdrew82 (changes requested):
Just a few tweaks.

nautobot_ssot/integrations/infoblox/diffsync/adapters/infoblox.py
jdrew82: I'm not aware of a method of excluding DHCP pools but I could just be missing it.

garymccann:
Will close out this PR as stale, feel free to re-open when the opportunity arises to complete it.

Issue #962, opened 2025-09-10
Meraki get-organization-devices API may omit notes key in response


MeganerdDev:
### Environment
* Python version:  3.11
* Nautobot version:  2.4.2
* nautobot-ssot version:  3.9.3a0

It appears that the Meraki get-organization-devices API may omit notes key in its response

### Expected Behavior
- [get_or_instantiate](https://github.com/nautobot/nautobot-app-ssot/blob/23527c2607003eaec36a6a8de86218d136392ad5/nautobot_ssot/integrations/meraki/diffsync/adapters/meraki.py#L109-L126) to create the missing object during the `load_devices()` diffsync operation

### Observed Behavior
- Got an exception for slicing a missing dictionary key of "notes" on line-116 [here](https://github.com/nautobot/nautobot-app-ssot/blob/23527c2607003eaec36a6a8de86218d136392ad5/nautobot_ssot/integrations/meraki/diffsync/adapters/meraki.py#L116C43-L116C48)

### Steps to Reproduce
1. Configure SSoT App for Meraki
2. Perform a Dry Run of the job
```
Traceback (most recent call last):
  File "/usr/local/lib/python3.11/site-packages/nautobot_ssot/integrations/meraki/diffsync/adapters/meraki.py", line 90, in load_devices
    self.get(self.device, dev["name"])
  File "/usr/local/lib/python3.11/site-packages/diffsync/__init__.py", line 725, in get
    return self.store.get_model_obj_identifier(identifier)
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/diffsync/store/local.py", line 49, in get
    raise ObjectNotFound(f"{modelname} {uid} not present in {str(self)}")
diffsync.exceptions.ObjectNotFound: device 12345 not present in LocalStore

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib/python3.11/site-packages/celery/app/trace.py", line 477, in trace_task
    R = retval = fun(*args, **kwargs)
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/celery/app/trace.py", line 760, in _protected_call_
    return self.run(*args, **kwargs)
  File "/usr/local/lib/python3.11/site-packages/nautobot/extras/jobs.py", line 1288, in run_job
[...]

PR #963, opened 2025-09-10
Use query for notes key on Meraki get-organization-devices API response since key might be omitted

MeganerdDev:
# Closes: #962 

## What's Changed
Since the `notes` key might not always be included in the API response, use a `.get()` instead of a slice

## To Do

- [ ] Unit, Integration Tests


jdrew82 (approved):
LGTM! I hadn't actually seen any instances of that not existing but I could see cases of it so fine with the change.

Issue #964, opened 2025-09-10
[CLOUDVISION SSOT] Sync job Fails when same IP address assigned to multiple interfaces in same Namespace


pato23arg:
### Environment
* Python version:  3.11.4
* Nautobot version:  2.4.2
* nautobot-ssot version:  3.81

### Expected Behavior
Same IP can be assigned to multiple interfaces in same namespace (for example, an Anycast IP)

### Observed Behavior
Sync Job crashes hard:

  File "/usr/local/lib/python3.11/site-packages/nautobot_ssot/jobs/base.py", line 171, in sync_data
    self.load_target_adapter()
  File "/usr/local/lib/python3.11/site-packages/nautobot_ssot/integrations/aristacv/jobs.py", line 110, in load_target_adapter
    self.target_adapter.load()
  File "/usr/local/lib/python3.11/site-packages/nautobot_ssot/integrations/aristacv/diffsync/adapters/nautobot.py", line 202, in load
    self.load_ip_addresses()
  File "/usr/local/lib/python3.11/site-packages/nautobot_ssot/integrations/aristacv/diffsync/adapters/nautobot.py", line 147, in load_ip_addresses
    self.add(new_map)
  File "/usr/local/lib/python3.11/site-packages/diffsync/__init__.py", line 810, in add
    return self.store.add(obj=obj)
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/diffsync/store/local.py", line 107, in add
    raise ObjectAlreadyExists(f"Object {uid} already present", obj)
diffsync.exceptions.ObjectAlreadyExists: ('Object 10.95.225.5/32__Global__DEVICE_TEST_1__Loopback1 already present', ipassignment "10.95.225.5/32__Global__DEVICE_TEST_1__Loopback1")
 

### Steps to Reproduce
1. Assign Same IP address to multiple interfaces in a single or multiple Arista switches.
2. Ip address needs to be in same namespace (global for example)
3. Run the sync job.

PR #965, opened 2025-09-11
[CLOUDVISION SSOT] Sync job Fails when same IP address assigned to multiple interfaces in same Namespace #964 

pato23arg:
# Closes: #964 

## What's Changed

- Fixes behavior when syncing a EVPN fabric. Duplicate IPAddresses are usual in such setup, this changes support such use case.
- Implements TAG deduplication.
- Removes log noise when duplicate IP or IP assignments are found - visible in debug mode only.
- Fixes cert failures for CVaaS implementations (verify cert = False). Such cases can happen when SSL inspection for nautobot workers is in place.
- Gets rid of device vrf tags, which lead to inconsistent syncs.

## To Do

- [ ] Unit, Integration Tests

### DIFF
--- a/changes/964.fixed
+++ b/changes/964.fixed
@@ -1,4 +1,4 @@
-Fixes sync of anycast IPs.
-Implements tag deduplication.
-Removes log noise when duplicate IP or IPAssignments are found.
-Fixes cert failures for CVaaS implementations (verify cert = False). Such cases can happen when SSL inspection for nautobot workers is in place.
+Fixes sync of anycast IPs for CloudVision SSOT.
+Implements tag deduplication for CloudVision SSOT.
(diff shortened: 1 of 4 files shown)


changes/964.fixed
```diff
@@ -0,0 +1,4 @@
+Fixes sync of anycast IPs.
```
jdrew82, line 1: Can we get these updated to specify that they're for the CVP integration specifically? Just to make the release notes clearer when they're generated.

jdrew82 (changes requested):
A small note.

changes/964.fixed
pato23arg: @jdrew82 good idea, DONE.

jdrew82 (approved):
LGTM!

Issue #1070, opened 2026-01-20
Duplicate Job Logs


jdrew82:
### Environment
* Python version:  3.10.4
* Nautobot version:  3.0.3
* nautobot-ssot version:  4.0.0

### Expected Behavior
Expect to see job logs that are not duplicated when using parallel processing.

### Observed Behavior
When the parallel processing option is selected some of the logs related to time are duplicated. This PR cleans those up.

### Steps to Reproduce
1. Setup SSoT.
2. Run Job with parallel processing enabled.
3. Review logs.

PR #1074, opened 2026-01-21
Fix Job Logger

jdrew82:
# Closes: #1070 

## What's Changed
This PR fixes the issue where the Job logger wasn't restored so had to use JobLogEntry. I figured out how to restore it so we don't need to use JobLogEntry objects directly.

## To Do

- [ ] Unit, Integration Tests


nopg (approved):
LGTM!

Issue #1028, opened 2025-11-18
Meraki Integration: Add Setting to Allow Syncing of DHCP-Based Management IPs


michalis1:
### Environment
* Nautobot version:  2.4.17
* nautobot-ssot version:  3.9.4

### Proposed Functionality

Currently, the Meraki SSoT integration does not sync DHCP-assigned management IPs, as they are considered dynamic operational data. However, some environments require management IPs to be present in Nautobot for CMDB sync purposes — and these IPs change infrequently.

Proposed Enhancement:
  * Introduce a new setting in nautobot_config.py.
  * This setting:
    * Is disabled by default to maintain current behavior.
    * When enabled, DHCP-assigned management IPs are synced into Nautobot.

### Use Case

This enhancement provides flexibility for installations that rely on Meraki devices using DHCP for management, while keeping default behavior safe for those who don’t need it. Teams using CMDB integrations (e.g. ServiceNow) or IP-based automation workflows where device management IPs are mandatory.

Issue #1027, opened 2025-11-18
Meraki Integration: Add Job Input to Assign All Meraki Devices Under a Single Location


michalis1:
### Environment
* Nautobot version:  2.4.17
* nautobot-ssot version: 3.9.4

### Proposed Functionality
The current Meraki SSoT integration creates Nautobot Location objects based on Meraki network names. However, this behavior may not be desirable in scenarios where location hierarchies are not aligned between Meraki and other systems, such as during mergers, acquisitions, or when Locations are centrally managed. In such cases, users may prefer Meraki devices to be assigned to a single, predefined location in Nautobot, without generating Locations from the Meraki API.

**Proposed Enhancement**:
  * Add a new job input that allows the user to select an existing Nautobot Location.
  * When this input is used:
    * Location sync should be skipped entirely.
    * All Meraki devices should be assigned under the selected location.

This provides full flexibility for users who prefer to manage physical/logical Locations exclusively within Nautobot, while still syncing Meraki device data.

### Use Case

Organizations that treat Meraki networks as devices only and do not want duplicate or undesired Location records created.

PR #1082, opened 2026-01-27
Meraki Integration: Sync DHCP IPs and Support Default location

michalis1:
# Closes: #1028

## To Do

Introduce `meraki_allow_dhcp_mgmt_ips"  setting in nautobot_config.py.
This setting:
 * Is disabled by default to maintain current behavior.
 * When enabled, DHCP-assigned management IPs are synced into Nautobot.

Also a new Job Input Var has been created based on #1027 to allow the use of a default location for all meraki devices.


jdrew82 (approved):
LGTM!

PR #1097, opened 2026-02-16
Fix: DataSyncBaseJob.run ignores dryrun/memory_profiling when passed as positional arguments

vmotov:
**Issue**
The `DataSyncBaseJob.run` method currently uses `kwargs.get()` to retrieve execution parameters such as `dryrun` and `memory_profiling`.

```python
def run(self, *args, **kwargs):
    self.dryrun = kwargs.get("dryrun", True)
```
However, many integration plugins (such as `nautobot-app-ssot-vsphere`) call `super().run()` and pass these arguments positionally. In Python, when a method signature is `def run(self, *args, **kwargs)`, positional arguments from a subclass call end up in `args`, not `kwargs`.

As a result, even if a user starts a job with **Dry Run: False**, the base class fails to see it in `kwargs` and falls back to the default `True`, making it impossible to actually commit changes to the database in certain scenarios.
User sees the following entries in the logs:
```
Starting job with the following options: Debug: False, Dry Run: False, Sync Tagged Only: True, Cluster Filter:
......
As dryrun is set, skipping the actual data sync.
```

**Solution**
This PR updates the `run` method signature to explicitly include `dryrun` and `memory_profiling` with their default values.

```python
def run(self, dryrun=True, memory_profiling=False, *args, **kwargs):
    self.dryrun = dryrun
    self.memory_profiling = memory_profiling
```
Python handles both positional and keyword arguments correctly with this signature, ensuring that the job configuration is respected regardless of how the subclass invokes the super method.

**Impact**

- Fixes "forced dry-run" issues in SSoT integrations.
- Improves code clarity and type hinting for the core Job entry point.
- Maintains backward compatibility for any existing `kwargs` usage.

### DIFF
--- a/nautobot_ssot/jobs/base.py
+++ b/nautobot_ssot/jobs/base.py
@@ -612,10 +612,10 @@ class DataSyncBaseJob(Job):  # pylint: disable=too-many-instance-attributes
         """Icon corresponding to the data_target."""
         return getattr(cls.Meta, "data_target_icon", None)
 
-    def run(self, *args, **kwargs):
+    def run(self, dryrun=True,  memory_profiling=False, *args, **kwargs):
         """Job entry point from Nautobot - do not override!"""
-        self.dryrun = kwargs.get("dryrun", True)
-        self.memory_profiling = kwargs.get("memory_profiling", False)
+        self.dryrun = dryrun
+        self.memory_profiling = memory_profiling
         self.parallel_loading = kwargs.get("parallel_loading", False)
         self.sync = Sync.objects.create(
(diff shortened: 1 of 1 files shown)


jdrew82:
@vmotov This was done as an expected breaking change in 4.1.0 with the addition of the `parallel_loading` variable. I had a note in the release notes of that version mentioning the change but I can add specific documentation noting the change and how to handle it. I've already confirmed with many integrations that as long as you have the expected arguments saved in your run() method they are functional. I'm going to add additional documentation around this topic for future developers.

Issue #1098, opened 2026-02-16
Device42 integration: "load_sites()" passes "LocationType.name" (string) instead of "LocationType" instance to "Location.objects.filter()"


molhamalnasr:
### Environment
* Python version: 3.12
* Nautobot version: 3.0.x
* nautobot-ssot version: 4.1.1a0

### Expected Behavior

Running the "Device42" SSoT Data Source job should successfully load site/building data from Nautobot by filtering `Location` objects by their `location_type` ForeignKey field. The job should complete without errors.

### Observed Behavior

The job fails during the `load_target_adapter()` phase with an `AttributeError`. The root cause is in `nautobot_ssot/integrations/device42/diffsync/adapters/nautobot.py`, line 149 (`load_sites` method):

```python
# Current (broken) code:
for site in Location.objects.filter(location_type=self.job.building_loctype.name):
```

`Location.location_type` is a Django `ForeignKey` to `LocationType`. When using `.filter()` on a ForeignKey field, Django expects either:
- A model instance (e.g., `self.job.building_loctype`), or
- A primary key value (UUID), or
- An explicit related field lookup (e.g., `location_type__name="Site"`)

Passing `self.job.building_loctype.name` (a string like `"Site"`) causes Django to treat it as a PK lookup, which fails because the `LocationType` PK is a UUID, not a string name. This either raises a `ValueError`/`ValidationError` or returns an empty queryset depending on the Django/database version, which cascades into downstream errors (e.g., `AttributeError: 'NoneType' object has no attribute 'name'` in `load_racks`).

**Full traceback observed:**

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.12/site-packages/celery/app/trace.py", line 479, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/usr/local/lib/python3.12/site-packages/celery/app/trace.py", line 779, in __protected_call__
    return self.run(*args, **kwargs)
  File "/usr/local/lib/python3.12/site-packages/nautobot/extras/jobs.py", line 1312, in run_job
    result = job(*args, **kwargs)
  File "/usr/local/lib/python3.12/site-packages/nautobot/extras/jobs.py", line 186, in __call__
    return self.run(*args, **deserialized_kwargs)
  File "/usr/local/lib/python3.12/site-packages/nautobot_ssot/integrations/device42/jobs.py", line 162, in run
[...]

PR #1100, opened 2026-02-16
fix(device42): Pass LocationType instance instead of name to Location filter (#1098)

molhamalnasr:
# Closes: #1098

## What's Changed

Fixed the `load_sites()` method in the Device42 Nautobot DiffSync adapter.

The `Location.objects.filter()` call was passing `self.job.building_loctype.name` (a string) to the `location_type` ForeignKey field. Django expects a model instance or a primary key (UUID) for FK lookups, not a name string. This caused the filter to fail silently or raise errors, resulting in downstream `AttributeError` exceptions during the Device42 sync job.

**Before:**
```python
for site in Location.objects.filter(location_type=self.job.building_loctype.name):
```

**After:**
```python
for site in Location.objects.filter(location_type=self.job.building_loctype):
```

## To Do

- [x] Unit, Integration Tests — N/A, one-line bugfix correcting argument type; no behavioral logic change
- [x] Documentation Updates — N/A, bugfix only, no feature changes

### DIFF
--- a/changes/1098.fixed
+++ b/changes/1098.fixed
@@ -1,0 +1,1 @@
+Fixed Device42 integration `load_sites()` passing `LocationType.name` string instead of `LocationType` instance to `Location.objects.filter()`, causing the sync job to fail.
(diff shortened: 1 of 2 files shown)


jdrew82 (approved):
This looks good to me.

molhamalnasr:
Hi @jdrew82 this merge request was merged to nautobot:ltm-2.4 and didn't reach the release. is that correct!!
do we need to create a new merge request so we get the fix in the next release, because in production it still brake the job.

Issue #1096, opened 2026-02-13
CloudVision Sync job for Single Source Of Truth App Fails with Error


pastoreerrante:
A daily scheduled Arista CloudVision Sync job for the Single Source Of Truth App is failing with Error: 

```json
{
    "exc_message": [
        {
            "name": [
                "ValidationError: ['Namespace with this Name already exists.']"
            ]
        },
        null,
        null
    ],
    "exc_module": "django.core.exceptions",
    "exc_type": "ValidationError"
}
```

Our environment details:

Nautobot version: 2.4.24

Single Source of Truth App version: 3.10.0

Job result: nautobot_ssot.integrations.aristacv.jobs.CloudVisionDataSource created at 2026-02-08 07:20:00.118659+00:00 (FAILURE)

Any idea on how to fix this? thanks

pastoreerrante:
After some more digging, it was found out the namespace who was causing the issue was "default". After deleting the "default" namespace (no objects were inside it), the job was run again. 

The default namespace was recreated but at the end the job failed again clashing with another namespace called INTERNET. 

It seems the logic to check against existing namespaces must be enhanced.

PR #1118, opened 2026-03-05
Fixed Arista CV creating duplicate namespaces

joewesch:
# Closes: #1096

## What's Changed

The root cause of the issue seems to stem from the fact that we only load namespaces (and prefixes) in Nautobot that are related to existing Arista devices. In the case of a fresh sync to a database with other device types present, we may already have a namespace with a duplicate name, but the SSoT integration doesn’t know about it. So, it tries to create it and fails due to the duplicate name.

This PR introduces two main changes to combat this:
- We now load all namespaces and prefixes from Nautobot, regardless if they were associated to Arista devices or not
  - In order to deal with pre-existing objects, we now set `DiffSyncModelFlags.SKIP_UNMATCHED_DST` on these objects so they don't try and delete them (except when configured to do so from the second change)
- There are two new configuration settings introduced that will allow users to enable the deletion of unmatched namespaces and prefixes: `aristacv_delete_namespaces_on_sync` & `aristacv_delete_prefixes_on_sync`
  - This follows the same pattern of the existing `aristacv_delete_devices_on_sync` config setting

## To Do

- [x] Unit, Integration Tests

NTC-5169

### DIFF
--- a/nautobot_ssot/integrations/aristacv/diffsync/adapters/nautobot.py
+++ b/nautobot_ssot/integrations/aristacv/diffsync/adapters/nautobot.py
@@ -112,9 +112,7 @@ class NautobotAdapter(Adapter):
                 name=ns.name,
                 uuid=ns.id,
             )
-            if self.job.app_config.delete_namespaces_on_sync:
-                new_ns.model_flags = DiffSyncModelFlags(0)
-            else:
+            if not self.job.app_config.delete_namespaces_on_sync:
                 new_ns.model_flags = DiffSyncModelFlags.SKIP_UNMATCHED_DST
             self.add(new_ns)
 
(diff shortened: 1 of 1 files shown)


nautobot_ssot/integrations/aristacv/diffsync/adapters/nautobot.py
```diff
@@ -102,6 +105,33 @@ def load_interfaces(self):
...
 
+    def load_namespaces(self):
+        """Add Nautobot Namespace objects as DiffSync Namespace models."""
+        for ns in OrmNamespace.objects.all():
+            new_ns = self.namespace(
+                name=ns.name,
+                uuid=ns.id,
+            )
+            if self.job.app_config.delete_namespaces_on_sync:
+                new_ns.model_flags = DiffSyncModelFlags(0)
```
jdrew82: I'm confused as to why you're doing this instead of just setting the flag if setting is False?

PR #1154, opened 2026-03-23
Fix unittest wrapper run with uv.

pszulczewski:
This fixes django test runner when env is controlled with `uv run nautobot-server test nautobot_ssot`.

Tested the app in uv environment for running nbshell and tests.

```
uv run nautobot-server shell
settings.PLUGINS_CONFIG.keys()
dict_keys(['nautobot_ssot', 'nautobot_ssot_custom'])

uv run nautobot-server test nautobot_ssot_custom
settings.PLUGINS_CONFIG.keys()
dict_keys([])
```
Looks like when uv runs the test wrapper on first module initialization to discover tests, config is not yet fully loaded due to Django LazyLoad. The solution is to get the PLUGIN_CONFIG key gracefully with `.get()` to avoid KeyError.

### DIFF
--- a/changes/1154.fixed
+++ b/changes/1154.fixed
@@ -1,0 +1,1 @@
+Support uv controlled development environment to run unittest.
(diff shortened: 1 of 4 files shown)


jdrew82 (approved):
LGTM!

PR #1161, opened 2026-03-30
Fixed pagination when retrieving locations from DNA Center

### LINKED ISSUE #1160: DNA Center `get_locations` doesn't retrieve all locations when there are an exact multiple of 50 (>=100) total locations
### Environment
* Python version:  3.10.19
* Nautobot version:  3.0.0
* nautobot-ssot version:  4.1.1a0

### Expected Behavior
I expect the integration to retrieve all locations from DNA Center

### Observed Behavior
When there are 100 or more locations to be retrieved, some are skipped due to where the offset lands (details will be provided in PR).

NTC-5064

joewesch:
# Closes: #1160

## What's Changed

This PR changes the `DnaCenterClient.get_locations` method to set an explicit page size (limit) instead of relying on the number of returned items.

## Detailed Analysis

In the existing implementation, we use the default page size for the API call, which is 50. We explicitly set the starting offset to 1 and then increment the offset by how many records we have fully retrieved. This works fine if there are 100 or fewer locations, but it starts to break down once you hit exactly 100 locations:

| Call | Offset | Items Returned | loc_data Length | Unique Items | Duplicate |
|------|--------|----------------|-----------------|--------------|-----------|
| 1    | 1      | 1–50           | 50              | 50           | —         |
| 2    | 50     | 50–99          | 100             | 99           | item 50   |

As you can see in the table above, we retrieved location 50 twice, and then we remove the duplicates as part of the for loop after we retrieve them all:

https://github.com/nautobot/nautobot-app-ssot/blob/0d9ded09ad82769ca980e2a08ba96c95d5e1cc7b/nautobot_ssot/integrations/dna_center/utils/dna_center.py#L51-L54

Once you have over 100 locations, the loop "self corrects" and retrieves all locations and deals with the duplicates afterward...until you hit another iteration of the default page size (150):

| Call | Offset | Items Returned | loc_data Length | Unique Items | Duplicate |
|------|--------|----------------|-----------------|--------------|-----------|
| 1    | 1      | 1–50           | 50              | 50           | —         |
| 2    | 50     | 50–99          | 100             | 99           | item 50   |
| 3    | 100    | 100–149        | 150             | 149          | —  |

## To Do

- [x] Unit, Integration Tests

NTC-5064

### DIFF
--- a/nautobot_ssot/tests/dna_center/test_utils_dna_center.py
+++ b/nautobot_ssot/tests/dna_center/test_utils_dna_center.py
@@ -33,7 +33,7 @@ class TestDnaCenterClient(TestCase):  # pylint: disable=too-many-public-methods
         self.verify = False
         self.dnac = DnaCenterClient(self.url, self.username, self.password, verify=self.verify)
         self.dnac.conn = MagicMock()
-        self.dnac.conn.sites.get_site_count.return_value = {"response": 4}
+        self.dnac.conn.sites.get_site_count.return_value = {"response": 31}
         self.dnac.conn.devices.get_device_count.return_value = {"response": 3}
 
         self.mock_response = create_autospec(Response)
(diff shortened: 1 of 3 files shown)


nautobot_ssot/tests/dna_center/test_utils_dna_center.py
```diff
@@ -70,6 +70,26 @@ def test_get_locations(self):
...
+        """Test the get_locations method correctly paginates without skipping items.
+
+        Simulates a 1-based offset API returning a fixed page size. With N total items
+        and page_size P, an off-by-one in offset causes one duplicate per page. When the
+        accumulated duplicates inflate loc_data to >= total before all unique items are
+        fetched, the loop exits early and items are silently skipped.
+        """
+        all_locations = RECV_LOCATION_FIXTURE["response"][:30]
+
+        def fake_get_site(offset=1, limit=10, **kwargs):
```
jdrew82: Is there a reason you're using a method to return a dict instead of just returning it directly?

garymccann (approved):
LGTM

PR #1193, opened 2026-04-23
fix: DNAC adapters to account for mtu=0 as an output from DNAC.

swanandtambe:
# Closes: # NTC-5652<ISSUE NUMBER GOES HERE>

## What's Changed
Changed default value of MTU to be None instead of 1500 and account for 0 as input.

## To Do

- [ ] Added change log fragment(s) (for more information see [the documentation](https://docs.nautobot.com/projects/core/en/stable/development/core/#creating-changelog-fragments))
- [ ] Unit, Integration Tests

### DIFF
--- a/changes/NTC-5652.fixed
+++ b/changes/NTC-5652.fixed
@@ -1,0 +1,1 @@
+Change MTU in DNA Center integration to account for 0 as an output from DNAC and save it in Nautobot as None.
--- a/nautobot_ssot/integrations/dna_center/diffsync/adapters/dna_center.py
+++ b/nautobot_ssot/integrations/dna_center/diffsync/adapters/dna_center.py
@@ -578,7 +578,7 @@ class DnaCenterAdapter(Adapter):
                     "port_type": port_type,
                     "port_mode": "tagged" if port["portMode"] == "trunk" else "access",
                     "mac_addr": port["macAddress"].upper() if port.get("macAddress") else None,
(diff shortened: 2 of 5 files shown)


garymccann:
@swanandtambe can you explain why this change is needed please?

It's a breaking change - folks who rely on this behaviour would have to be aware of it - so i'd recommend some documentation changes here also. 


Previous Behaviour 

If MTU set, use that, otherwise set as 1500
0 -> 0
MTU -> MTU
None -> 1500


New behaviour

If MTU set, and not 0, then use that, otherwise set to None. 

0 - > None
MTU -> MTU
None - > None

garymccann (changes requested):
Awaiting feedback on why the change is required and whether this is breaking change requiring handling such as documentation.

swanandtambe:
I believe the issue in the previous behavior is that is cannot set 0 -> 0 because the MTU field itself cannot be set to 0 https://github.com/nautobot/nautobot/blob/develop/nautobot/dcim/models/device_components.py#L552 because of the validator on that field. If the None -> None is the breaking change then I can add that back in code but setting a None input from the SoR to 1500 did not seem correct. 

The change was needed because a controller device in DNAC reports MTU=0 which is not acceptable as an input so the logic was throwing a validation error during the sync job run.

garymccann (approved):
LGTM

PR #1194, opened 2026-04-28
Ssot performance designs 2

itdependsnetworks:
This is not intended to be merged or even considered for review, but instead attempt to show the art of the possible with vibe coded commits that are layered on in a specific manner. Please see the markdown docs for deeper analysis, but the high level can be found here:

# SSoT Performance & Validation Menu

Nautobot SSoT and DiffSync provide a straightforward ETL (Extract, Transform,
Load) framework with many features out of the box. ETL is not a new idea, so
the goal isn't to invent a new technology — it's to provide structure and
best practices around one. The platform's default behavior follows a KISS
(Keep It Simple, Stupid) approach: the slow, safe, boring path that's
correct out of the box. With that approach and the breadth of features
available, there are — as always — tradeoffs worth considering.

These features aren't always easy to understand, and their performance
implications (wall-clock time, CPU, memory, I/O) aren't always obvious from
the API surface. The most common pattern for loading data into Nautobot
includes the following:

* Every object is validated for correctness
* Every object creates a changelog entry tracking the change
* Every object dispatches webhooks
* Every object fires job hooks
* Every object publishes an event
* Some objects fire additional Django signals for:
    * Cache invalidation
    * Data cleanup
    * Other custom business logic
* All data is committed together, or none of it is (atomic transactions)
* The ability to run in dry-run mode
* The ability to track changes per job run

Some — or even all — of these features may not be needed for your
implementation. For example, if you're migrating from a legacy system
where data ownership is not yet established, do you really need a changelog?
Do you want webhooks firing if you're syncing thousands of records in one
go?

Sometimes the answer is a resounding yes; other times a clean no. Sometimes
you want most of these features but are willing to trade a slightly weaker
correctness guarantee for speed. Sometimes you already know the data is
valid and don't need to validate it again. There are many reasons to choose
speed over the KISS default — and this document is intended to make those
decisions clear, and to show you how each one can be implemented.

As the below results from a simple 8k object sync show, the difference was 132x faster depending on what features you use in the one example. While results may vary based on many factors including size and shape of data, it should represent a clear picture about how the choices you make will effect the performance.

| Preset | medium total | speedup vs prod | audit chain |
|---|---:|---:|---|
| **`validated_save`** ← **PRODUCTION today** | **163.31 s** | **1.00 ×** | **FULL** — per-row OC + webhooks/jobhooks/events |
| `validated_save_no_cl` (no change context — reference only) | 40.95 s | 3.99 × | none — signal handler short-circuits |
| `save` (no clean, no change context) | 29.31 s | 5.57 × | none |
| `save_immediate_cl` (production minus clean) | 132.32 s | 1.23 × | full (no clean) |
| `save_deferred_cl` | 77.63 s | 2.10 × | full (no clean), batched OC |
| `bulk_b250` | 4.91 s | 33.28 × | none |
| `bulk_b1000` (b=1000) | 5.04 s | 32.38 × | none |
| **`bulk_b250_audit`** (`bulk_create` + `REFIRE_POST_SAVE` + deferred CL) | **4.85 s** | **33.67 ×** | **FULL** — same audit as production, no streaming |
| `stream_tier1` (`STREAMING`) | 88.28 s | 1.85 × | **FULL** — deferred CL |
| **`stream_tier1_5`** (`STREAMING \| BULK_WRITES \| VALIDATE_SOURCE_SHAPE \| VALIDATE_ON_DUMP`) | 1.83 s | **89.27 ×** | none (validation only) |
| **`stream_tier1_7`** (`STREAM_TIER1_5 \| VALIDATE_RELATIONS`) | 2.04 s | 79.93 × | none (validation only) |
| **`stream_tier2_audit`** (`STREAMING \| BULK_WRITES \| REFIRE_POST_SAVE` + `deferred_change_logging`) | **1.50 s** | **108.98 ×** | **FULL** — bulk write + REFIRE + deferred OC + webhook fan-out |
| `stream_tier2` (`STREAMING \| BULK_WRITES`, SQLite store) | 1.28 s | 127.93 × | none |
| `stream_tier2_pydict` (PyDict store — benchmark instrument) | 1.23 s | 132.80 × | none — PyDict has no `.conn` |


docs/dev/performance_validation_menu.md
```diff
@@ -0,0 +1,875 @@
...
+* Every object is validated for correctness
+* Every object creates a changelog entry tracking the change
+* Every object dispatches webhooks
+* Every object fires job hooks
+* Every object publishes an event
+* Some objects fire additional Django signals for:
+    * Cache invalidation
+    * Data cleanup
+    * Other custom business logic
+* All data is committed together, or none of it is (atomic transactions)
```
jeffkala, line 25: Is this true or is this note true? I'm not sure anymore https://docs.nautobot.com/projects/ssot/en/latest/user/performance/#escaping-the-atomic-transaction:~:text=In%20Nautobot%202.0%2C%20jobs%20will%20no%20longer%20be%20atomic%20by%20default%20so%20this%20section%20will%20not%20apply%20anymore.

docs/dev/performance_validation_menu.md
```diff
@@ -0,0 +1,875 @@
...
+
+---
+
+### 2. Change logging — `ObjectChange` rows
+
+**What it controls.** Whether each create/update/delete writes an
+`ObjectChange` row in the Nautobot database recording who changed what
+when.
+
+**Default behavior.** Every `validated_save()` (or plain `save()`) inside
```
jeffkala, line 169: Would be curious if the `ObjectChange` object creation is slow mostly becuase of the `object_data` and `object_datav2` fields on that model. 

Why I'm asking. Would be interesting to be able to turn on a "lite" ObjectChange that just keeps the many model attributes and doesn't do the `object_data` fields.

docs/dev/performance_validation_menu.md
```diff
@@ -0,0 +1,875 @@
...
+    deferred_change_logging_for_bulk_operation,
+    web_request_context,
+)
+
+with web_request_context(user, context_detail="my-job"):
+    with deferred_change_logging_for_bulk_operation():
+        # any saves / bulk_creates inside here
+        # writes are batched at end of block
+        ...
+```
```
jeffkala, line 211: Not clear to me how this code directly relates to a SSoT Datasource inherited job class. Is this just in the `run` method or implemented elsewhere.

docs/dev/performance_validation_menu.md
```diff
@@ -0,0 +1,875 @@
...
+
+**Cost & tradeoffs.** Webhooks are unconditionally per-row. SSoT does
+not currently provide a "batched webhook" primitive — that would need
+either a Nautobot-core change to webhook dispatch, or a per-integration
+deferred context that captures webhook calls and batches them
+(`deferred_webhook` is hypothetical today).
+
+**How to wire it.** Driven by axis 2 (change logging). Disable
+changelog → webhooks don't fire. Keep changelog → webhooks fire normally.
+
```
jeffkala, line 251: can just disable the webhook itself too before a ssot execution right?

docs/dev/performance_validation_menu.md
```diff
@@ -0,0 +1,875 @@
...
+doesn't lose 8,000 successful writes. Or coarser — commit-per-Job is
+already the default, but you might explicitly want a full
+`transaction.atomic()` wrapping every phase including pre-flight
+validation.
+
+**Alternatives today.** SSoT does not currently expose a knob for
+transaction scope; it inherits Nautobot's per-Job atomic behavior. The
+streaming pipeline's per-batch flushes commit incrementally — if the
+sync fails mid-stream, already-flushed batches stay committed unless
+the outer transaction rolls them back.
```
jeffkala, line 382: Feel like in a 3.x environment at least I'm not seeing transaction "default", seems more like atomic transaction isn't the default.

docs/dev/performance_validation_menu.md
itdependsnetworks: not a bad point.

docs/dev/performance_validation_menu.md
itdependsnetworks: In infoblox i thought [it was](https://github.com/nautobot/nautobot-app-ssot/blob/69b4e3407d8b9523631d91d1073503b42d772df1/nautobot_ssot/integrations/infoblox/diffsync/models/nautobot.py#L188)

jdrew82: This used to be true. As no longer have atomic transactions by default. That was a Nautobot Jobs 1.x constraint.

docs/dev/performance_validation_menu.md
jdrew82: My bigger concern is what does ObjectChange get us that the SyncLogEntry doesn't already give us?

docs/dev/performance_validation_menu.md
jdrew82: I'm fairly certain this is a hallucination. There isn't any existing code that does this. We've done it for clients with a context manager.

docs/dev/performance_validation_menu.md
jdrew82: > can just disable the webhook itself too before a ssot execution right?

Wouldn't that require knowing about the webhook ahead of time?

docs/dev/performance_validation_menu.md
jdrew82: Correct. It was default in 1.x.

docs/dev/performance_validation_menu.md
itdependsnetworks: I think @jeffkala is saying to remove the signal in a context manager (e.g. disconnect the signal and then add the signal back).

docs/dev/performance_validation_menu.md
itdependsnetworks: a little of column A a little of column B. The thing is, this was referring to just the performance tests, not to just running it via a job as that wasn't my goal.

jdrew82:
Thanks for this. Was quite the read. I'm assuming these are suggested improvements or is it suggesting these are actual existing features? If the latter, quite a bit is inaccurate like SSoTFlags and the steaming. If it's suggesting improvements it does touch on things we've already noted needs to be addressed and actually need to be done in DiffSync, not SSoT. I actually have a draft PR with some possible improvements there that are related to these.
However, the biggest thing I noticed is that none of these really address the core of the problem, which is that at a certain scale trying to load the dataset and processing it needs to be handled in a much different manner vs smaller ones. There are definitely plenty of places got improvement but at a certain point you're going to need to optimize to extreme levels for it to be reasonable to use.

docs/dev/performance_validation_menu.md
jdrew82: Ahh gotcha, yes, and potentially handled multiple triggers of said webhooks

itdependsnetworks:
> Thanks for this. Was quite the read. I'm assuming these are suggested improvements or is it suggesting these are actual existing features?

As mentioned in the first sentence "but instead attempt to show the art of the possible"

> If it's suggesting improvements it does touch on things we've already noted needs to be addressed and actually need to be done in DiffSync, not SSoT

As mentioned in the comment "These features aren't always easy to understand, and their performance implications (wall-clock time, CPU, memory, I/O) aren't always obvious from the API surface."

So put another way, "it is not clear to the average user what decisions they can make and how that effects performance." and this are of the possible style PR puts that all in one place showing you an actual example.

> However, the biggest thing I noticed is that none of these really address the core of the problem, which is that at a certain scale trying to load the dataset and processing it needs to be handled in a much different manner vs smaller ones. 

In regards to "really address the core of the problem", that is not my experience. While I know you personally are working on some large data loads, the core of the problem has been noted above. To ensure there is no ambiguity, the core problem is the lack of clarity to the design choices being made and their performance implications. There is nothing unique or clever in the vibe coded implementation, I really didn't even review it. The important part to me is having the conversation started and at least some data to show what effect each design choice has.

docs/dev/performance_validation_menu.md
progala: Some external systems rely on using ChangeLog objects for audits. Having one place for querying those means the existing processes don't need to be amended.

jdrew82:
Closing per dicussion

PR #1195, opened 2026-04-29
Fixed creating and assigning device software for the Arista SSoT inte…

### LINKED ISSUE #1172: Arista SSoT integration - missing mandatory SoftwareVersion status field after SSoT sync job
### Environment
* Python version:            3.13
* Nautobot version:  .       3.0.x
* nautobot-ssot version:   4.2.0

### Expected Behavior

the SoftwareVersion status field should get updated when running a SSoT sync job against arista devices.

the NautobotDevice.update() method in nautobot-app-ssot/nautobot_ssot/integrations/aristacv/diffsync/models/nautobot.py is missing to set the mandatory status field on SoftwareVersion. It should have a default value for setting the status.

```python
if "version" in attrs:
            dev.software_version = SoftwareVersion.objects.get_or_create(
                version=attrs["version"],
                platform=dev.platform,
            )[0]
```
```python
class SoftwareVersion(PrimaryModel):
    ...
    status = StatusField(blank=False, null=False)
```

### Observed Behavior

The status field gets not updated because of the following error:

```
{
    "exc_message": [
        "null value in column \"status_id\" of relation \"dcim_softwareversion\" violates not-null constraint\nDETAIL:  Failing row contains (3cf38585-8881-41ec-92b8-0939605b6277, 2026-04-01 13:50:34.513441+00, 2026-04-01 13:50:34.513452+00, {}, 4.34.5M, , null, null, , f, f, e3ff7462-19b3-431f-8b10-d2ddbb47042c, null).\n"
    ],
    "exc_module": "django.db.utils",
    "exc_type": "IntegrityError"
}
```

### Steps to Reproduce
1. trigger a sync job against an arista device

cfr. NTC-5548

### LINKED ISSUE #1173: Arista SSoT integration - software version gets assigned only on the second run of the SSoT sync job
### Environment
* Python version:            3.13
* Nautobot version:  .       3.0.x
* nautobot-ssot version:   4.2.0

### Expected Behavior

the SoftwareVersion should be set on the first run of the SSoT job.

The problem is that software version is checked only inside the NautobotDevice update() method but not inside create(), which means we only assign the software version on the second run of the SSoT job.

In nautobot-app-ssot/nautobot_ssot/integrations/aristacv/diffsync/models/nautobot.py NautobotDevice create() method does not check and assign attrs["version"] to any SoftwareVersion:

```python
        new_device = OrmDevice(
            status=OrmStatus.objects.get(name=attrs["status"]),
            device_type=device_type_object,
            role=role,
            platform=platform,
            location=site,
            name=ids["name"],
            serial=attrs["serial"] if attrs.get("serial") else "",
        )
```
### Observed Behavior

SoftwareVersion is being set only on the second run of the SSoT sync job

### Steps to Reproduce
1. trigger a sync job against an arista device

cfr. NTC-5548

joewesch:
…gration

# Closes: #1173

## What's Changed
During the Arista SSoT job, we were gathering the software version (`sw_ver`) from CloudVision but not assigning it to the device. However, we were assigning it during an update. This PR fixes that. In addition, we weren't setting the status on the Software object so if it encountered a new software and attempted to create one then it would fail as well.

## To Do

- [x] Unit, Integration Tests

NTC-5548

### DIFF
--- a/changes/1173.fixed
+++ b/changes/1173.fixed
@@ -1,0 +1,2 @@
+Fixed Arista SSoT integration to assign a device's software version on the first sync run.
+Fixed Arista SSoT integration to create new SoftwareVersion records with an Active status.
--- a/nautobot_ssot/integrations/aristacv/diffsync/models/nautobot.py
+++ b/nautobot_ssot/integrations/aristacv/diffsync/models/nautobot.py
@@ -7,7 +7,6 @@ from nautobot.core.settings_funcs import is_truthy
 from nautobot.dcim.models import Device as OrmDevice
 from nautobot.dcim.models import Interface as OrmInterface
 from nautobot.dcim.models import Platform as OrmPlatform
-from nautobot.dcim.models import SoftwareVersion
 from nautobot.extras.models import Status as OrmStatus
 from nautobot.ipam.models import IPAddress as OrmIPAddress
 from nautobot.ipam.models import IPAddressToInterface
@@ -76,6 +75,9 @@ class NautobotDevice(Device):
             platform = OrmPlatform.objects.get(name=ARISTA_PLATFORM)
 
         device_type_object = nautobot.verify_device_type_object(attrs["device_model"])
+        software_version_object = (
+            nautobot.verify_software_version_object(attrs["version"], platform) if attrs.get("version") else None
+        )
 
         new_device = OrmDevice(
             status=OrmStatus.objects.get(name=attrs["status"]),
@@ -85,6 +87,7 @@ class NautobotDevice(Device):
             location=site,
             name=ids["name"],
             serial=attrs["serial"] if attrs.get("serial") else "",
+            software_version=software_version_object,
         )
 
         if config.apply_import_tag:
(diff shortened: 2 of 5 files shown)


garymccann (approved):
Looks good to me, pending the below update issue, might be worth fixing?

A couple of small points from our automated review

**`update()` mishandles `version=""`** — `attrs.get("version")` is falsy for `""`, so the `elif` branch silently clears an existing `software_version`. Distinguish "absent" from "empty string" explicitly and add a test.

**`MultipleObjectsReturned` not handled** in `verify_software_version_object` — pre-existing pattern gap; document the `unique_together` assumption or guard.

**Missing `Returns:` in new docstring** — [nautobot_ssot/integrations/aristacv/utils/nautobot.py:231](#). All callers depend on the return; new public function should set the right example.

joewesch:
> **`update()` mishandles `version=""`** — `attrs.get("version")` is falsy for `""`, so the `elif` branch silently clears an existing `software_version`. Distinguish "absent" from "empty string" explicitly and add a test.

I'm not sure I understand. A value of `""` is the same as `None` here is it not? A `SoftwareVersion` must have a [non-empty, non-null value](https://github.com/nautobot/nautobot/blob/494f6e61aec124d82c7bcaf7c68070d38434a85d/nautobot/dcim/models/devices.py#L1513) for `version`.

joewesch:
> **`MultipleObjectsReturned` not handled** in `verify_software_version_object` — pre-existing pattern gap; document the `unique_together` assumption or guard.

This is also constrained by the DB's [unique_together](https://github.com/nautobot/nautobot/blob/494f6e61aec124d82c7bcaf7c68070d38434a85d/nautobot/dcim/models/devices.py#L1530-L1533) constraint to not allow multiple versions for the same platform. Adding a conditional for something that cannot exist currently is an anti-pattern (ref: [YAGNI](https://lawsofsoftwareengineering.com/laws/yagni/)).

PR #1197, opened 2026-05-05
Fixed AristaCV SSoT to properly set the management IP

### LINKED ISSUE #1174: Arista SSoT integration - devices end up with no primary IP if primary IP is not assigned to the Management interface
### Environment
* Python version:            3.13
* Nautobot version:          3.0.x
* nautobot-ssot version:   4.2.0

### Expected Behavior

Arista switches whose primary IP is assigned to an SVI and not to the Management interface should be correctly recognized by nautobot SSoT and the primary IP should be recognized as such. 

This is how SSoT currently finds the primary ip, i.e. assuming it is assigned to the mgmt interface.

```python
# in nautobot-app-ssot/nautobot_ssot/integrations/aristacv/diffsync/adapters/cloudvision.py 
# we have the load_ip_addresses()
...
attrs={"primary": bool("Management" in intf["interface"])},
...
```
Arista Cloudvision itself will calculate the management IP, which shows on the Device Overview page (not sure if that page is exposed via API).

Aside from getting that directly from CVP, at minimum if there is a single IP available it should become the primary IP.

Additional observation: Cloudvision merely chooses the first IP (on the routed interfaces part) as the primary IP when the IP is not on the management port. 

### Observed Behavior

Arista devices end up with no primary IP assigned

### Steps to Reproduce
1. setup an Arista switch/router whose primary ip address is not assigned to the mgmt interface but to an SVI
2. launch the SSoT sync job
3. the device in nautobot won't show the Primary ip

cfr. NTC-5560

joewesch:
# Closes: #1174

## What's Changed
Currently, we guess that all management (primary) IPs live on an interface that includes the word "Management" in it. In reality, the management IP may not always live on an Management interface.

This PR changes the API call that we make to get the list of devices from the gRPC request to the [Inventory REST API](https://aristanetworks.github.io/cloudvision-apis/examples/REST/inventory#get-all-devices-from-inventory). All of the same fields are present from the gRPC request, but we also get the `ipAddress` field that CloudVision uses as the real Management IP. Using this, we can properly set the `primary` flag in Nautobot.

## Screenshots

Screenshot 2026-05-05 at 4 49 34 PM

Screenshot 2026-05-05 at 3 15 26 PM

### After the fix

Screenshot 2026-05-05 at 4 48 10 PM

## To Do

- [x] Unit, Integration Tests

NTC-5560


garymccann (commented):
I'm not sure this pattern for deprecation is correct. The dosctring suggests that it will be removed in a future version, yet the code is now deleted and pointed only to the new code. 

Would it be possible to keep the existing function and point the user to the new function as default, while allowing a path to the old (with appropriate warnings) so as not to be a breaking change for anyone reliant on the existing behaviour?

In addition, automated review highlighted a few minor issues if you feel they are valid:

### MUST FIX
1. **Dropped `enumerate(..., start=1)`** — [cloudvision.py adapter:70](#). Debug log now reads "Loading 0° device". Restore `start=1`.

### SHOULD FIX
2. **`_connect_rest_client` runs unconditionally in `__init__`** — doubles the failure surface even when `create_controller=False`. Make REST connection lazy or document the new requirement.
3. **`test_load_ip_addresses` passes `primary_ip="192.0.2.1/24"`** — a CIDR will never match the bare-IP comparison in production, so the test asserts nothing about the `primary` flag. Use a bare IP, or make it an explicit no-match negative test.
4. **No adapter-level test for `create_controller=True`** — the `self.conn.get_version()` wiring path is untested at integration level. Add `test_load_devices_with_controller`.
5. **`get_version` / `get_inventory` docstrings missing `Returns:`** — D-rule lint risk for public methods on a public class.
6. **Non-standard `Deprecated:` docstring section** on `get_cvp_version` — Google style doesn't define one; put it in the summary line or use `.. deprecated::`.

### CONSIDER
7. `primary_ip = dev["ipAddress"]` — CloudVision can return `""`; current behavior is correct but implicit. Default with `dev.get("ipAddress") or ""` and a brief comment.
8. `test_load_devices` doesn't assert device count — duplicate-handling regressions would slip through.
9. Mixed `with patch()` styles in the same test class — `test_load_interfaces` still uses deep nesting while siblings were flattened.

joewesch:
> I'm not sure this pattern for deprecation is correct. The dosctring suggests that it will be removed in a future version, yet the code is now deleted and pointed only to the new code.
> 
> Would it be possible to keep the existing function and point the user to the new function as default, while allowing a path to the old (with appropriate warnings) so as not to be a breaking change for anyone reliant on the existing behaviour?

I disagree. The deprecation is merely a formality since this function was never really listed as a "public" method that users should import and use. The chances of it being used elsewhere is very small, but that's why I deprecated it instead of simply removed it.

Since the "deprecation" is the function/method **name** and not the underlying code, as long as it still returns the same result as before it's perfectly acceptable to change the old code to a simple wrapper around the new code:

```python
def old_function(x):
    warnings.warn(...)
    return new_function(x)
```

joewesch:
> 1. **Dropped `enumerate(..., start=1)`** — [cloudvision.py adapter:70](#). Debug log now reads "Loading 0° device". Restore `start=1`.

I updated the log message.

joewesch:
> 2\. **`_connect_rest_client` runs unconditionally in `__init__`** — doubles the failure surface even when `create_controller=False`. Make REST connection lazy or document the new requirement.

We are using `get_inventory` from the REST API now instead of `get_devices` that used the gRPC connection. We now need both every time. It isn't a new requirement as it uses the same URL and credentials.

joewesch:
> 3\. **`test_load_ip_addresses` passes `primary_ip="192.0.2.1/24"`** — a CIDR will never match the bare-IP comparison in production, so the test asserts nothing about the `primary` flag. Use a bare IP, or make it an explicit no-match negative test.

I stripped the CIDR from the test

joewesch:
> 4\. **No adapter-level test for `create_controller=True`** — the `self.conn.get_version()` wiring path is untested at integration level. Add `test_load_devices_with_controller`.

This is outside the scope of this PR. I didn't add this path to the code, I simply changed the function call from the now deprecated function to the new method.

joewesch:
> 5\. **`get_version` / `get_inventory` docstrings missing `Returns:`** — D-rule lint risk for public methods on a public class.

Which rule is this and why is it not enforced by ruff if true? I found [this rule](https://docs.astral.sh/ruff/rules/docstring-missing-returns/), but it says it is currently in preview only.

joewesch:
> 6\. **Non-standard `Deprecated:` docstring section** on `get_cvp_version` — Google style doesn't define one; put it in the summary line or use `.. deprecated::`.

Updated

joewesch:
@garymccann I've addressed what I think is valid.

garymccann (approved):
Looks good to me

PR #1202, opened 2026-05-10
fix(infoblox): pass *args **kwargs cleanly to super().run() to avoid TypeError

AmSach:
## Summary

Fixed TypeError in InfobloxDataSource.run() by properly passing *args **kwargs to the base class.

## Root Cause

The initial fix (PR #1) removed explicit kwargs from super().run(), which fixed the TypeError. However, the subclass's run() signature still had named parameters like dryrun, memory_profiling, and debug. When Python sees named parameters in the signature, it strips those keys from kwargs *before* resolving the call. This meant the base class DataSyncBaseJob never received dryrun, memory_profiling, etc.

## Fix

1. Both InfobloxDataSource and InfobloxDataTarget now extract kwargs values for their instance attributes using kwargs.get()
2. They then pass *args **kwargs to super().run()
3. Added getattr() guard for enable_sync_to_nautobot/infoblox checks to prevent AttributeError if config is None

This preserves the kwargs dict for the base class while allowing the subclass to use the parameter values.

## Reviewer Feedback

This addresses the concern raised by @molhamalnasr in nautobot/nautobot-app-ssot#1191.

## Testing

- Syntax check passed
- Pattern matches how DataSyncBaseJob.run() handles kwargs

### DIFF
--- a/changes/1202.fixed
+++ b/changes/1202.fixed
@@ -1,0 +1,1 @@
+Fixes TypeError when passing multiple values for keyword argument 'sync' to super().run() in InfobloxDataSource by passing *args **kwargs cleanly.
--- a/nautobot_ssot/integrations/infoblox/jobs.py
+++ b/nautobot_ssot/integrations/infoblox/jobs.py
@@ -99,7 +99,7 @@ class InfobloxDataSource(DataSource):  # pylint: disable=too-many-instance-attri
         self.debug = kwargs.get("debug")
         self.dryrun = kwargs.get("dryrun")
         self.config = kwargs.get("config")
-        if not getattr(self.config, "enable_sync_to_nautobot", True):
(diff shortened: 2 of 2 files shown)


garymccann:
Hey @AmSach, a few things to look at please:

1. Missing changelog fragment — no changes/1191.fixed (or equivalent) for this PR. Towncrier/**CI will fail**.

2. InfobloxDataTarget.run() not updated symmetrically — same # pylint: disable=arguments-differ and getattr(self.config, ...) guard should be applied; otherwise pylint flags it and a None config still crashes on the target side.

3. getattr(self.config, "enable_sync_to_nautobot", True) fails open — default of True lets a None config slip through and explode deep inside load_source_adapter(). Recommend to use **False** so the descriptive error message fires.

garymccann (approved):
Looks good, thanks for taking the time to PR this improvement

PR #1204, opened 2026-05-11
Fix AttributeError in Infoblox SSoT Prefix update

### LINKED ISSUE #1203: AttributeError: 'InfobloxDataSource' object has no attribute 'get' in NautobotNetwork.update
### Environment
* Python version:  3.12.3
* Nautobot version:  3.1.2
* nautobot-ssot version:  4.2.2

### Expected Behavior
The `Infoblox ⟹ Nautobot` synchronization job should successfully update Prefixes and their assigned VLANs without crashing.

### Observed Behavior
When running the `Infoblox ⟹ Nautobot` job, the synchronization crashes when it tries to assign new VLANs to an existing Prefix in Nautobot. The failure happens in `nautobot_ssot.integrations.infoblox.diffsync.models.nautobot.NautobotNetwork.update` on line 258.

The code attempts to use `.get("debug")` on the `job` object:
```python
if self.adapter.job.get("debug"):
```
Because `self.adapter.job` is an instance of `InfobloxDataSource` (which inherits from Nautobot's `Job` class) and not a dictionary, it does not have a `.get()` method. This results in the job failing with the following error in the Result Data:
```json
{
    "exc_message": [
        "'InfobloxDataSource' object has no attribute 'get'"
    ],
    "exc_module": "builtins",
    "exc_type": "AttributeError"
}
```

And the following traceback:
```python
  ...
  File "/usr/local/lib/python3.12/site-packages/nautobot_ssot/integrations/infoblox/diffsync/models/nautobot.py", line 258, in update
    if self.adapter.job.get("debug"):
       ^^^^^^^^^^^^^^^^^^^^
AttributeError: 'InfobloxDataSource' object has no attribute 'get'
```

### Steps to Reproduce
1. Run the "Infoblox ⟹ Nautobot" sync job.
2. Ensure there is an existing Prefix in Nautobot that requires an update to its assigned VLANs based on the data coming from Infoblox (specifically, the number of VLANs assigned to the prefix in Infoblox must be greater than the number currently assigned in Nautobot).
3. The job will crash with an `AttributeError` when evaluating line 258.

molhamalnasr:
# Closes: #1203 

## What's Changed

Fixed a bug in the Infoblox SSoT integration where updating a Prefix with new VLAN assignments caused an `AttributeError`. The code was incorrectly attempting to call `.get("debug")` on the Job object instead of accessing the `.debug` attribute directly.

## To Do

### DIFF
--- a/changes/1203.fixed
+++ b/changes/1203.fixed
@@ -1,0 +1,1 @@
+Fixed an AttributeError in Infoblox SSoT integration when updating Prefix VLAN assignments.
--- a/nautobot_ssot/integrations/infoblox/diffsync/models/nautobot.py
+++ b/nautobot_ssot/integrations/infoblox/diffsync/models/nautobot.py
@@ -206,8 +206,8 @@ class NautobotNetwork(Network):
         # Only attempt associating to VLANs if they were actually loaded
         if attrs.get("vlans") and adapter.vlan_map:
             relation = adapter.relationship_map["Prefix -> VLAN"]
+            index = 0
             for _, _vlan in attrs["vlans"].items():
-                index = 0
                 try:
                     found_vlan = adapter.vlan_map[_vlan["group"]][_vlan["vid"]]
                     if found_vlan:
@@ -269,7 +269,7 @@ class NautobotNetwork(Network):
                             self.adapter.job.logger.debug(
                                 f"Unable to find VLAN {item['vid']} {item['name']} in {item['group']} to assign to prefix {_pf.prefix}."
                             )
-                            continue
+                        continue
             else:
                 for vlan in current_vlans:
                     if vlan.vid not in attrs["vlans"]:
--- a/nautobot_ssot/integrations/infoblox/diffsync/models/nautobot.py
+++ b/nautobot_ssot/integrations/infoblox/diffsync/models/nautobot.py
@@ -255,7 +255,7 @@ class NautobotNetwork(Network):
                     try:
                         vlan = OrmVlan.objects.get(vid=item["vid"], name=item["name"], vlan_group__name=item["group"])
                         if vlan not in current_vlans:
-                            if self.adapter.job.get("debug"):
+                            if self.adapter.job.debug:
                                 self.adapter.job.logger.debug(f"Adding VLAN {vlan.vid} to {_pf.prefix}.")
                             OrmRelationshipAssociation.objects.get_or_create(
                                 relationship_id=self.adapter.relationship_map["Prefix -> VLAN"],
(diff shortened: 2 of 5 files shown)


garymccann (changes requested):
Looks good overall, a few small issues to address please

No regression test — test_network_update_network never populates attrs["vlans"] or adapter.vlan_map, so the fixed line is still uncovered. Add a TestModelNautobotNautobotPrefix case that sets both and asserts the RelationshipAssociation is created.

Pre-existing logic bug nearby — lines 268–272: continue is indented under if self.adapter.job.debug: instead of under except OrmVlan.DoesNotExist:. When debug is off, the loop silently falls through instead of continuing. Not introduced here but visible in the diff context; worth a follow-up.

Test fixtures use nb_adapter.job = Mock() — .debug auto-resolves to a truthy Mock, so the debug=False path is never tested. Use Mock(debug=True) / Mock(debug=False) explicitly.

molhamalnasr:
Hi @garymccann 

Thanks for the thorough review! I've addressed all the points and even found one more related bug while I was at it:
1. **Regression Test:** I've updated `test_network_update_network` to populate the `vlans` attribute and verify that the `RelationshipAssociation` is correctly created. I also added a new test case `test_network_update_network_no_debug` to ensure the update logic works correctly when debug logging is disabled.
2. **Logic Bug:** Fixed the indentation of the `continue` statement in `NautobotNetwork.update`. I also found and fixed a similar pre-existing bug in `NautobotNetwork.create` where the `index` was being reset inside the loop.
3. **Mock debug:** Updated the test suite to use explicit `Mock(debug=True)` or `Mock(debug=False)`. This ensures that we are actually testing both the logging and non-logging code paths.
I've pushed these updates to the branch. Let me know if there's anything else!

garymccann:
hey @molhamalnasr - can you please look to increase the test coverage? This will greatly help the project overall and ensure your code is working.

molhamalnasr:
Hi @garymccann , thanks for the feedback. I went through and significantly increased test coverage for the Infoblox Nautobot model paths in this PR.

I added focused tests in nautobot_ssot/tests/infoblox/test_nautobot_models.py to cover:

Prefix update VLAN branches (including the previously missed paths)
VLANGroup update and delete
Namespace update and delete behavior
VLAN update behavior
DNS Host and PTR update paths
process_ext_attrs branch matrix
IPAddress update branch matrix
Network create branches for ranges and VLAN mapping/error handling

With the current targeted run, coverage for nautobot_ssot/integrations/infoblox/diffsync/models/nautobot.py is now 74% (359/484 lines), and the previously flagged lines are now covered.

garymccann (approved):
Looks good. Thanks for submitting this PR, approved for merge.

PR #1213, opened 2026-05-16
Update tests to fix LocationType parent handling

justinbrink:
# Closes: #8980

## What's Changed
Updated unittest fixtures to use update_or_create to set expected `Region` to `Site` parent relationships when the default `Site` LocationType is already created with no parent and `Region` isn't defined at all. The existing tests were built on the standards for 1.x but in a fresh 3.x install a LocationType for `Region` is not created. This covers all cases of created, partially created or not created.

- [X] Added change log fragment(s)
- [X] Unit, Integration Tests

### DIFF
--- a/changes/1213.fixed
+++ b/changes/1213.fixed
@@ -1,0 +1,1 @@
+Fixed unittests for integrations that used outdated default LocationType parents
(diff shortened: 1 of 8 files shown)


garymccann (approved):
LGTM

PR #1214, opened 2026-05-16
IP Fabric integration caching and optimization

justinbrink:
## What's Changed

- DB queries per sync: ~5–10× reduction at typical environment sizes
- Per-record N+1s eliminated in the loader (devices/interfaces/VLANs)
- Inline tag/status/device lookups eliminated in the diff-application path
- 9+ real correctness bugs found and fixed along the way
- N+1 regression guards now in place so this doesn't quietly slip backward

## To Do

- [X] Added change log fragment(s)
- [X] Unit, Integration Tests

### DIFF
--- a/changes/1214.added
+++ b/changes/1214.added
@@ -1,0 +1,1 @@
+Added thread safe caching for Nautobot objects an prefetching to reduce query overhead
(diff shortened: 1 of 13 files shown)


garymccann (approved):
LGTM

jdrew82 (dismissed):
Most of this looks good but it appears we're missing some test coverage. Can we get coverage for the missing lines?

justinbrink:
Should have good coverage now and a few little issues fixed.

garymccann (approved):
Looks good to me!

PR #1224, opened 2026-05-20
Fix Infoblox SSoT sync bug - blanking out Infoblox `comment` field when any field of nautobot Prefix object gets updated

### LINKED ISSUE #1221: Infoblox SSoT - when any field on a nautobot prefix object changes, the `comment` field on the corresponding Infoblox network object becomes blank
### Environment
* Python version:   3.13
* Nautobot version:   3.11
* nautobot-ssot version:   4.2.3a0 (develop branch)

### Description

When syncing a prefix/network from nautobot to infoblox, the comments on the Infoblox `network` are repeatedly removed on one run, then added back on the next run. This happens when any prefix field is updated, but the comment (`description` on the same `prefix` object) is not.

The anomaly arises from the line `comment=attrs.get("description", "")` in `nautobot_ssot/integrations/infoblox/diffsync/models/infoblox.py` class InfobloxNetwork update() method. This line gets executed even if "description" has not been changed at all. In this case "description" won't be found in `attrs` but the comment field will be set to "" even if `description` on nautobot is not "".

In a nutshell the condition to trigger the bug are: attrs is non-empty AND "description" not in attrs.

### Expected Behavior

If only one attribute of a `prefix` model instance inside nautobot is updated, **only** the corresponding field should get updated on the Infoblox Network object. Nothing else should get changed. Hence`comment` field on Infoblox should always be in sync with the `description` field on nautobot.

### Observed Behavior

During the SSoT Sync from Nautobot to Infoblox the `comment` field gets blanked out even if it was not changed whatsoever. Any prefix field changed other than `description` can trigger the blanking of the infoblox `comment` field.

### Steps to Reproduce
1. setup a nautobot ssot integration against an Infoblox instance
2. create a`Prefix` object and set the `description` field to "SomeDescription" or anything else
3. perform an initial sync so that both sides are in sync
4. change the value of any field other than `description` on the Prefix object
5. sync again against Infoblox
6. the result on Infoblox will be a blank `comment` field on the corresponding Network object

#### Another way to reproduce the issue without a live Infoblox environment

Another way to "see" the issue without a live Infoblox environment is to create an additional test in `nautobot-app-ssot/nautobot_ssot/tests/infoblox/test_infoblox_models.py`, specifically inside the `class TestModelInfobloxNetwork(TestCase)`

```python
    def test_network_update_does_not_blank_comment_when_only_ext_attrs_change(self, mock_tag_involved_objects):
        """Regression: when only ext_attrs differ, comment must not be blanked in Infoblox.

        Reproduces the flapping-comment bug where ``attrs.get("description", "")`` substitutes
        an empty string when "description" is absent from ``attrs``, clearing the Infoblox
        comment on every sync that updates only custom fields/extensible attributes.
        """
        nb_network_atrs = {
            "description": "SameDescription",
            "ext_attrs": {"Site": "DC2"},
        }
        nb_ds_network = self.nb_adapter.prefix(**_get_network_dict(nb_network_atrs))
        self.nb_adapter.add(nb_ds_network)
        self.nb_adapter.load()
        with unittest.mock.patch(
            "nautobot_ssot.integrations.infoblox.utils.client.InfobloxApi", autospec=True
        ) as mock_client:
            infoblox_adapter = InfobloxAdapter(conn=mock_client, config=self.config)
            inf_ds_namespace = infoblox_adapter.namespace(
                name="Global",
                ext_attrs={},
            )
            infoblox_adapter.add(inf_ds_namespace)
            inf_network_atrs = {
                "description": "SameDescription",
                "ext_attrs": {"Site": "DC1"},
            }
            inf_ds_network = infoblox_adapter.prefix(**_get_network_dict(inf_network_atrs))
            infoblox_adapter.add(inf_ds_network)
            infoblox_adapter.job = Mock()
            self.nb_adapter.sync_to(infoblox_adapter)
            for call in infoblox_adapter.conn.update_network.call_args_list:
                self.assertNotEqual(
                    call.kwargs.get("comment"),
                    "",
                    f"update_network was called with comment='' which would blank the Infoblox comment: {call}",
                )
            mock_tag_involved_objects.assert_called_once()
```

pastoreerrante:
# Closes: #1221

## What's Changed

This should fix an annoying bug when nautobot `prefix` objects get ssot synced against Infoblox.

Any prefix field who get changed and later synced against Infoblox renders the `comment` field blank (with the exception of the Prefix `description` field).

The fix is very easy. We just check if the key `description` is inside the `attrs` dict which is passed to InfobloxNetwork.update() method. If so, we do the update, else we don't. 

## To Do

- [x] Unit, Integration Tests

### DIFF
--- a/changes/1211.fixed
+++ b/changes/1211.fixed
@@ -1,0 +1,1 @@
+Fixed Infoblox SSoT sync bug - blanking out Infoblox `comment` field when any field of nautobot Prefix object gets updated
--- a/nautobot_ssot/integrations/infoblox/diffsync/models/infoblox.py
+++ b/nautobot_ssot/integrations/infoblox/diffsync/models/infoblox.py
@@ -68,19 +68,20 @@ class InfobloxNetwork(Network):
             network_view_to_namespace_map=self.adapter.config.infoblox_network_view_to_namespace_map,
             direction="ns_to_nv",
         )
-        network_type = attrs.get("network_type", self.network_type)
-        if network_type != "container":
-            self.adapter.conn.update_network(
-                prefix=self.get_identifiers()["network"],
-                network_view=network_view,
-                comment=attrs.get("description", ""),
-            )
-        else:
-            self.adapter.conn.update_network_container(
-                prefix=self.get_identifiers()["network"],
-                network_view=network_view,
-                comment=attrs.get("description", ""),
-            )
+        if "description" in attrs:
+            network_type = attrs.get("network_type", self.network_type)
+            if network_type != "container":
+                self.adapter.conn.update_network(
+                    prefix=self.get_identifiers()["network"],
+                    network_view=network_view,
+                    comment=attrs.get("description", ""),
+                )
+            else:
+                self.adapter.conn.update_network_container(
+                    prefix=self.get_identifiers()["network"],
+                    network_view=network_view,
+                    comment=attrs.get("description", ""),
+                )
         if attrs.get("ranges"):
             self.adapter.job.logger.warning(
                 f"Prefix, {self.network}-{self.namespace}, has a change of Ranges in Nautobot, but"
--- a/nautobot_ssot/tests/infoblox/test_infoblox_models.py
+++ b/nautobot_ssot/tests/infoblox/test_infoblox_models.py
@@ -196,6 +196,49 @@ class TestModelInfobloxNetwork(TestCase):
             )
             mock_tag_involved_objects.assert_called_once()
 
+    @unittest.mock.patch(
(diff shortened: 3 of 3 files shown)


jdrew82:
@pastoreerrante Can we get test coverage for all of the new lines you're adding? Thanks!

pastoreerrante:
@jdrew82, actually I just added one line: `if "description" in attrs:` in [2dbd6fb](https://github.com/nautobot/nautobot-app-ssot/commit/2dbd6fb82ebf230b22489d6cc4a5e5191a0fcb1d). Not sure why it looks like I added many more. But it's just one line and I believe it is already covered by the unit test `test_network_update_does_not_blank_comment_when_only_ext_attrs_change`.

jdrew82 (approved):

PR #1225, opened 2026-05-22
Add skip_component_autocreation opt-in (Phases 1-3 of SSoT feature)

### LINKED ISSUE #1227: Phase 1 — Core mechanism for skip_component_autocreation
Sub-issue of #1226.

Phase 1 of the `skip_component_autocreation` feature. Delivers the underlying mechanism so later phases can opt in.

### Scope

- `nautobot_ssot/contrib/component_autocreation.py` — `skip_component_autocreation` context manager, `is_suppression_active()` helper, idempotent `install_patches()` / `uninstall_patches()` that wrap `Device.create_components` and `Module.create_components` with a `contextvars`-driven shim.
- `nautobot_ssot/__init__.py` — register `skip_component_autocreation` in `default_settings`; call `install_patches()` from `NautobotSSOTAppConfig.ready()` inside a broad try/except.
- `nautobot_ssot/contrib/__init__.py` — re-export the new public API.

### Acceptance criteria

- [x] Module created and public API re-exported.
- [x] Patch is inert by default (behaviour unchanged unless a caller opts in).
- [x] `ContextVar` (not `threading.local`) used.
- [x] `install_patches()` is idempotent and tolerates Nautobot import failure.
- [x] Wrapper preserves `alters_data = True` and exposes `_ssot_wrapped` / `_ssot_original` sentinels.

Implemented in #1225.

jdrew82:
# Closes: #1227, #1228, #1229 (parent: #1226)

## What's Changed

Implements **Phases 1–3** of the `skip_component_autocreation` feature (see #1226 for the full feature request and phase breakdown). SSoT jobs can now opt in to suppressing Nautobot's automatic `Device.create_components()` / `Module.create_components()` calls during a sync.

When a `Device` or `Module` is created in Nautobot, `save()` unconditionally instantiates one component per template defined on the related `DeviceType`/`ModuleType`. For SSoT this is almost always undesirable — the source of truth owns the component inventory, and the auto-created components either collide with or pollute what the sync is about to write.

### Phase 1 — Core mechanism (#1227)

- **`nautobot_ssot/contrib/component_autocreation.py`** *(new)* — exposes:
  - `skip_component_autocreation` — class-based context manager (reentrant, exception-safe, `contextvars`-backed).
  - `is_suppression_active()` — helper for tests and adapter code.
  - `install_patches()` / `uninstall_patches()` — idempotent monkey-patch installers that wrap `Device.create_components` and `Module.create_components` with a delegating shim. Preserves `alters_data = True`, stores `_ssot_wrapped` / `_ssot_original` sentinels for introspection and rollback.
- **`nautobot_ssot/__init__.py`** — adds `skip_component_autocreation: False` to `default_settings` and calls `install_patches()` from `AppConfig.ready()` inside a broad try/except so a patch failure never blocks app startup.
- **`nautobot_ssot/contrib/__init__.py`** — re-exports the new public API; lazy-loads the ORM-dependent `NautobotAdapter` / `NautobotModel` via module `__getattr__` so importing the contrib namespace during Django startup does not eagerly import ORM models.

### Phase 2 — Job-level integration (#1228)

- **`nautobot_ssot/jobs/base.py`**:
  - Adds `skip_component_autocreation: bool = False` class attribute on `DataSyncBaseJob` (documented in the class docstring).
  - Adds the `_maybe_suppress_component_autocreation` decorator and applies it to `sync_data()`. The decorator resolves the effective flag as the class attribute **OR** `PLUGINS_CONFIG["nautobot_ssot"]["skip_component_autocreation"]` (either source set to `True` opts in) and, when set, runs the entire `sync_data()` body inside the `skip_component_autocreation()` context manager. Decorating rather than re-indenting keeps the `sync_data()` body unchanged for review.
- **`nautobot_ssot/jobs/examples.py`** — silences pylint `no-name-in-module` for the now lazily-exported contrib names.

### Why a monkey-patch?

`create_components()` is called unconditionally inside `Device.save()` / `Module.save()` in Nautobot core, with no signal, setting, or hook gating it. `bulk_create()` is used internally so `post_save` listeners on components can't intercept it either. The monkey-patch is the only reachable extension point today. Phase 6 (#1232) tracks proposing a proper upstream hook in `nautobot/nautobot`, after which this app will feature-detect and deprecate the patch path.

### Opt-in API (three levels)

```python
# 1. Context manager — arbitrary code

    Device.objects.create(...)   # no auto-instantiated components

# 2. Job class attribute — recommended

# 3. Global setting — operator-wide
# PLUGINS_CONFIG["nautobot_ssot"]["skip_component_autocreation"] = True
```

### Verified locally

- `ruff check` / `ruff format --check` clean on all changed files.
- AST parse of all changed Python files.
- Standalone smoke test of the Phase 1 module via `importlib`: context-manager default off, nested correctly, exception-safe, wrapper delegates when off and returns `[]` when on, `alters_data`/`_ssot_wrapped`/`_ssot_original` set on wrapper, `install_patches()` logs and returns gracefully when `nautobot.dcim` is unimportable.
- Standalone test of the Phase 2 decorator's OR-semantics truth table (4/4 cases), suppression released after `sync_data()` returns, and INFO-log gating matched to suppression.
- Full suite run against the SSoT dev stack: 19 new tests pass; existing `test_jobs.py` (75 tests) unaffected.

## To Do

- [x] Added change log fragment(s) (`changes/1226.added`)
- [ ] Attached Screenshots, Payload Example — N/A (no UI/API surface)
- [x] Unit, Integration Tests — **Phase 3** (#1229): `nautobot_ssot/tests/test_component_autocreation.py` (19 tests) covers the mechanism, end-to-end Device/Module suppression, and the job-level integration. Existing `test_jobs.py` (75 tests) still passes.
- [ ] Documentation Updates — **Phase 4** (#1230, separate PR).
- [ ] Phase 5 — release prep (#1231).
- [ ] Phase 6 — upstream Nautobot core hook (#1232).

## Reviewer Notes

- Default behaviour is unchanged: with neither the class attribute nor the setting enabled, `sync_data()` runs exactly as before, and the patched `create_components` delegates to the original.
- `_maybe_suppress_component_autocreation` is applied to the base `sync_data()`. Subclasses that fully override `sync_data()` without calling `super()` should use the `skip_component_autocreation()` context manager directly — documented in Phase 4.
- `install_patches()` is idempotent and tolerant of import failures, so it is safe under Django autoreload and minimal/headless test environments.
- `ContextVar` (not `threading.local`) is deliberate — it isolates correctly under asyncio tasks and Celery workers, including under `CELERY_TASK_ALWAYS_EAGER=True`.

## Upstream coordination

An upstream feature request for a public extension point has been filed in Nautobot core: nautobot/nautobot#9026. The core team has asked for the core integration to land first, with SSoT then implementing consumer-side support that feature-detects the upstream symbol. The disposition of this PR (hold as draft, close in favour of a fresh consumer-side PR after #9026 lands, or merge as-is with a deprecation path) is to be confirmed by the maintainers.

### DIFF
--- a/changes/1226.added
+++ b/changes/1226.added
@@ -1,0 +1,1 @@
+Added `skip_component_autocreation` opt-in for SSoT jobs, suppressing Nautobot's automatic Device/Module component instantiation during sync.
--- a/changes/1229.housekeeping
+++ b/changes/1229.housekeeping
@@ -1,0 +1,1 @@
+Added unit tests for the `skip_component_autocreation` mechanism, end-to-end Device/Module suppression, and the `DataSyncBaseJob` job-level integration.
--- a/nautobot_ssot/__init__.py
+++ b/nautobot_ssot/__init__.py
@@ -8,6 +8,7 @@ from django.conf import settings
 from nautobot.core.settings_funcs import is_truthy
 from nautobot.extras.plugins import NautobotAppConfig
 
+from nautobot_ssot.contrib.component_autocreation import install_patches
 from nautobot_ssot.integrations.utils import each_enabled_integration_module
 
 logger = logging.getLogger("nautobot.ssot")
@@ -150,8 +151,6 @@ class NautobotSSOTAppConfig(NautobotAppConfig):
         # the PLUGINS_CONFIG setting of the same name). See
         # nautobot_ssot/contrib/component_autocreation.py for details.
         try:
-            from nautobot_ssot.contrib.component_autocreation import install_patches
-
             install_patches()
         except Exception:  # pylint: disable=broad-except
             # Failure to install the suppression patches must never block app startup.
--- a/nautobot_ssot/__init__.py
+++ b/nautobot_ssot/__init__.py
@@ -122,6 +122,7 @@ class NautobotSSOTAppConfig(NautobotAppConfig):
         "servicenow_password": "",
         "servicenow_username": "",
         "enable_global_search": True,
+        "skip_component_autocreation": False,
     }
     config_view_name = "plugins:nautobot_ssot:config"
     docs_view_name = "plugins:nautobot_ssot:docs"
@@ -142,6 +143,19 @@ class NautobotSSOTAppConfig(NautobotAppConfig):
                 "Sync",
                 "SyncLogEntry",
             ]
+
+        # Install the opt-in Device/Module create_components() suppression patches.
+        # The patches are inert by default and only take effect inside a
+        # skip_component_autocreation() context (driven by the job class attribute or
+        # the PLUGINS_CONFIG setting of the same name). See
+        # nautobot_ssot/contrib/component_autocreation.py for details.
+        try:
+            from nautobot_ssot.contrib.component_autocreation import install_patches
+
+            install_patches()
+        except Exception:  # pylint: disable=broad-except
+            # Failure to install the suppression patches must never block app startup.
+            logger.exception("Failed to install Device/Module component_autocreation patches.")
 
 
 config = NautobotSSOTAppConfig  # pylint:disable=invalid-name
--- a/nautobot_ssot/contrib/__init__.py
+++ b/nautobot_ssot/contrib/__init__.py
@@ -1,6 +1,10 @@
 """SSoT Contrib."""
 
 from nautobot_ssot.contrib.adapter import NautobotAdapter
+from nautobot_ssot.contrib.component_autocreation import (
+    is_suppression_active,
+    skip_component_autocreation,
+)
 from nautobot_ssot.contrib.model import NautobotModel
 from nautobot_ssot.contrib.types import (
     CustomFieldAnnotation,
@@ -14,4 +18,6 @@ __all__ = (
     "NautobotAdapter",
     "NautobotModel",
     "RelationshipSideEnum",
+    "is_suppression_active",
+    "skip_component_autocreation",
 )
--- a/nautobot_ssot/contrib/__init__.py
+++ b/nautobot_ssot/contrib/__init__.py
@@ -4,8 +4,18 @@ Keep imports in this module safe during Django startup by avoiding eager imports
 of modules that touch ORM models at import time.
 """
 
+from typing import TYPE_CHECKING
+
 from nautobot_ssot.contrib.component_autocreation import is_suppression_active, skip_component_autocreation
 from nautobot_ssot.contrib.types import CustomFieldAnnotation, CustomRelationshipAnnotation, RelationshipSideEnum
+
+if TYPE_CHECKING:
+    # Imported for static analysis only (type checkers, and mkdocstrings/griffe doc
+    # collection). These are NOT imported at runtime -- TYPE_CHECKING is False then --
+    # so Django app startup still avoids eagerly importing ORM-backed modules. At
+    # runtime the names below resolve lazily via __getattr__.
+    from nautobot_ssot.contrib.adapter import NautobotAdapter
+    from nautobot_ssot.contrib.model import NautobotModel
 
 
 def __getattr__(name):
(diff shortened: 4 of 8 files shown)


itdependsnetworks:
> For SSoT this is almost always undesirable — the source of truth owns the component inventory, and the auto-created components either collide with or pollute what the sync is about to write.

Can you expand on this? I think you are saying you want to re-build the logic that the device type creation process does, but that seems counter intuitive. I can see the challenge that you load the state of the data from nautobot, then change the state in `create_components` process, and then have issues, but I want to make sure we understand and agree this is the best path forward. Pinging @glennmatthews for visibility.

jdrew82:
> > For SSoT this is almost always undesirable — the source of truth owns the component inventory, and the auto-created components either collide with or pollute what the sync is about to write.
> 
> 
> 
> Can you expand on this? I think you are saying you want to re-build the logic that the device type creation process does, but that seems counter intuitive. I can see the challenge that you load the state of the data from nautobot, then change the state in `create_components` process, and then have issues, but I want to make sure we understand and agree this is the best path forward. Pinging @glennmatthews for visibility. 

Actually it's more about disabling the create_components() call in Device and Module.save() call. Ultimately the best option is a mechanism in Nautobot to disable it but until that's done this is a stop gap.

glennmatthews:
> > > For SSoT this is almost always undesirable — the source of truth owns the component inventory, and the auto-created components either collide with or pollute what the sync is about to write.
> > 
> > 
> > Can you expand on this? I think you are saying you want to re-build the logic that the device type creation process does, but that seems counter intuitive. I can see the challenge that you load the state of the data from nautobot, then change the state in `create_components` process, and then have issues, but I want to make sure we understand and agree this is the best path forward. Pinging @glennmatthews for visibility.
> 
> Actually it's more about disabling the create_components() call in Device and Module.save() call. Ultimately the best option is a mechanism in Nautobot to disable it but until that's done this is a stop gap.

Yeah, I'd rather that we *start* with the Nautobot feature (opting out of `create_components()` when desired for patterns such as SSOT). It feels backwards to me to create the tooling that will use this API before we have an agreement as to what said API will look like?

jdrew82:
> > > > For SSoT this is almost always undesirable — the source of truth owns the component inventory, and the auto-created components either collide with or pollute what the sync is about to write.
> > > 
> > > 
> > > Can you expand on this? I think you are saying you want to re-build the logic that the device type creation process does, but that seems counter intuitive. I can see the challenge that you load the state of the data from nautobot, then change the state in `create_components` process, and then have issues, but I want to make sure we understand and agree this is the best path forward. Pinging @glennmatthews for visibility.
> > 
> > 
> > Actually it's more about disabling the create_components() call in Device and Module.save() call. Ultimately the best option is a mechanism in Nautobot to disable it but until that's done this is a stop gap.
> 
> Yeah, I'd rather that we _start_ with the Nautobot feature (opting out of `create_components()` when desired for patterns such as SSOT). It feels backwards to me to create the tooling that will use this API before we have an agreement as to what said API will look like?

I'm totally fine starting with the Nautobot feature first if desired. It was designed this way as it could be implemented for the client through SSoT updates quicker and easier than core updates. The SSoT changes were only ever a patch til the core feature was implemented.

jdrew82:
Closing in favour of a core-first approach. The public extension point this PR worked around is now tracked upstream at nautobot/nautobot#9026. Once that lands, a fresh SSoT PR will be filed scoped to just the consumer-side wiring (DataSyncBaseJob class attribute, PLUGINS_CONFIG flag, tests, and docs) against the upstream symbol. The branch `u/jdrew-skip-component-autocreation` and the commits here remain available as reference. Sub-issues #1227, #1228, and #1229 closed as superseded. Phases 4 (#1230), 5 (#1231), and 6 (#1232) remain open against parent #1226.

PR #1238, opened 2026-05-28
Resolve KeyError with LibreNMS IP's

nopg:
LibreNMS would crash if searching for the wrong type of IP address, this resolves that crash. Also removed a stray print/debug statement left in code previously.

### DIFF
--- a/nautobot_ssot/integrations/librenms/utils/librenms.py
+++ b/nautobot_ssot/integrations/librenms/utils/librenms.py
@@ -172,7 +172,7 @@ class LibreNMSApi(ApiEndpoint):  # pylint: disable=too-few-public-methods
             ipaddress_ip_interface = ipaddress.ip_address(ip_address)
             if isinstance(ipaddress_ip_interface, ipaddress.IPv4Address):
                 for ipv4_address in device_ips["addresses"]:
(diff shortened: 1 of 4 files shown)


nautobot_ssot/integrations/librenms/utils/librenms.py
```diff
@@ -172,7 +172,7 @@ def get_librenms_ipinfo_for_device_ip(self, librenms_device_id: int, ip_address:
             ipaddress_ip_interface = ipaddress.ip_address(ip_address)
             if isinstance(ipaddress_ip_interface, ipaddress.IPv4Address):
                 for ipv4_address in device_ips["addresses"]:
-                    if ipv4_address["ipv4_address"] == ip_address:
+                    if ipv4_address.get("ipv4_address") == ip_address:
```
jdrew82, line 175: Can we get some test coverage for these new pieces of code?

jdrew82 (changes requested):

nautobot_ssot/integrations/librenms/utils/librenms.py
nopg: Done.

jdrew82 (approved):

PR #1244, opened 2026-06-02
Add skip_auto_component_creation opt-in for SSoT jobs (#1226)

### LINKED ISSUE #1226: Add `skip_component_autocreation` opt-in to suppress automatic Device/Module component instantiation during SSoT syncs
### Environment
* Nautobot version: `>=3.0.0,<4.0.0` (matches `nautobot-ssot` `develop` constraint)
* nautobot-ssot version: `4.2.2a0` (current `develop`)

### Proposed Functionality

Add an opt-in mechanism that suppresses Nautobot's automatic `Device.create_components()` and `Module.create_components()` calls during SSoT sync jobs. The feature is exposed at three levels of granularity, all driven by the same underlying mechanism:

1. **Context manager** — `from nautobot_ssot.contrib import skip_component_autocreation`. Any `Device` or `Module` saved inside the `with` block does not trigger automatic component instantiation.

   ```python

       Device.objects.create(...)  # no auto-instantiated components
   ```

2. **Job class attribute** — `skip_component_autocreation: bool = True` on a `DataSource` / `DataTarget` subclass causes the entire `sync_data()` lifecycle to run inside the context manager. This is the recommended opt-in path for most users.

   ```python

       # ... rest of the job
   ```

3. **Global setting** — `PLUGINS_CONFIG[\"nautobot_ssot\"][\"skip_component_autocreation\"] = True` opts every SSoT job in. Intended for operators who want this behaviour app-wide. Defaults to `False`.

**Implementation:** A process-wide patch is installed once from `NautobotSSOTAppConfig.ready()`, wrapping `Device.create_components` and `Module.create_components` with a delegating shim that consults a `contextvars.ContextVar`. When the flag is unset (the default), the wrappers delegate to the original methods and behaviour is identical to today; when the flag is set, the wrappers return `[]` without instantiating anything.

`contextvars` (not `threading.local`) is used so the flag isolates correctly under asyncio tasks and Celery workers. The patch is idempotent (Django autoreload-safe), preserves Django's `alters_data = True` template-safety attribute, and stores `_ssot_wrapped` / `_ssot_original` sentinels on the wrapper for introspection and rollback.

The mechanism is currently a monkey-patch because Nautobot core calls `create_components()` unconditionally from inside `Device.save()` / `Module.save()` with no signal, setting, or hook gating it (and the internal `bulk_create()` means `post_save` listeners on the components themselves can't intercept it either). A follow-up issue will be filed against `nautobot/nautobot` proposing a public extension point (`nautobot.apps.dcim.skip_auto_component_creation()` context manager or a `pre_create_components` signal). Once that lands, this app will feature-detect and deprecate the patch path.

**Deliverables:**
- `nautobot_ssot/contrib/component_autocreation.py` — context manager, `is_suppression_active()` helper, idempotent `install_patches()` / `uninstall_patches()`.
- Re-exports in `nautobot_ssot/contrib/__init__.py`.
- `skip_component_autocreation` registered in `default_settings`; `install_patches()` called from `AppConfig.ready()` inside a broad try/except so app startup is never blocked by patch failures.
- `DataSyncBaseJob.sync_data()` resolves the effective flag (`self.skip_component_autocreation or settings.PLUGINS_CONFIG[...].get(...)`) and wraps the sync body in the context manager when set.
- Unit tests covering: default off, context-manager nesting, exception safety, wrapper attributes (`alters_data`, `_ssot_wrapped`, `_ssot_original`), Device/Module end-to-end suppression, class-attribute opt-in, settings-flag opt-in, settings vs. attribute OR semantics, isolation under `ContextVar`, propagation under `CELERY_TASK_ALWAYS_EAGER=True`.
- Documentation page under `docs/admin/install/` (or equivalent) explaining the three opt-in paths and calling out the ModuleBay implication (see Use Case caveat below).
- Changelog fragment: `changes/<this-issue>.added` — "Added `skip_component_autocreation` opt-in for SSoT jobs, suppressing Nautobot's automatic Device/Module component instantiation during sync."

### Use Case

I am the author of a `DataTarget` SSoT job that syncs devices from an external source-of-truth into Nautobot. The external source-of-truth defines its own canonical interface inventory per device (`Ethernet1/1`, `Ethernet1/2`, …, `mgmt0`, etc.). Today, when my job creates a Nautobot `Device`, Nautobot's `Device.save()` calls `create_components()`, which instantiates one component per template on the assigned `DeviceType` — typically dozens of `Interface`, `ConsolePort`, `PowerPort`, etc. objects with template-defined names like `GigabitEthernet0/1`.

My adapter then proceeds to create the source-of-truth's actual interface set. One of two things happens:

1. **Names overlap**: the source-of-truth has `Ethernet1/1`, but the template-created interface is also called `Ethernet1/1`. The diff resolves to a no-op "update," but every subsequent sync repeats the wasted `bulk_create` + per-template lookup.
2. **Names don't overlap**: I end up with both — Nautobot now holds the union of source-of-truth components *and* template-derived components. The latter are pure pollution that I must clean up via a second pass in the adapter (an interface-delete sweep keyed on "interfaces not in source-of-truth"), which is brittle and easy to get wrong (e.g., wiping interfaces that the user added manually).

After this feature lands, my job sets `skip_component_autocreation = True` (or wraps adapter code in the context manager) and the problem goes away — Nautobot creates the Device, no components are instantiated, and my adapter is the sole authority for component inventory. The cleanup pass disappears and sync output is deterministic.

This pattern applies to essentially every SSoT integration that touches `Device` or `Module` and sources its component inventory from outside Nautobot — Meraki, DNA Center, IP Fabric, Infoblox (where it touches Device), Device42, ACI, vSphere, etc.

**Caveat to call out in the docs:** suppressing autocreation also skips `ModuleBay` template instantiation. If a sync subsequently creates `Module` instances that expect specific `ModuleBay` slots, the adapter is responsible for syncing the bays too. Adapter authors that opt in own the full component graph.

### Related

- Upstream feature request in Nautobot core: nautobot/nautobot#9026 — public extension point proposed; core-first sequencing means SSoT consumer-side work (Phases 2-5) will switch to feature-detect that upstream symbol when it ships.
- Initial in-app implementation (closed in favour of core-first approach): #1225

jdrew82:
# Closes: #1226

## What's Changed

Implements the SSoT consumer-side wiring for the public extension point added in Nautobot core via nautobot/nautobot#9026 (PR nautobot/nautobot#9027). SSoT jobs can now opt in to suppressing Nautobot's automatic `Device` / `Module` component instantiation during a sync via three layers:

1. **Direct context manager** (re-exported from core):

    ```python
    from nautobot_ssot.contrib import SkipAutoComponentCreation

    with SkipAutoComponentCreation():
        device = Device.objects.create(...)   # no template-derived components
    ```

2. **`DataSyncBaseJob` class attribute** — recommended for SSoT jobs:

    ```python

        skip_auto_component_creation = True
    ```

3. **`PLUGINS_CONFIG` global setting** — operator-wide opt-in:

    ```python
    PLUGINS_CONFIG = {
        "nautobot_ssot": {
            "skip_auto_component_creation": True,
        },
    }
    ```

Either source set to `True` opts in (OR semantics). Default behaviour is unchanged for any job that does not opt in.

## How it works

- `nautobot_ssot/contrib/component_creation.py` *(new)* — feature-detects `nautobot.apps.dcim.SkipAutoComponentCreation`. On a Nautobot version that ships the upstream symbol, the re-export is transparent. On older versions the wrapper provides a no-op fallback that emits a one-shot logger warning the first time it is entered, so opt-in users see why suppression is not taking effect rather than the feature silently doing nothing.
- `nautobot_ssot/contrib/__init__.py` — re-exports `SkipAutoComponentCreation` and `is_auto_component_creation_suppressed` so apps can import from `nautobot_ssot.contrib`.
- `nautobot_ssot/__init__.py` — adds `skip_auto_component_creation: False` to `default_settings` so the `PLUGINS_CONFIG` path is well-defined out of the box.
- `nautobot_ssot/jobs/base.py`:
  - Adds `skip_auto_component_creation: bool = False` class attribute on `DataSyncBaseJob`, documented in the class docstring.
  - Adds `_maybe_suppress_auto_component_creation` decorator and applies it to `sync_data()`. The decorator resolves the effective flag (class attribute **OR** setting) and, when set, wraps the entire `sync_data()` body in a `SkipAutoComponentCreation()` context. Decorating rather than re-indenting keeps the `sync_data()` body unchanged for review.

## Commit layout

Split into four focused commits for ease of review:

1. Add feature-detect wrapper for `nautobot.apps.dcim.SkipAutoComponentCreation`
2. Wire `skip_auto_component_creation` into `DataSyncBaseJob`
3. Add tests
4. Add changelog fragment

## To Do

- [x] Added change log fragment(s) (`changes/1226.added`)
- [x] Attached Screenshots, Payload Example — N/A (no UI surface)
- [x] Unit, Integration Tests — `nautobot_ssot/tests/test_component_creation.py` (11 tests). Against a Nautobot that does not yet ship the upstream symbol, 4 run + 7 skip. All 11 run once nautobot/nautobot#9027 merges.
- [ ] Documentation Updates — pending; can land in a follow-up PR alongside the user-facing rollout
- [ ] Outline Remaining Work, Constraints from Design — see *Forward-compatibility* below

## Forward-compatibility / upstream dependency

This PR is **safe to merge before** nautobot/nautobot#9027 lands. On a Nautobot installation without the upstream symbol the opt-in is a no-op (template components are still instantiated) and a one-shot warning is logged the first time the suppression path is exercised. Once #9027 ships in a Nautobot release and the SSoT `pyproject.toml` floor is raised to that release, the warning path becomes unreachable and the feature is fully active without any further SSoT code change.

## Verification

- `ruff check` + `ruff format --check` clean on all changed files.
- `invoke unittest --label nautobot_ssot.tests.test_component_creation`: **11/11 pass** (4 run + 7 skipped against a pre-#9027 Nautobot).
- `invoke unittest --label nautobot_ssot.tests.test_jobs`: **78/78 pass** (no regression from the `sync_data()` decorator).
- Author has confirmed end-to-end functionality against a Nautobot build that includes #9027.

## Reviewer notes

- The PR previously associated with parent #1226 (the in-app monkey-patch implementation) was closed as superseded once the core team requested a core-first approach. This branch is the consumer-side rework that depends on the upstream extension point instead of monkey-patching.
- `SkipAutoComponentCreation` here is a class re-export from `nautobot.apps.dcim` when available, and a small class fallback otherwise. Either way, it is used as `with SkipAutoComponentCreation():`.
- Suppression is scoped per-task via the upstream's `contextvars.ContextVar`, so the flag does not leak across Celery worker invocations.

### DIFF
--- a/changes/1226.added
+++ b/changes/1226.added
@@ -1,0 +1,1 @@
+Added `skip_auto_component_creation` opt-in for SSoT jobs, suppressing Nautobot's automatic Device/Module component instantiation during sync via Nautobot core's `nautobot.apps.dcim.SkipAutoComponentCreation` extension point (nautobot/nautobot#9026).
--- a/nautobot_ssot/__init__.py
+++ b/nautobot_ssot/__init__.py
@@ -122,6 +122,7 @@ class NautobotSSOTAppConfig(NautobotAppConfig):
         "servicenow_password": "",
         "servicenow_username": "",
         "enable_global_search": True,
+        "skip_auto_component_creation": False,
     }
     config_view_name = "plugins:nautobot_ssot:config"
     docs_view_name = "plugins:nautobot_ssot:docs"
--- a/nautobot_ssot/contrib/__init__.py
+++ b/nautobot_ssot/contrib/__init__.py
@@ -1,6 +1,10 @@
 """SSoT Contrib."""
 
 from nautobot_ssot.contrib.adapter import NautobotAdapter
+from nautobot_ssot.contrib.component_creation import (
+    SkipAutoComponentCreation,
+    is_auto_component_creation_suppressed,
+)
 from nautobot_ssot.contrib.model import NautobotModel
 from nautobot_ssot.contrib.types import (
     CustomFieldAnnotation,
@@ -14,4 +18,6 @@ __all__ = (
     "NautobotAdapter",
     "NautobotModel",
     "RelationshipSideEnum",
+    "SkipAutoComponentCreation",
+    "is_auto_component_creation_suppressed",
 )
--- a/nautobot_ssot/contrib/component_creation.py
+++ b/nautobot_ssot/contrib/component_creation.py
@@ -24,7 +24,10 @@ logger = logging.getLogger(__name__)
 
 
 try:
-    from nautobot.apps.dcim import SkipAutoComponentCreation, is_auto_component_creation_suppressed  # pylint: disable=unused-import
+    from nautobot.apps.dcim import (  # pylint: disable=unused-import
+        SkipAutoComponentCreation,
+        is_auto_component_creation_suppressed,
+    )
 
     _UPSTREAM_AVAILABLE = True
 except ImportError:
--- a/nautobot_ssot/contrib/component_creation.py
+++ b/nautobot_ssot/contrib/component_creation.py
@@ -1,0 +1,68 @@
+"""SSoT integration with Nautobot core's SkipAutoComponentCreation extension point.
+
+Re-exports :class:`nautobot.apps.dcim.SkipAutoComponentCreation` and
+:func:`nautobot.apps.dcim.is_auto_component_creation_suppressed` when running
+against a Nautobot version that ships them (the upstream extension point is
+proposed in https://github.com/nautobot/nautobot/issues/9026 and tracked by
+https://github.com/nautobot/nautobot/pull/9027).
+
+On older Nautobot versions where the upstream symbol is unavailable, the
+re-exports fall back to a no-op context manager and a function that always
+returns ``False``. The fallback context manager emits a one-shot logger
+warning the first time it is entered, so opt-in users can see why their
+suppression is not taking effect rather than the feature silently doing
+nothing.
+
+Apps should import these names from :mod:`nautobot_ssot.contrib` (or from
+this module) rather than directly from :mod:`nautobot.apps.dcim`, so the
+upgrade path remains forward-compatible.
+"""
+
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+try:
+    from nautobot.apps.dcim import SkipAutoComponentCreation, is_auto_component_creation_suppressed
+
+    _UPSTREAM_AVAILABLE = True
+except ImportError:
+    _UPSTREAM_AVAILABLE = False
+
+    class SkipAutoComponentCreation:
+        """No-op fallback when the upstream ``nautobot.apps.dcim`` API is unavailable.
+
+        Emits a one-shot logger warning the first time the context is entered so
+        opt-in users see why suppression is not taking effect. The context
+        manager otherwise behaves like a no-op: it does not raise and does not
+        affect the wrapped block.
+        """
+
+        _warned = False
+
+        def __enter__(self):
+            """Emit a one-shot warning that the upstream API is unavailable; otherwise no-op."""
+            cls = type(self)
+            if not cls._warned:
+                logger.warning(
+                    "nautobot.apps.dcim.SkipAutoComponentCreation is not available in "
+                    "this Nautobot installation; SSoT's skip_auto_component_creation "
+                    "opt-in will be a no-op. Upgrade Nautobot to a version that ships "
+                    "the upstream extension point to use this feature."
+                )
+                cls._warned = True
+            return self
+
+        def __exit__(self, exc_type, exc, tb):
+            """Do not suppress exceptions raised inside the with block."""
+            return False
+
+    def is_auto_component_creation_suppressed() -> bool:
+        """Return ``False`` always; the upstream extension point is unavailable."""
+        return False
+
+
+def upstream_available() -> bool:
+    """Return ``True`` if Nautobot core's ``SkipAutoComponentCreation`` API is importable."""
+    return _UPSTREAM_AVAILABLE
--- a/nautobot_ssot/contrib/component_creation.py
+++ b/nautobot_ssot/contrib/component_creation.py
@@ -24,7 +24,7 @@ logger = logging.getLogger(__name__)
 
 
 try:
-    from nautobot.apps.dcim import SkipAutoComponentCreation, is_auto_component_creation_suppressed
+    from nautobot.apps.dcim import SkipAutoComponentCreation, is_auto_component_creation_suppressed  # pylint: disable=unused-import
 
     _UPSTREAM_AVAILABLE = True
 except ImportError:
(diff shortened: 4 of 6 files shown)


jdrew82:
## Added unit test coverage for the opt-in decision logic

Pushed `571dbdec` adding tests that close a coverage gap in the existing suite. **No production code changed** — tests only.

### The gap

The existing opt-in tests (`test_class_attribute_opt_in`, `test_settings_opt_in`, `test_or_semantics_*`, `test_suppression_released_between_sequential_runs`) all `skipTest(...)` when `nautobot.apps.dcim.SkipAutoComponentCreation` is unavailable. On the pinned Nautobot the suite runs against today (pre-nautobot/nautobot#9027), that means **7 tests skip** and the two pieces of logic this PR actually introduces were left unexercised in CI:

1. The OR-semantics decision in `_maybe_suppress_auto_component_creation` (class attribute **OR** `PLUGINS_CONFIG` setting).
2. The no-op fallback `SkipAutoComponentCreation` (one-shot warning, exception pass-through).

### What was added

Both additions run **deterministically regardless of upstream availability**, so coverage no longer depends on a Nautobot that ships the extension point.

- **`FallbackContextManagerTestCase`** (3 tests) — forces the `ImportError` branch via a `_forced_fallback_module()` helper that intercepts the import and reloads the module (restoring it on exit). Asserts the fallback: reports upstream unavailable + is a no-op; emits its warning **exactly once**; does **not** suppress exceptions raised inside the block.
- **Decorator truth-table** (4 tests) — spies on the suppression context to assert the full attribute×setting matrix (neither / attribute-only / setting-only / both), each pinning the `PLUGINS_CONFIG` value explicitly via `patch.dict` to avoid ambient-config dependence. This adds the previously-missing OR row (setting `True` / attribute `False`) and makes the decision matrix verifiable without upstream.

### Verification

- `ruff check` + `ruff format --check`: clean.
- `invoke unittest --label nautobot_ssot.tests.test_component_creation`: **18 run / 7 skip** (the 7 upstream-requiring suppression-effect tests still skip as before; the 7 new tests all run).
- `invoke unittest --label nautobot_ssot.tests.test_jobs`: **78/78 pass** — no regression.
- Both suites in one process: **96 pass** — confirms the module-reload helper restores state and does not contaminate other tests.

### Reviewer note

While testing I noticed that with a `PLUGINS_CONFIG["nautobot_ssot"]["skip_auto_component_creation"] = True` override active, the existing `test_default_no_suppression` would fail on a Nautobot that *does* ship the upstream symbol (the job would suppress during sync). The committed default is `False`, so CI is unaffected — but once nautobot/nautobot#9027 lands and the floor is raised, it may be worth pinning the setting explicitly in that test too, consistent with the new tests.

garymccann (approved):
LGTM

PR #1256, opened 2026-06-11
Fix CICD bug and increase test coverage to 100% for non-integration code

### LINKED ISSUE #1255: Flakey CICD "wall clock" timing issue on Parallel Adapter tests.
### Environment
* Python version:  3.13
* Nautobot version:  latest

### Expected Behaviour

The parallel loading tests are consistent in any test environment and do not routinely fail due to environmental conditions on the test runner

### Observed Behaviour

The parallel loading tests routinely fail when ran in a contended CICD environment due to slow python test flows.

### Steps to Reproduce
1. Run CICD and note the time the run takes. 
2. Wait for the CICD step (most notably the mysql matrix test)
3. Observe that the test fails regardless of the code under test. 

======================================================================
FAIL: test_parallel_loading_with_mock_adapters (nautobot_ssot.tests.test_jobs.BaseJobTestCase.test_parallel_loading_with_mock_adapters)
Test parallel loading with mock adapters that simulate work.
----------------------------------------------------------------------
  File "/source/nautobot_ssot/tests/test_jobs.py", line 269, in test_parallel_loading_with_mock_adapters
    self.assertLess(end_time - start_time, 0.15)
    ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: 0.19577240943908691 not less than 0.15

Stdout:


----------------------------------------------------------------------
Ran 1341 tests in 253.379s

FAILED (failures=1, skipped=18)
Destroying test database for alias 'default'...

Error: Process completed with exit code 1.


### Root Cause

This is down to a "wall clock" test. It tests that the two adapters which sleep for 0.1 seconds each, would fail against a 0.2 second timer if they operate sequentially, and alternatively pass if they operate in parallel. However this test does not take into account that the test itself can, and does, breach the 0.2 second timer due to a slow test execution. Slow test execution is a result of the tests themselves running on a busy resource - in this case a Github runner, open to many parallel github actions.

The test is appropriate and tests the correct thing, but the method is flawed when considered against CICD Runners. 

I would propose a logic of checking that the threads exist in parallel (regardless of timings) rather than rely on a realtime clock ("wall clock") to validate.

garymccann:
# Closes: #1255

## What's Changed

This PR fixes the intermittent wall-clock CI failure tracked in #1255 as the main benefit. 
This PR also brings the **core** `nautobot_ssot` modules (everything that is **not** under `contrib/` or `integrations/`) to **100% unit-test coverage** and restructures the jobs test suite so it runs faster and is no longer order/timing-fragile, hardens the suite against `--keepdb` reuse.

It is a **tests-only** change apart from one dead-code removal in `jobs/base.py` (detailed below).

### Summary of changes

| # | Change | Type | Area | Detail |
|---|--------|------|------|--------|
| 1 | Remove unreachable defensive guard | housekeeping | `jobs/base.py` | `sync_data()` had an `if adapter_load_end_time is None:` fallback that both the parallel and sequential branches make unreachable. Removed the dead guard (behavior-neutral) rather than excluding it from coverage with a pragma. |
| 2 | Fix intermittent wall-clock CI failure | housekeeping | `tests/test_jobs.py` | **(#1255)** The parallel-loading tests asserted wall-clock **upper bounds** (e.g. `end_time - start_time < 0.15s`), which flaked on busy MySQL runners (observed `0.196 > 0.15`). Replaced the timing race with a deterministic `threading.Barrier(2)` concurrency proof and removed the remaining wall-clock upper-bound assertions (sleep-guaranteed lower bounds retained). |
| 3 | Core modules to 100% coverage | housekeeping | `nautobot_ssot/*` (non-contrib/integration) | Added/extended tests so every core module reports 100%: `jobs/base.py`, `jobs/__init__.py`, `views.py`, `models.py`, `metrics.py`, `utils/*`, all `templatetags/*`, `exceptions.py`, `template_content.py`, `navigation.py`, `urls.py`, `api/urls.py`, and the app `__init__.py`. |
| 4 | Harden suite against `--keepdb` | housekeeping | `tests/utils/test_orm.py`, `tests/utils/test_cache.py`, `tests/test_utils.py` | `TransactionTestCase` flushes the migration-seeded default `Status` rows and does not restore them, so a `--keepdb` re-run failed Status-dependent setups with `Status.DoesNotExist`. Those setups now call `populate_status_choices()` first, so they self-heal on a reused DB. |
| 5 | New focused test modules | housekeeping | `tests/` | Added `test_exceptions.py`, `test_humanize_bytes.py`, `test_metrics.py`, `test_jobs_base.py`, `test_integration_loading.py`, `test_app_config.py`, `test_template_content.py`. |
| 6 | Extended existing test modules | housekeeping | `tests/` | Extended `test_models.py`, `test_render_diff.py`, `test_utils.py`, `test_views.py`, and `tests/utils/test_orm.py` to cover the previously-missed branches (e.g. `render_diff` pagination edge cases, `Sync` display/encoder paths, `orm` generic-FK + guard branches, view panels and the config view). |
| 7 | Restructure `test_jobs.py` | housekeeping | `tests/test_jobs.py` | Split the monolithic `BaseJobTestCase(TransactionTestCase)` — whose ~35 behavior tests were **re-run 3×** via `DataSource`/`DataTarget` subclasses — into `JobBehaviorTestCase(TestCase)` (non-threaded behavior, runs once), `ParallelLoadingTestCase(TransactionTestCase)` (only the real-thread parallel-loading tests), and two tiny property-only classes. Same assertions, no inheritance fan-out. Also isolates `structlog` per test: `Job.run()` reconfigures `structlog` process-wide, so each jobs test now captures/restores that config to prevent a job-bound log processor from leaking into later tests (which could otherwise write `SyncLogEntry` rows across transactions). |
| 8 | Light fixture de-duplication | housekeeping | `tests/utils/job_helpers.py`, `test_template_content.py`, `test_views.py`, `test_integration_loading.py` | Added a shared `create_example_sync()` helper, and collapsed the three near-identical module `_add_integrations` tests into one `subTest`-driven test. |

### Coverage

| Scope | Before | After |
|-------|--------|-------|
| Core modules (non-contrib, non-integration) | ~77% | **100%** (0 missing lines across all 30 files) |
| `jobs/base.py` | ~91% | 100% |
| `views.py` | ~96% | 100% |
| `metrics.py`, `utils/__init__.py`, `exceptions.py`, `humanize_bytes.py`, … | mixed | 100% |
| `jobs/examples.py` (shipped demo code) | 49% | 49% — **deferred** (see below) |

### Suite health

| Metric | Before | After |
|--------|--------|-------|
| Core test count | 303 | **239** (removed 3× redundant jobs runs) |
| Core suite runtime (local) | ~85s | **~21s** |
| `--keepdb` re-run | failed (`Status.DoesNotExist`) | passes |
| CI wall-clock flake (#1255) | ~20% on MySQL matrix | eliminated (deterministic) |

## To Do
- [x] Unit, Integration Tests

### Notes / constraints
- **No production behavior change** other than removing the one provably-unreachable defensive line in `jobs/base.py`.
- **Out of scope (intentional):** `jobs/examples.py` coverage is deferred to another future PR.


garymccann:
Anecdotal - however, we have a net 2 extra tests and a 58 second improvement on the CICD tests runtime. 

10th June 2026
[unittest (3.10, postgresql, 3.0.0)](https://github.com/nautobot/nautobot-app-ssot/pull/1251#logs)
Ran 1339 tests in 264.080s  OK (skipped=25)

This PR
[unittest (3.10, postgresql, 3.0.0)](https://github.com/nautobot/nautobot-app-ssot/pull/1256#logs)
Ran 1341 tests in 206.057s   OK (skipped=25)

jdrew82 (approved):

PR #1258, opened 2026-06-15
Fix contrib adapter to handle children with OneToOneField

pszulczewski:
## What's Changed

In my use-case I have a custom model which is `OneToOneField` related to the `Device`.
When this custom model is referenced in a DiffSycModel `_children` then it raises `AttributeError`  `object has no attribute 'all'` as `_handle_children` only expects reverse FK as M2M manager with `.all()`.

This fixes the issue described above and allows to use `OneToOneField` under `_children`.

## To Do

### DIFF
--- a/nautobot_ssot/contrib/adapter.py
+++ b/nautobot_ssot/contrib/adapter.py
@@ -156,8 +156,14 @@ class NautobotAdapter(Adapter, BaseNautobotAdapter):  # pylint: disable=too-many
 
     def _handle_children(self, database_object, diffsync_model: BaseNautobotModel):
         """Recurse through all the children for this model."""
+        available_fields = {field.name for field in diffsync_model._model._meta.get_fields()}
         for children_parameter, children_field in diffsync_model._children.items():
-            if not hasattr(database_object, children_field):  # covers OneToOneField
+            if children_field not in available_fields:
+                raise AttributeError(
+                    f"'{diffsync_model._model.__name__}' has no field '{children_field}'. "
+                    f"Check the '_children' mapping on '{diffsync_model.__class__.__name__}'."
+                )
+            if not hasattr(database_object, children_field):  # covers OneToOneField with no related object
                 continue
             _children = getattr(database_object, children_field)
             children = _children.all() if hasattr(_children, "all") else [_children]
--- a/nautobot_ssot/contrib/adapter.py
+++ b/nautobot_ssot/contrib/adapter.py
@@ -157,7 +157,10 @@ class NautobotAdapter(Adapter, BaseNautobotAdapter):  # pylint: disable=too-many
     def _handle_children(self, database_object, diffsync_model: BaseNautobotModel):
         """Recurse through all the children for this model."""
         for children_parameter, children_field in diffsync_model._children.items():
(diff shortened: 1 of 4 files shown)


nautobot_ssot/contrib/adapter.py
```diff
@@ -157,7 +157,10 @@ def _load_single_object(self, database_object, diffsync_model, parameter_names):
     def _handle_children(self, database_object, diffsync_model: BaseNautobotModel):
         """Recurse through all the children for this model."""
         for children_parameter, children_field in diffsync_model._children.items():
-            children = getattr(database_object, children_field).all()
+            if not hasattr(database_object, children_field):  # covers OneToOneField
+                continue
```
garymccann, line 167: hasattr guard silently hides misconfiguration.
Previously, a typo in a _children field name raised AttributeError. Now any unrecognised children_field is silently continued, so a genuine mistake produces no children and no error.

Would it be possible to check the class and object to differentiate an empty OneToOne from a modelling error?

garymccann (changes requested):
Possible fix required, open to opinions. The PR is fine, just want to ensure no unwanted side effects (developers will make mistakes, needs clear alerting, not silent fail)

nautobot_ssot/contrib/adapter.py
pszulczewski: I was thinking the same, but this is actually handled by pydantic validation on import.

If the attribute does not exist, an AttributeError exception is raised on class import.
```
    module = __import__(module_name)
             ^^^^^^^^^^^^^^^^^^^^^^^
  File "/source/nautobot_ssot/tests/test_contrib_adapter.py", line 22, in <module>
    from nautobot_ssot.tests.contrib_base_classes import (
  File "/source/nautobot_ssot/tests/contrib_base_classes.py", line 144, in <module>
    class NautobotDeviceInvalidChild(NautobotModel):
  File "/usr/local/lib/python3.12/site-packages/pydantic/_internal/_model_construction.py", line 270, in __new__
    super(cls, cls).__pydantic_init_subclass__(**kwargs)  # type: ignore[misc]
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/diffsync/__init__.py", line 147, in __pydantic_init_subclass__
    raise AttributeError(f"_children {cls._children} references missing or un-annotated attr {attr}")
AttributeError: _children {'device_bay': 'invalid_attr'} references missing or un-annotated attr invalid_attr
```

I added a test to check this specific scenario, see the latest commit 😉 .

pszulczewski: I've looked again and I see that I missed field declaration in the DiffSync model, so the attribute error. Looks like it can be declared in the way that invalid Django model attribute is silently continued, will fix it in a while.

pszulczewski: Fixed

garymccann: thanks

garymccann (approved):
LGTM

PR #1259, opened 2026-06-15
Backport #1244: skip_auto_component_creation opt-in for SSoT jobs (#1226)

### LINKED ISSUE #1226: Add `skip_component_autocreation` opt-in to suppress automatic Device/Module component instantiation during SSoT syncs

jdrew82:
# Backport of #1244 to `ltm-2.4`

Backports the `skip_auto_component_creation` opt-in for SSoT jobs (closes #1226) onto the `ltm-2.4` LTM line. Original feature merged to `develop` in #1244 (squash commit `c84ef5b7`).

## What's changed

SSoT jobs can opt in to suppressing Nautobot's automatic `Device`/`Module` component instantiation during a sync, via three layers (OR semantics — any source set to `True` opts in):

1. **Context manager** — `from nautobot_ssot.contrib import SkipAutoComponentCreation`
2. **`DataSyncBaseJob` class attribute** — `skip_auto_component_creation = True`
3. **`PLUGINS_CONFIG` setting** — `"skip_auto_component_creation": True`

Default behaviour is unchanged for any job that does not opt in.

## Upstream dependency status

The upstream extension point this feature consumes — `nautobot.apps.dcim.SkipAutoComponentCreation` — has merged into Nautobot core (issue nautobot/nautobot#9026, PR nautobot/nautobot#9027, merged 2026-06-08 into core `develop`) and will ship in the next Nautobot feature release.

The matching **LTM backport into Nautobot core's `ltm-2.4` branch** is tracked in **nautobot/nautobot#9066** (currently open). Once #9066 merges and ships in an LTM Nautobot release, the extension point is present on the LTM line and this opt-in is fully active there.

This SSoT backport is safe to merge regardless of #9066's timing:

- `nautobot_ssot/contrib/component_creation.py` feature-detects the upstream symbol. Where it is present (an LTM Nautobot release that includes #9066, or any release with #9027), the opt-in is fully active. Where it is absent, the opt-in is a **no-op** and emits a **one-shot warning** the first time the suppression path is exercised, so operators see why suppression isn't taking effect rather than it silently doing nothing.
- No further SSoT code change is required once the Nautobot floor on a given install includes the upstream symbol.

## Backport notes

- Applied as a clean cherry-pick of the squashed merge commit `c84ef5b7`.
- One conflict resolved in `nautobot_ssot/jobs/base.py`: `ltm-2.4` does not contain develop's parallel-adapter-loading machinery, so only the feature-relevant pieces were carried over — the `import functools`, the `_maybe_suppress_auto_component_creation` decorator, and its application to `sync_data()`. `ltm-2.4`'s existing `sync_data()` signature (`# pylint: disable=too-many-statements`) was preserved; no unrelated develop-only code was pulled in.

## Verification (against the `ltm-2.4` base)

- `ruff check` + `ruff format --check`: clean.
- `invoke pylint`: **10.00/10**, no findings on changed files.
- `invoke unittest --label nautobot_ssot.tests.test_component_creation`: **18 run / 7 skip** (the 7 upstream-requiring suppression-effect tests skip because the LTM Nautobot in the test environment does not yet ship the extension point; the upstream-independent decorator and fallback tests all run and pass). All 11 of those upstream tests run once the LTM Nautobot floor includes nautobot/nautobot#9066.


garymccann (approved):
LGTM

Issue #940, opened 2025-08-20
Citrix ADM Throws Error When Creating IP Address


jdrew82:
### Environment
* Python version:  3.10.4
* Nautobot version:  2.4.13
* nautobot-ssot version:  3.9.2

### Expected Behavior
I expect to see the Job complete successfully.

### Observed Behavior
I see an error about IPAddress already existing:

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.12/site-packages/celery/app/trace.py", line 453, in trace_task
    R = retval = fun(*args, **kwargs)
                 ^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/celery/app/trace.py", line 736, in __protected_call__
    return self.run(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/nautobot/extras/jobs.py", line 1288, in run_job
    result = job(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/nautobot/extras/jobs.py", line 164, in __call__
    return self.run(*args, **deserialized_kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/nautobot_ssot/integrations/citrix_adm/jobs.py", line 143, in run
    super().run(dryrun=self.dryrun, memory_profiling=self.memory_profiling, *args, **kwargs)
  File "/usr/local/lib/python3.12/site-packages/nautobot_ssot/jobs/base.py", line 367, in run
    self.sync_data(memory_profiling)
  File "/usr/local/lib/python3.12/site-packages/nautobot_ssot/jobs/base.py", line 209, in sync_data
    self.execute_sync()
  File "/usr/local/lib/python3.12/site-packages/nautobot_ssot/jobs/base.py", line 100, in execute_sync
    self.source_adapter.sync_to(self.target_adapter, flags=self.diffsync_flags)
  File "/usr/local/lib/python3.12/site-packages/diffsync/__init__.py", line 627, in sync_to
    return target.sync_from(self, diff_class=diff_class, flags=flags, callback=callback, diff=diff)
[...]

PR #1262, opened 2026-06-15
Make Arista CloudVision NautobotIPAddress.create idempotent (existing IP no longer aborts sync)

### LINKED ISSUE #1261: Arista CloudVision SSoT: sync aborts on existing IP (NautobotIPAddress.create not idempotent)
### Environment
* Python version: 3.11
* Nautobot version: 3.1.0 (also reproduces on current `develop`)
* nautobot-ssot version: 4.2.2 (and `develop`)

### Expected Behavior

The "Arista CloudVision ⟹ Nautobot" SSoT job should be idempotent against pre-existing IPAM data — if CloudVision reports an IP address that already exists in Nautobot (same address / parent prefix / namespace), the job should reuse the existing `IPAddress` and continue, not abort.

### Observed Behavior

The job fails on every run when CloudVision reports an already-existing IP:

```
django.core.exceptions.ValidationError: {'__all__': ['IP address with this Parent and Host already exists.']}
```

Root cause: `NautobotIPAddress.create()` in `nautobot_ssot/integrations/aristacv/diffsync/models/nautobot.py` unconditionally instantiates a new `IPAddress` and calls `validated_save()` with no get-or-create / existing-object path:

```python
new_ip = OrmIPAddress(address=ids["address"], parent=..., status=OrmStatus.objects.get(name="Active"))
new_ip.validated_save()
```

When the address already exists under the same parent, the uniqueness `ValidationError` propagates uncaught and the entire sync aborts. (The sibling `NautobotIPAssignment.create()` in the same file already fetches existing objects via `.get()`; this is the same bug class as #940, fixed for the Citrix ADM integration.)

### Steps to Reproduce
1. Have an `IPAddress` already present in Nautobot IPAM (e.g. from a prior sync or another source).
2. Run the Arista CloudVision ⟹ Nautobot SSoT job where CloudVision reports that same address under the same parent prefix/namespace.
3. The job aborts with `ValidationError: IP address with this Parent and Host already exists.`

Verified on Nautobot 3.1 by calling `validated_save()` on a second `IPAddress` with the same address+parent — raises the exact error above; `IPAddress.objects.get_or_create(address=..., parent=..., defaults={"status": ...})` resolves it idempotently (reuses the existing row; still creates new IPs normally).

### Suggested Fix

Make `NautobotIPAddress.create()` idempotent via `get_or_create` (matching the codebase idiom). Open follow-up question for maintainers: whether an existing IP's attributes (e.g. status) should be *reconciled* to the CloudVision-reported values on match, or left as-is — these objects reach `create()` (not `update()`), so they may not currently be on the attribute-reconciliation path.

nathanhitchcock:
# Closes: #1261

## What's Changed

`NautobotIPAddress.create()` in the Arista CloudVision integration unconditionally instantiated a new `IPAddress` and called `validated_save()`. When CloudVision reported an address that already existed in Nautobot under the same parent, the uniqueness `ValidationError` ("IP address with this Parent and Host already exists") propagated uncaught and aborted the entire sync — the integration was not idempotent against pre-existing IPAM data.

This makes the create path idempotent with `get_or_create` (matching the codebase idiom — Citrix ADM uses `get_or_create`, and the sibling `NautobotIPAssignment.create()` already fetches existing objects). Same bug class as #940 (Citrix ADM).

**Reproduction** (core ORM, Nautobot 3.1, scrubbed RFC5737 IPs):

```
# pre-existing IPAddress 192.0.2.10/24, then the old create() path:
IPAddress(address="192.0.2.10/24", parent=..., status=Active).validated_save()
#   -> ValidationError: ['IP address with this Parent and Host already exists.']  (sync aborts)

# with the fix:
IPAddress.objects.get_or_create(address="192.0.2.10/24", parent=..., defaults={"status": Active})
#   existing -> created=False, reuses the existing row (no crash)
#   new IP   -> created=True (happy path unchanged)
```

## Possible follow-up (out of scope here)

On match, this reuses the existing `IPAddress` as-is rather than reconciling its attributes (e.g. status) to the CloudVision-reported values. Whether the source should overwrite existing Nautobot data is a precedence decision for maintainers — and these objects reach `create()` rather than `update()`, so they may not be on the attribute-reconciliation path. Flagged in #1261; intentionally not addressed here to keep the crash fix minimal.

## To Do
- [x] Added change log fragment(s) (`changes/1261.fixed`)
- [ ] Attached Screenshots, Payload Example — N/A (behavior change in create path)
- [ ] Unit, Integration Tests — `NautobotIPAddress.create()` has no existing unit coverage; happy to add if maintainers want it

### DIFF
--- a/nautobot_ssot/integrations/aristacv/diffsync/models/nautobot.py
+++ b/nautobot_ssot/integrations/aristacv/diffsync/models/nautobot.py
@@ -266,14 +266,13 @@ class NautobotIPAddress(IPAddress):
     @classmethod
     def create(cls, adapter, ids, attrs):
         """Create IPAddress in Nautobot."""
-        new_ip = OrmIPAddress(
+        OrmIPAddress.objects.get_or_create(
             address=ids["address"],
             parent=OrmPrefix.objects.get(
                 prefix=ids["prefix"], namespace=OrmNamespace.objects.get(name=ids["namespace"])
             ),
-            status=OrmStatus.objects.get(name="Active"),
+            defaults={"status": OrmStatus.objects.get(name="Active")},
         )
-        new_ip.validated_save()
         return super().create(ids=ids, adapter=adapter, attrs=attrs)
 
     def delete(self):
--- a/changes/1261.fixed
+++ b/changes/1261.fixed
@@ -1,0 +1,1 @@
+Fixed Arista CloudVision SSoT sync aborting when CloudVision reports an `IPAddress` that already exists in Nautobot, by making `NautobotIPAddress.create()` idempotent with `get_or_create`.


nautobot_ssot/integrations/aristacv/diffsync/models/nautobot.py
```diff
@@ -266,14 +266,13 @@ class NautobotIPAddress(IPAddress):
     @classmethod
     def create(cls, adapter, ids, attrs):
         """Create IPAddress in Nautobot."""
-        new_ip = OrmIPAddress(
+        OrmIPAddress.objects.get_or_create(
```
joewesch, line 269: I feel like this is an anti-pattern. If we got to this create method, we flagged the IP Address as not existing in Nautobot and `get_or_create` is redundant. If the IP does indeed exist in Nautobot, it should have landed on the update method, not create.

I think we should evaluate why it didn't load the IP.

garymccann:
@nathanhitchcock - happy to close this one out in favour of #1268 ?

nathanhitchcock:
@garymccann Sounds good, #1268 is the more complete fix (loads existing IPs so it goes down the update path). Closed in favor.

PR #1265, opened 2026-06-15
Fixed token auth for on-prem Arista CV

### LINKED ISSUE #1235: Token authentication broken against on-prem Arista CloudVision portal after SSoT upgrade to 4.3.0
### Environment
* Python version:   3.13
* Nautobot version:   3.x
* nautobot-ssot version:   **4.3.0**

### Expected Behavior
Successfull authentication against an on-premise Cloudvision portal with a cvp_user and a bearer token.

### Observed Behavior
after the upgrade from 4.2.2 to 4.3.0, the Arista SSoT authentication starts failing with:

```python
{
    "exc_message": [
        "nautobot_ssot.exceptions",
        "AuthFailure",
        [
            "Unable to login to CloudVision Portal. <cvp host>: Authenticate: https://<cvp host>:443/web/login/authenticate.do: Request Error: Invalid credentials\n"
        ],
        "AuthFailure('Unable to login to CloudVision Portal. \\<cvp host>: Authenticate: https://<cvp host>:443/web/login/authenticate.do: Request Error: Invalid credentials\\n')"
    ],
    "exc_module": "celery.utils.serialization",
    "exc_type": "UnpickleableExceptionWrapper"
}
```

### Steps to Reproduce
1. setup a working Arista SSoT 4.2.2 integration against an on-premise cloudVision portal instance, with token authentication
2. upgrade SSoT to 4.3.0
3. authentication stops working

### Analysis

This is a regression introduced in 4.3.0 by commit 9b81dc60 ("Optimized AristaCV device loading"), in
`nautobot_ssot/integrations/aristacv/utils/cloudvision.py`

In 4.2.2 there was no persistent `self.cvpclient` attribute on the CloudvisionApi class. It [appeared in 4.3.0 ](https://github.com/nautobot/nautobot-app-ssot/blob/8435f3f4a4581bef5cebe9715b136783c2f3dd29/nautobot_ssot/integrations/aristacv/utils/cloudvision.py#L86) due to the above optmization effort.

I think the bug is inside the `_connect_rest_client()` which is used to initialize `self.cvpclient`:

```python
try:
    if config.token and not config.is_on_premise:
        client.connect(
            nodes=[parsed_url.hostname],
            username="",
            password="",
            is_cvaas=True,
            api_token=config.token,
        )
    else:
        client.connect(
            nodes=[parsed_url.hostname],
            username=config.cvp_user,
            password=config.cvp_password,
            is_cvaas=False,
        )
```

Basically you are allowed to use the token only if `config.is_on_premise` is `False`. Else you're using `config.cvp_password` which is likely blank. That's why local auth is failing.

This bug was latent in 4.2.2 as well but one could bump into it only if `create_controller` setting was set to True (default to False), since the problematic path was not run. Also in 4.2.2 the REST client was constructed only as a local variable inside the standalone function `get_cvp_version(config)`.

Cf. NTC-5983

joewesch:
# Closes: #1235

## What's Changed
This PR fixes a bug that was exposed via #1197 where we were only allowing either token auth for CVaaS or username/password for on-prem. Now, we accept token auth for on-prem Arista CV.

## To Do

- [x] Unit, Integration Tests

NTC-5983

### DIFF
--- a/changes/1235.fixed
+++ b/changes/1235.fixed
@@ -1,0 +1,1 @@
+Fixed token authentication against on-premise Arista CloudVision portals.
--- a/nautobot_ssot/integrations/aristacv/utils/cloudvision.py
+++ b/nautobot_ssot/integrations/aristacv/utils/cloudvision.py
@@ -249,20 +249,25 @@ class CloudvisionApi:  # pylint: disable=too-many-instance-attributes, too-many-
         if not parsed_url.hostname:
             raise ValueError(f"Invalid URL provided for CloudVision. {config.url}")
         try:
-            if config.token and not config.is_on_premise:
+            if config.token:
                 client.connect(
                     nodes=[parsed_url.hostname],
                     username="",
                     password="",
-                    is_cvaas=True,
+                    is_cvaas=not config.is_on_premise,
                     api_token=config.token,
                 )
-            else:
+            elif config.cvp_user and config.cvp_password:
                 client.connect(
                     nodes=[parsed_url.hostname],
                     username=config.cvp_user,
                     password=config.cvp_password,
                     is_cvaas=False,
+                )
+            else:
+                raise AuthFailure(
+                    error_code="Missing Credentials",
+                    message="Unable to authenticate due to missing credentials.",
                 )
         except CvpLoginError as err:
             raise AuthFailure(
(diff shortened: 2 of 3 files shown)


garymccann (approved):
LGTM and approve test coverage (low score is due to not following an obvious exception path)

PR #1267, opened 2026-06-16
[AristaCV] Fixed parsing interface attributes across split notifications

### LINKED ISSUE #1245: Arista CV SSoT: gRPC split-notification streams drop interface state data resulting in KeyError
### Environment
* Python version: 3.12.13
* Nautobot version: v3.1.0
* nautobot-ssot version: 4.2.2

### Expected Behavior
All gRPC state-streaming response notifications from Arista CloudVision for a given network object (such as physical interfaces or port-channels) should be dynamically combined and synchronized into a single unified record. This must occur regardless of whether a specific notification frame contains the `intfId` or `name` update keys. 

Fields like `enabled` (derived from `enabledState`), `operStatus`, and `linkStatus` must be correctly merged into the final state payload by extracting and utilizing the trailing path element (`notif["path_elements"][-1]`) as a reliable object anchor. This tracking standard ensures data integrity across the Single Source of Truth (SSoT) and perfectly aligns with the aggregation pattern previously implemented for IP addresses in PR #1188.

### Observed Behavior
The SSoT sync job ("Arista CloudVision → Nautobot") fails during the interface processing phase with a fatal `KeyError: 'enabled'` inside the `load_interfaces` method:

```json
{
    "exc_message": ["enabled"],
    "exc_module": "builtins",
    "exc_type": "KeyError"
}
```

#### Technical Root Cause
Inside `nautobot_ssot/integrations/aristacv/utils/cloudvision.py`, multiple utility methods evaluate streamed notifications assuming all key-value attributes arrive simultaneously in every single iteration loop. 

Based on my debugging, Arista CloudVision streams updates asynchronously as differential slices:
* **Frame 1:** CloudVision streams structural updates along with the explicit identifier key `intfId: "Ethernet1"`.
* **Frame 2:** A subsequent frame updates the interface state (`'updates': {'enabledState': {'Name': 'enabled', 'Value': 1}}`) but entirely omits the `intfId` key. The identity of the interface is only present within the metadata path elements: `['Sysdb', ..., 'intfStatus', 'Ethernet1']`.

Because current utility methods utilize rigid structural guard clauses (e.g., `if not results.get("intfId"): continue`), Frame 2 is completely discarded. Downstream parsing logic eventually encounters half-populated datasets (such as `{'interface': 'Ethernet1', 'mac_addr': 'fc:bd:67:8c:25:6c'}`) that lack the expected state keys, resulting in a crashed sync worker.

#### Impacted Functions Audited
1. **`get_interfaces_fixed()`**: Skips notification frames lacking the explicit `intfId` key, losing the `enabledState` data.
2. **`get_interfaces_chassis()`**: Implements the exact same loop vulnerability inside modular linecard (`queryLC`) evaluations.
3. **`get_interfaces_port_channel()`**: Uses a fragile mix of `results.get("intfId")` and `results.get("name")` across status/config paths, causing split asynchronous updates to result in incomplete port-channel records.

---

### Suggested Architectural Fix
Refactor the vulnerable streaming loops to use a temporary state aggregation cache dictionary (`per_intf`) keyed by the final element of `path_elements`. This safely stitches fragmented gRPC responses back together before compiling the final list.

#### Remediation Pattern for `get_interfaces_fixed`:
```python
def get_interfaces_fixed(client: CloudvisionApi, dId: str):
    """Gets information about interfaces for a fixed system device."""
    pathElts = ["Sysdb", "interface", "status", "eth", "phy", "slice", "1", "intfStatus", Wildcard()]

    per_intf = {}
    for batch in clean_query_results(pathElts, client, dId):
        for notif in batch["notifications"]:
            # Secure the anchor identifier from the gRPC metadata path elements
            intf_name = notif["path_elements"][-1]
            entry = per_intf.setdefault(intf_name, {"interface": intf_name})
            
            updates = notif["updates"]
            if updates.get("intfId"):
                entry["interface"] = updates["intfId"]
            if updates.get("enabledState"):
                entry["enabled"] = bool(updates["enabledState"]["Name"] == "enabled")
            if updates.get("burnedInAddr"):
                entry["mac_addr"] = updates["burnedInAddr"]
            if updates.get("mtu"):
                entry["mtu"] = updates["mtu"]
            if updates.get("operStatus"):
                entry["oper_status"] = "up" if updates["operStatus"]["Name"] == "intfOperUp" else "down"
            if updates.get("linkStatus"):
                entry["link_status"] = "up" if updates["linkStatus"]["Name"] == "linkUp" else "down"

    return list(per_intf.values())
```

---

### Steps to Reproduce
1. Configure the Arista CloudVision SSoT integration against an active CloudVision cluster managing devices where interface parameters are streamed across split notification frames.
2. Execute an inbound synchronization job with logging set to `debug=True`.
3. Observe the job logs: `get_interfaces_fixed()` returns unpopulated data slices for interfaces whose state updates arrived separately from their identity updates, immediately triggering a fatal `KeyError` in the job engine.

NTC-6027
NTC-6005

joewesch:
# Closes: #1245

## What's Changed
This PR changes the `get_interfaces_chassis`, `get_interfaces_fixed` and `get_interfaces_port_channel` functions on how they parse interface attributes to match how we parse other objects when using path wildcards. Instead of assuming a name is always present in the attributes for all notifications, it now uses the wildcard element of the gRPC path elements as the key and collates the attributes across notifications before returning them.

## To Do

- [x] Unit, Integration Tests

NTC-6027

### DIFF
--- a/changes/1245.fixed
+++ b/changes/1245.fixed
@@ -1,0 +1,1 @@
+Fixed Arista CloudVision SSoT sync failing when attributes are streamed in multiple gRPC notification frames.
--- a/nautobot_ssot/integrations/aristacv/utils/cloudvision.py
+++ b/nautobot_ssot/integrations/aristacv/utils/cloudvision.py
@@ -511,30 +511,33 @@ def get_interfaces_chassis(client: CloudvisionApi, dId):
     dataset = dId
     query = get_query(client, dataset, pathElts)
     queryLC = unfreeze_frozen_dict(query).keys()
-    intfStatusChassis = []
 
-    # Go through each linecard and get the state of all interfaces
+    # Group notifications by the interface name from the path elements. CloudVision may stream an
+    # interface's attributes across multiple frames, and a frame carrying state may not contain intfId, so
+    # anchoring on the path element and merging keeps those frames from being lost. Interface names
+    # are unique across linecards, so a single accumulator spans all of them.
+    per_intf = {}
     for lc in queryLC:
         pathElts = ["Sysdb", "interface", "status", "eth", "phy", "slice", lc, "intfStatus", Wildcard()]
 
         for batch in clean_query_results(pathElts, client, dataset):
             for notif in batch["notifications"]:
+                intf_name = notif["path_elements"][-1]
+                entry = per_intf.setdefault(intf_name, {"interface": intf_name})
                 results = notif["updates"]
-                if not results.get("intfId"):
-                    continue
-                new_intf = {"interface": results["intfId"]}
+                if results.get("intfId"):
+                    entry["interface"] = results["intfId"]
                 if results.get("linkStatus"):
-                    new_intf["link_status"] = "up" if results["linkStatus"]["Name"] == "linkUp" else "down"
+                    entry["link_status"] = "up" if results["linkStatus"]["Name"] == "linkUp" else "down"
                 if results.get("operStatus"):
-                    new_intf["oper_status"] = "up" if results["operStatus"]["Name"] == "intfOperUp" else "down"
+                    entry["oper_status"] = "up" if results["operStatus"]["Name"] == "intfOperUp" else "down"
                 if results.get("enabledState"):
-                    new_intf["enabled"] = bool(results["enabledState"]["Name"] == "enabled")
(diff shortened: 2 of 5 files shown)


garymccann (approved):
LGTM

PR #1268, opened 2026-06-16
Fixed Arista CV creating duplicate IPs

### LINKED ISSUE #1261: Arista CloudVision SSoT: sync aborts on existing IP (NautobotIPAddress.create not idempotent)

joewesch:
# Closes: #1261

## What's Changed
This PR replicates the same logic and settings that were added in #1118 to load all IP Addresses from Nautobot instead of just the ones that are attached to Arista devices. This fixes it trying to create a duplicate IP Address and instead attempts to update it instead.

## To Do

- [x] Unit, Integration Tests

NTC-6005


nautobot_ssot/integrations/aristacv/diffsync/adapters/nautobot.py
```diff
@@ -131,37 +131,19 @@ def load_prefixes(self):
...
-                self.add(new_ns)
-            try:
-                self.get(self.prefix, {"prefix": str(ipaddr.parent.prefix), "namespace": ipaddr.parent.namespace.name})
-            except ObjectNotFound:
-                new_pf = self.prefix(
-                    prefix=str(ipaddr.parent.prefix),
-                    namespace=ipaddr.parent.namespace.name,
-                    uuid=ipaddr.parent.id,
-                )
-                self.add(new_pf)
```
joewesch, line 153: All of this is now redundant since we load all Namespaces and Prefixes before loading all IP Addresses

nautobot_ssot/integrations/aristacv/diffsync/adapters/nautobot.py
```diff
@@ -131,37 +131,19 @@ def load_prefixes(self):
...
             new_ip = self.ipaddr(
                 address=str(ipaddr.address),
                 prefix=str(ipaddr.parent.prefix),
                 namespace=ipaddr.parent.namespace.name,
                 uuid=ipaddr.id,
             )
-            try:
-                self.add(new_ip)
-            except ObjectAlreadyExists as err:
-                self.job.logger.warning(f"Unable to load {ipaddr.address}. IPAddress already in diffsync store. {err}")
```
joewesch, line 163: This should have probably been redundant before because of the `.distinct()` that was on the filter above, but it definitely is now.

joewesch:
This PR replaces #1262

jdrew82:
@joewesch Can we get tests covering those missing lines that you added?

joewesch:
Delicious coverage

jdrew82 (approved):

PR #1274, opened 2026-06-21
Pass Port Parameter to Parent Class in LibreNMSApi

ankitk111:
## Summary
Fixed a bug where the port parameter was not being passed to the 
parent ApiEndpoint class in LibreNMSApi.__init__, causing all 
connections to always use port 443.

## How to reproduce
1. Configure LibreNMS External Integration with port 80 in extra_config
2. Run SSoT Sync from LibreNMS job
3. Job always connects to port 443 instead of taking the value from extra_config

## Fix
Added port=port to super().__init__() call on line 107:
super().__init__(url=url, port=port)

## Testing
Verified fix works on a live Nautobot 3.0.11 deployment on 
OpenShift, successfully syncing 32 devices from LibreNMS on port 80.

### DIFF
--- a/changes/1274.fixed
+++ b/changes/1274.fixed
@@ -1,0 +1,1 @@
+Fixed a bug in the LibreNMS integration where the port parameter was not passed to the parent ApiEndpoint class, causing all connections to default to port 443 regardless of the configured port.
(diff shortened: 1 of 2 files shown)


jdrew82 (approved):

jdrew82:
@ankitk111 It appears we're missing a changelog fragment for this change. Please review the contribution guide and add the missing fragment so we can get this merged. Thanks!

garymccann (approved):
LGTM

PR #1276, opened 2026-06-22
Paginate ServiceNow table loads and select only mapped fields

### LINKED ISSUE #1275: ServiceNow integration: all_table_entries() truncates tables >10k rows (no pagination) and fetches all fields
### Environment
* Python version: N/A (defect in app logic, not version-specific)
* Nautobot version: any
* nautobot-ssot version: 4.4.x / current `develop`

### Expected Behavior
The ServiceNow integration's table load (`ServiceNowClient.all_table_entries()`) should:
1. Return **all** records in a table, regardless of row count.
2. Fetch only the fields the mappings actually reference, rather than every column.

### Observed Behavior

**1. Tables larger than 10,000 rows are silently truncated (data-integrity bug).**

`all_table_entries()` (`nautobot_ssot/integrations/servicenow/servicenow.py`) loads records via the streamed reader:

```python
yield from self.resource(api_path=f"/table/{table}").get(query=query, stream=True).all()
```

With the bundled pysnow defaults this sends `sysparm_limit=10000` and **no field selection**. The streamed read path (`third_party/pysnow/response.py`) parses a single response page and never follows ServiceNow's pagination (`Link`) header — the offset/`Link` pagination loop exists only in pysnow's deprecated `legacy_request.py`, which this code path does not use. As a result, any table with more than 10,000 rows returns only the first page; the remaining records are silently dropped.

Because ServiceNow is the *destination* side of the sync, the dropped rows appear absent to DiffSync, which then plans to **create** them — producing a large number of spurious create operations for records that already exist in ServiceNow.

**2. Every column is fetched even though only a few are mapped (inefficiency).**

`all_table_entries()` never sets `sysparm_fields`, so the API returns all columns (150+ on large instances) when the YAML mappings reference only a handful. This inflates payload size and load time with no benefit.

### Steps to Reproduce
1. Configure the ServiceNow SSoT integration against an instance where a mapped table has more than 10,000 rows.
2. Run the sync.
3. Observe (a) the diff plans creates for records that already exist in ServiceNow (the rows beyond the first 10,000), and (b) each table request retrieves all columns rather than just the mapped ones.

### Proposed Resolution
- Paginate `all_table_entries()` so it returns all matching records regardless of count (offset loop over pages, or honoring the `Link` header), instead of relying on a single capped page.
- Auto-derive `sysparm_fields` for each table load from the columns already declared in the mappings (`column` / `reference.key`, plus `sys_id`), with an optional explicit override, so only the required fields are fetched.
- Optionally expose page size / `limit` as configuration.

<sub>Reported by a customer via NTC support.</sub>

nathanhitchcock:
# Closes: #1275

## What's Changed

The ServiceNow integration loaded each table via a single streamed page (`ServiceNowClient.all_table_entries()` → `resource().get(stream=True).all()`). With the pysnow default `sysparm_limit=10000`, and because the streamed read path does not follow ServiceNow's `Link` pagination header (that loop lives only in the deprecated `legacy_request.py`), any table with more than 10,000 rows was silently truncated to its first page. Since ServiceNow is the destination side of the sync, the dropped rows looked absent to DiffSync, which then planned to **create** them — producing spurious create operations for records that already existed in ServiceNow.

### Pagination (fix)
`all_table_entries()` now pages through the full result set using `sysparm_offset`, fetching `limit`-sized pages until a short page signals the end. Behavior is unchanged for tables that fit within a single page.

### Field selection (added)
`all_table_entries()` previously requested every column (150+ on large instances) even though the mappings reference only a handful. `load_table()` now derives `sysparm_fields` from the columns already declared in each table's mappings (`column` / `reference.key`, plus `sys_id`), with an optional explicit `fields` override and a configurable page size (`limit`) in the mapping entry. When neither is set, behavior matches the prior defaults — all columns, 10,000-row pages — now correctly paginated.

A design note for reviewers: field selection defaults to **auto-derivation** from the mappings (with override) rather than requiring manual `fields` config, since the mappings already enumerate exactly the needed columns. Happy to adjust if you'd prefer an opt-in.

## Request / payload example

> Illustrative — constructed to show the request and record shape, not captured from a production instance.

For a `core_company` table with 25,001 rows whose mappings reference only `name` and `manufacturer`:

**Before**
```
GET /api/now/table/core_company?sysparm_limit=10000
# one page -> 10,000 of 25,001 rows; the remaining 15,001 are silently dropped
# every row carries the full column set (~150 fields)
```

**After**
```
GET /api/now/table/core_company?sysparm_limit=10000&sysparm_offset=0&sysparm_fields=manufacturer,name,sys_id
GET /api/now/table/core_company?sysparm_limit=10000&sysparm_offset=10000&sysparm_fields=manufacturer,name,sys_id
GET /api/now/table/core_company?sysparm_limit=10000&sysparm_offset=20000&sysparm_fields=manufacturer,name,sys_id
# three pages -> all 25,001 rows; iteration stops on the short final page
```

Per-record payload shrinks to just the mapped columns:

```jsonc
// Before (one record, abbreviated)
{ "sys_id": "7200ad...", "name": "Acme", "manufacturer": "true",
  "sys_updated_on": "2021-07-12 20:19:23", "sys_created_by": "admin",
  "lat_long_error": "", "...": "~150 fields total" }

// After
{ "sys_id": "7200ad...", "name": "Acme", "manufacturer": "true" }
```

## To Do
- [x] Attached Screenshots, Payload Example (illustrative request/payload example; no live instance for a prod capture)
- [x] Unit, Integration Tests

## Tests
Added to `test_adapter_servicenow.py`:
- `ServiceNowClientPaginationTestCase` — paginates beyond a single page (25,001 rows over offsets 0/10000/20000), terminates on an exact page multiple, and preserves the prior all-columns behavior when `fields` is unspecified.
- `FieldsForMappingsTestCase` — `sysparm_fields` derivation includes `sys_id`, mapped columns, and reference keys.
- Extended `test_filtering` to assert the derived field set is passed through for the `core_company` table.

The pagination loop and field derivation were also verified against the real source in isolation. We don't have a live ServiceNow instance to capture before/after payloads against — happy to add a real-instance example if a maintainer or the reporter can run one.

### DIFF
--- a/changes/1275.added
+++ b/changes/1275.added
@@ -1,0 +1,1 @@
+Added automatic `sysparm_fields` selection to the ServiceNow integration so each table load only requests the columns referenced by its mappings, and made the ServiceNow page size (`limit`) configurable.
--- a/changes/1275.fixed
+++ b/changes/1275.fixed
@@ -1,0 +1,1 @@
+Fixed ServiceNow integration silently truncating tables larger than 10,000 rows; `all_table_entries()` now paginates through the entire result set.
--- a/changes/1275.fixed
+++ b/changes/1275.fixed
@@ -1,1 +1,2 @@
 Fixed ServiceNow integration silently truncating tables larger than 10,000 rows; `all_table_entries()` now paginates through the entire result set.
+Fixed ServiceNow integration raising a spurious `ObjectNotUpdated` on successful updates; updates now write and verify only the mapped fields and key on the record's known `sys_id`, instead of comparing server-managed columns such as `sys_updated_on`.
--- a/nautobot_ssot/integrations/servicenow/diffsync/adapter_servicenow.py
+++ b/nautobot_ssot/integrations/servicenow/diffsync/adapter_servicenow.py
@@ -129,6 +129,17 @@ class ServiceNowDiffSync(Adapter):  # pylint: disable=too-many-instance-attribut
         populated = template.render(config)
         return yaml.safe_load(populated)
 
+    @staticmethod
+    def fields_for_mappings(mappings):
+        """Collect the ServiceNow columns referenced by a set of mappings, for use as `sysparm_fields`."""
+        fields = {"sys_id"}
+        for mapping in mappings:
+            if "column" in mapping:
+                fields.add(mapping["column"])
+            elif "reference" in mapping and "key" in mapping["reference"]:
+                fields.add(mapping["reference"]["key"])
+        return sorted(fields)
+
     def load_table(self, modelname, table, mappings, **kwargs):
         """Load data from the ServiceNow "table" into the DiffSync model.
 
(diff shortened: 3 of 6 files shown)


garymccann (approved):
LGTM

PR #1282, opened 2026-07-01
Allow `_children` mappings to reference model properties, not only concrete fields

### LINKED ISSUE #1281: NautobotAdapter._handle_children guard rejects _children targets that are properties, not concrete fields
Python version: 3.12.12
Nautobot version: 3.0.0
nautobot-ssot version: 4.5.0

**Expected Behavior**
A _children mapping on a contrib NautobotModel should be able to reference a model property that returns a queryset/manager, not just a concrete model field.

For example, mapping a Device's children to its all_interfaces property (a @property that returns Interface.objects.filter(...)) should load those interfaces as children:

class SyncNetworkDataDevice(NautobotModel):
    _model = Device
    _modelname = "device"
    _identifiers = ("name",)
    _children = {"interface": "all_interfaces"}  # all_interfaces is a @property returning a queryset
    ...

**Observed Behavior**
As of 4.5.0, loading raises AttributeError before the children are ever accessed:

`raise AttributeError(\nAttributeError: \'Device\' has no field \'all_interfaces\'. Check the \'_children\' mapping on \'SyncNetworkDataDevice\'.\n',`

This is caused by the new validation guard added to NautobotAdapter._handle_children in 4.5.0:

```
available_fields = {field.name for field in diffsync_model._model._meta.get_fields()}
for children_parameter, children_field in diffsync_model._children.items():
    if children_field not in available_fields:
        raise AttributeError(...)
```
_meta.get_fields() only returns concrete fields and relations — it does not include properties. So a property like all_interfaces is rejected even though it returns a valid queryset and the next lines (getattr(...).all()) would handle it correctly. 

A proposed fix is to widen the guard to also allow attributes resolvable on the model class:
```
if children_field not in available_fields and not hasattr(diffsync_model._model, children_field):
    raise AttributeError(...)
```
This still raises on typos/invalid names (concrete field absent and not an attribute), but allows property-backed _children.

Steps to Reproduce
On nautobot-ssot 4.5.0, define a contrib NautobotModel for Device with _children = {"interface": "all_interfaces"} (where Device.all_interfaces is a @property returning a queryset).
Build a NautobotAdapter that includes that model and a corresponding interface child model, then call adapter.load().
Observe AttributeError: 'Device' has no field 'all_interfaces'. Check the '_children' mapping on '...'. The same setup loads successfully on nautobot-ssot ≤ 4.4.

Dav-C:
Closes: #1281

### What's Changed

The field-existence guard added to `NautobotAdapter._handle_children` in 4.5.0 rejected any `_children` target not present in `_model._meta.get_fields()`. That set only contains concrete fields and relations — it excludes `@property` attributes. As a result, `_children` mappings that point at a property (e.g. a `Device`'s `all_interfaces`, which returns `Interface.objects.filter(...)`) raised:

```
AttributeError: 'Device' has no field 'all_interfaces'.
Check the '_children' mapping on '...'.
```

This is a regression — property-backed `_children` mappings worked in 4.0–4.4, and the very next lines in `_handle_children` (`getattr(...).all()`) handle a returned queryset correctly. The guard rejected the mapping before that handling ever ran.

### Fix

Widen the guard so it accepts attributes resolvable on the model **class** in addition to concrete fields:

```python
if children_field not in available_fields and not hasattr(diffsync_model._model, children_field):
    raise AttributeError(...)
```

- Concrete fields/relations → still pass (in `available_fields`).
- Properties / descriptors → now pass (`hasattr` on the class). Checking the class returns the descriptor without invoking the getter, so there are no side effects or DB queries.
- Genuinely invalid names (typos) → still raise, preserving the misconfiguration safety the guard was added for.

The OneToOneField handling also added in 4.5.0 is untouched.

### Scope / known limitation

This restores the ability to *reference* a property in `_children`. Properties that return a **queryset or manager** (the `all_interfaces` case) load correctly. Properties returning other types (e.g. a plain list or `None`) are still unsupported and may fail in the downstream handling — that broader robustness is intentionally deferred to a follow-up.

### Before / After

**Before (4.5.0):** loading any adapter with a property-backed `_children` mapping raises `AttributeError` immediately.
**After:** the same adapter loads the property's children successfully; invalid mapping names still raise.

## To Do

- [x] Unit, Integration Tests

### DIFF
--- a/changes/1281.fixed
+++ b/changes/1281.fixed
@@ -1,0 +1,1 @@
+Fixed `_children` mappings that reference a model property returning a queryset or manager (such as `Device.all_interfaces`) raising `AttributeError` during load. Properties returning other types are not yet supported.
--- a/nautobot_ssot/contrib/adapter.py
+++ b/nautobot_ssot/contrib/adapter.py
@@ -176,9 +176,12 @@ class NautobotAdapter(Adapter, BaseNautobotAdapter):  # pylint: disable=too-many
         """Recurse through all the children for this model."""
         available_fields = {field.name for field in diffsync_model._model._meta.get_fields()}
         for children_parameter, children_field in diffsync_model._children.items():
-            if children_field not in available_fields:
+            # `children_field` may be a concrete field/relation (present in `_meta.get_fields()`) or a
+            # class-level attribute such as a `@property` that returns a queryset/manager or a related
+            # object. Only raise when it is neither.
+            if children_field not in available_fields and not hasattr(diffsync_model._model, children_field):
                 raise AttributeError(
-                    f"'{diffsync_model._model.__name__}' has no field '{children_field}'. "
+                    f"'{diffsync_model._model.__name__}' has no field or attribute '{children_field}'. "
                     f"Check the '_children' mapping on '{diffsync_model.__class__.__name__}'."
                 )
             if not hasattr(database_object, children_field):  # covers OneToOneField with no related object
--- a/nautobot_ssot/tests/contrib/adapter/test_adapter.py
+++ b/nautobot_ssot/tests/contrib/adapter/test_adapter.py
@@ -30,6 +30,8 @@ from nautobot_ssot.tests.contrib.base import (
     NautobotDeviceBay,
     NautobotDeviceInvalidChildAttr,
     NautobotDeviceWithChildBay,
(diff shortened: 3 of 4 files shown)


garymccann (approved):
LGTM

Issue #1068, opened 2026-01-19
Native Support for Proxmox VE Integration


vmotov:
### Environment

- Python version: 3.13
- Nautobot version: 3.0.0
- nautobot-ssot version: 4.0.0

### Proposed Functionality
I am proposing the addition of a native **Proxmox VE (PVE)** adapter to the SSoT framework, similar to the existing vSphere integration.

Proposed features:
* **Infrastructure Sync:** Automatically map Proxmox Nodes to Nautobot `Devices` and Clusters to `Clusters`.
* **Virtualization Sync:** Map Proxmox VMs and LXC Containers to Nautobot `Virtual Machines` objects.
* **Resource Tracking:** Import hardware specifications including CPU cores, Memory (RAM), and Disk capacity.
* **Network Discovery:** Synchronize VM network interfaces, MAC addresses, and IP addresses into Nautobot IPAM.
* **Secure Auth:** Support for Proxmox API Tokens (PVEAPIToken) for secure, non-interactive access.

### Use Case
As a System Administrator migrating workloads from VMware to Proxmox, I need to maintain a single Source of Truth during and after the transition. Having a native Proxmox adapter allows for concurrent sync jobs from both platforms. I can see my legacy VMware estate and my new Proxmox environment in one place (Nautobot), ensuring that all migrated assets are accounted for, correctly named, and fully documented with their network configurations during the migration process.

garymccann:
noting as duplicate / similar to #876

PR #1288, opened 2026-07-06
Add Proxmox VE SSoT integration

### LINKED ISSUE #876: Feature request: Proxmox VE support
### Environment
* Nautobot version:  2.3.1
* nautobot-ssot version:  3.9.1a0

### Proposed Functionality
Integration with Proxmox VE as a source of truth for Virtual Machines. Nautobot-SSoT should be able to connect to one to N Proxmox VE clusters through Proxmox VE API https://pve.proxmox.com/pve-docs/api-viewer/index.html#/cluster/resources, using an API token.
Virtual Machine information could then be synchronized into database from recurring Nautobot-SSoT job (same as already existing vSphere).
The following elements should be imported:
- Virtual Machine Name
- Virtual Machine ID (through custom field)
- Status (depending on if VM is off/on)
- Resources:
   - number of vCPUs
   - total RAM space
   - total disk space
- Optional: Primary IPv4 (depends on qemu agent being installed)
- Optional: Primary IPv6 (same as above)
- Optional: Network interfaces

### Use Case
As a Nautobot-SSoT user, I should be able to:
- Configure Nautobot-SSoT plugin so that it connects to a Proxmox VE cluster for which I have API and administrative access.
- Execute the dedicated Proxmox VE synchronization Job to retrieve Virtual Machine information and have it saved in the Nautobot Database.
- Optional: depending on Proxmox VE API capabilities and/or already existing data, imported Virtual Machines should already be attached to their corresponding cluster and hypervisor.

FromMun:
# Closes: #876 and #1068 

## What's Changed

Adds a new one-way, read-only SSoT integration that syncs virtualization inventory from Proxmox VE into Nautobot: clusters, nodes (as Devices), QEMU VMs, LXC containers, VM/node interfaces, and IP addresses.

### Integration

  - `SSOTProxmoxConfig` model + UI (forms/filters/tables/views) for managing Proxmox VE instance config
  - `ProxmoxDataSource` job (`Proxmox VE ⟹ Nautobot`), with dry-run support and optional Cluster Filters
  - DiffSync adapters/models built on the `nautobot_ssot.contrib` framework, following the current
  reference integration (vSphere) layout
  - Signal-based bootstrap of required Tags, CustomFields, Statuses, ClusterType, and node-Device
  prerequisites on `nautobot_database_ready`
  - Objects are scoped and protected via an SSoT marker tag — untagged (manually created) objects are
  never touched; `VirtualMachine`/`VMInterface`/node `Interface` are deleted when they disappear from
  source, `Device`/`IPAddress`/`Prefix`/`Cluster`/`ClusterGroup` are preserved

### Cleanup pass
 
 - Deduplicated QEMU/LXC interface-loading logic, primary-IP selection branches, SSoT-tag-name resolution (was implemented independently in 3 places), and repeated lookup/save error handling in the Nautobot adapter
  - Centralized magic strings into `constants.py`, added type hints/docstrings throughout the adapter layer
  - Restructured test fixtures into a `proxmox_fixtures/` subpackage to match the vSphere reference layout
  - Added tests for previously uncovered branches (job auth/config error paths, default-secrets signal bootstrap, VM primary-IP/interface-link updates, `TagModel` create/update/delete, client exception propagation)

### Docs
 
 - User and admin guides, screenshots, mkdocs nav entries, README integration lists

### Constraints from Design

  - **One-way, read-only.** This integration only reads from Proxmox VE and writes into Nautobot. There is no Nautobot → Proxmox direction, and Proxmox is never modified.

  - **The SSoT tag is the ownership boundary, not just a label.** Every object this integration manages is tagged `SSoT Synced from Proxmox`, and that tag is also the *read filter* used on every subsequent run. Objects without the tag are invisible to the sync — never edited, never deleted — even if they collide on name with something Proxmox reports. This is what makes and-created Nautobot objects safe from the sync, but it also means a naming collision with an untagged object surfaces as a create-time uniqueness error rather than a silent overwrite.

  - **Deletion behavior is intentionally asymmetric.** `VirtualMachine`, `VMInterface`, and node `Interface` are deleted when they disappear from Proxmox. `Device` (nodes), `IPAddress`, `Prefix`, `Cluster`, and `ClusterGroup` are preserved even if missing from a given run. This is deliberate: a cluster-filtered run doesn't see the full inventory, and these are often shared IPAM/DCIM records — deleting them based on partial visibility would be destructive.

  - **VM IP discovery depends on the source path.** QEMU VM IPs come from the QEMU guest agent, which only reports data when the VM is powered on and the agent is installed/running. LXC container IPs come from the container config directly and don't need an agent. A powered-off or agent-less QEMU VM syncs successfully but with no interface IPs — this is expected, not a bug.

  - **Token authentication only.** No username/password login path to Proxmox VE — API tokens only.

  - **Cluster Filters scope Virtual Machines only.** Nodes, node interfaces, prefixes, and IPs are not narrowed by the job's Cluster Filter option; only which VMs get synced is affected.

  - **Unmapped VM statuses cause a skip, not a fallback.** If Proxmox reports a VM status with no corresponding entry in the configured status map, that VM is skipped entirely for the run (logged as a warning) rather than synced with a default/guessed status.

  - **Resource values are rounded.** Memory is stored in whole MB and disk in whole GB, rounded down from the bytes Proxmox reports — very small values can display as `0`.

  - **Duplicate VM names within a cluster are not merged.** Nautobot requires VM name uniqueness within a cluster; if Proxmox reports two resources with the same name in the same cluster, only the first is created and the rest are skipped with a warning.

  - **Tag management is unscoped, matching the existing vSphere pattern.** Like vSphere, this integration's `TagModel` loads/diffs against the *entire* Nautobot Tag table rather than a Proxmox-scoped subset. This is an existing cross-integration pattern (not introduced here), but worth flagging for reviewers: running this integration's sync alongside vSphere's (or any future integration with the same pattern) means each can see the other's marker tags as "not in my source" and attempt to delete them. Proxmox's marker tag is protected via a `PROTECT` FK from `SSOTProxmoxConfig.default_ssot_tag`, so that specific collision surfaces as a loud, non-silent sync error rather than a stray deletion — but the underlying unscoped-tag-table pattern is shared architecture, not something this PR introduces or fully solves.

  - **Primary IPs and interface topology resolve in a post-pass.** Node/VM primary IP assignment and intra-node interface relationships (bridge/bond/VLAN parent) are deferred to `sync_complete()`, since they require the interfaces/IPs to exist first. They're only guaranteed correct once the full job finishes, not mid-run.

  - **Errors on one object don't abort the sync.** The job runs with `CONTINUE_ON_FAILURE`; failures are visible only in the job log, not surfaced as a top-level job failure unless every object fails.

 - **Tag management can collide with the vSphere integration if both are enabled.** This integration's `TagModel` (like vSphere's) loads the entire Nautobot Tag table rather than a scoped subset — an existing pattern shared with vSphere, not something introduced here. In practice, this means a vSphere sync can see this integration's "SSoT Synced from Proxmox" tag as "not in its own source" and attempt to delete it; that delete fails loudly (the tag is protected by `SSOTProxmoxConfig.default_ssot_tag`) rather than succeeding silently. Fixing this properly requires the same scoping on vSphere's side, which is out of scope for this PR. Accepted as a known, documented limitation (see the Limitations section of the user guide) rather than fixed here.

###Screenshots
proxmox_dashboard

proxmox_detail

proxmox_jobresult

  ## To Do
  - [x] Unit, Integration Tests


FromMun:
Had to drop date + time support to just date. This opens up compatibility with Nautobot 3.0+

FromMun:
Moved integration anchor from Tag to Custom Field to combat cross integration pollution

FromMun:
Rebased with latest dev

garymccann (changes requested):
- Please remove Poetry lock file, it appears to be downgrading several packages. 
- Please remove Pylint R0801 directives from whole files
- Please consider human review of the documentation - many phrases in the documentation make no direct sense to a 3rd party, however it's the style AI coding tools use to discuss things in a familiar tone with the coder such as 
"The one thing", "is the **only** step here", "blah.. is **your** choice".

jdrew82:
@FromMun Please make the requested changes so we can get this PR merged.

PR #1291, opened 2026-07-07
Fix LibreNMS sync requiring a tenant to be selected

### LINKED ISSUE #1290: LibreNMS: Tenant Filter is effectively required - sync fails without a tenant
### Environment
* Nautobot SSoT version: 4.5.2 (develop)
* Integration: LibreNMS

### Expected Behavior
The `Tenant Filter` job variable on both LibreNMS jobs is declared `required=False` and described as an optional filter. Running a sync without selecting a tenant should work, since using tenants in Nautobot is optional.

### Observed Behavior
Three related problems:

1. **Sync crashes without a tenant.** `ensure_ip_address()` in `nautobot_ssot/integrations/librenms/diffsync/models/nautobot.py` names the IPAM Namespace after the tenant with no fallback: `Namespace.objects.get_or_create(name=adapter.job.tenant.name)`. When no tenant is selected this raises `AttributeError: 'NoneType' object has no attribute 'name'` for every device that has an IP address — which is nearly every LibreNMS device. Users are therefore forced to set a tenant.

2. **Perpetual diff noise.** `tenant` is listed in the `Device._attributes` tuple, but `NautobotDevice.update()` never applies it (it is only assigned at create time from the job's tenant). When no tenant filter is set, existing Nautobot devices that have a tenant show a diff on every sync that never resolves.

3. **Tenant filter ignored on DataTarget.** `LibrenmsDataTarget.load_source_adapter()` does not pass `tenant=self.tenant` to the `NautobotAdapter`, so the tenant filter is silently ignored in the Nautobot → LibreNMS direction.

### Steps to Reproduce
1. Configure the LibreNMS integration with an External Integration for a LibreNMS instance.
2. Run the "LibreNMS to Nautobot" job without selecting a Tenant Filter.
3. Sync fails with `AttributeError: 'NoneType' object has no attribute 'name'` when creating/updating any device that has an IP address.

alhogan:
# Closes: #1290

## What's Changed

The LibreNMS jobs declare the `Tenant Filter` as optional (`required=False`), but in practice a sync without a tenant failed, forcing users who don't use tenancy in Nautobot to select one anyway.

### Sync no longer crashes without a tenant
`ensure_ip_address()` named the IPAM Namespace after the tenant with no fallback (`adapter.job.tenant.name`), raising `AttributeError: 'NoneType' object has no attribute 'name'` for every device with an IP address. It now falls back to Nautobot's built-in `Global` namespace when no tenant is selected.

### Tenant Filter now works on the DataTarget job
`LibrenmsDataTarget.load_source_adapter()` never passed `tenant=self.tenant` to the `NautobotAdapter`, so the filter was silently ignored in the Nautobot → LibreNMS direction. It is now passed through.

### Out of scope
Tenant remains a synced attribute on the Device and Location DiffSync models, and existing tenant sync behavior is unchanged.

## To Do

- [x] Unit, Integration Tests

### DIFF
--- a/changes/1290.fixed
+++ b/changes/1290.fixed
@@ -1,3 +1,2 @@
 Fixed LibreNMS sync failing with an `AttributeError` when no Tenant Filter was selected; IP addresses are now placed in the Global namespace when no tenant is set.
-Fixed LibreNMS devices and locations reporting a tenant difference on every sync that could never be resolved; the tenant is now assigned only at creation time from the optional Tenant Filter.
 Fixed the Tenant Filter being silently ignored by the Nautobot to LibreNMS job.
--- a/changes/1290.fixed
+++ b/changes/1290.fixed
@@ -1,0 +1,3 @@
+Fixed LibreNMS sync failing with an `AttributeError` when no Tenant Filter was selected; IP addresses are now placed in the Global namespace when no tenant is set.
+Fixed LibreNMS devices and locations reporting a tenant difference on every sync that could never be resolved; the tenant is now assigned only at creation time from the optional Tenant Filter.
+Fixed the Tenant Filter being silently ignored by the Nautobot to LibreNMS job.
--- a/nautobot_ssot/integrations/librenms/diffsync/models/base.py
+++ b/nautobot_ssot/integrations/librenms/diffsync/models/base.py
@@ -11,11 +11,10 @@ class Location(DiffSyncModel):
 
     _modelname = "location"
     _identifiers = ("name",)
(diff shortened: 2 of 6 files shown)


FromMun:
Tested by me, lgtm

bile0026 (commented):
Need to have an internal discussion on intent and direction on this.

garymccann:
Code looks good. @bile0026 @alhogan - if you are happy with the direction, please let me know and i'll approve and merge.

jdrew82 (approved):

PR #1293, opened 2026-07-10
Fix: LibreNMS device location overwritten regardless of sync_locations setting

### LINKED ISSUE #1292: LibreNMS: sync_locations=False does not protect an existing device's location from being overwritten on sync
### Environment
  - Python version: 3.12
  - Nautobot version: 3.0.0
  - nautobot-ssot version: develop 

### Expected Behavior
When sync_locations is False, the LibreNMS→Nautobot sync should not modify an existing device's location field — the job's own help text describes this Var as "Whether to Sync Locations from LibreNMS to Nautobot." A common workflow is to onboard devices into a temporary catch-all Location on the first sync, then manually reparent each device to its real Location. Subsequent syncs with sync_locations=False should not silently revert that.

### Observed Behavior
NautobotDevice.update() (nautobot_ssot/integrations/librenms/diffsync/models/nautobot.py) unconditionally overwrites device.location whenever "location" in attrs — no sync_locations gate at all:

  if "location" in attrs:
      ...
      device.location = _location

  location/parent_location are permanent Device._attributes (diffsync/models/base.py), 

so they're diffed on every device, every run, regardless of sync_locations. That flag only gates whether standalone Location objects are loaded/diffed (NautobotAdapter.load()/LibrenmsAdapter.load()) — it has no effect on the per-device location attribute. The system_of_record/SKIP_UNMATCHED_DST pattern doesn't help either — that only suppresses delete-diffs for Nautobot-only devices, not updates to matched devices.

### Steps to Reproduce
  1. Run the sync once with a location_map routing a device to a catch-all Location (or no location_map at all).
  2. Manually change that device's Location in Nautobot to its real Location.
  3. Re-run the sync with sync_locations=False.
  4. Observe the Location reverts back, despite sync_locations being disabled.

FromMun:
# Closes: #1292 

## What's Changed
     
  Fixes #1292. NautobotDevice.update() (nautobot_ssot/integrations/librenms/diffsync/models/nautobot.py) was overwriting a device's location on every sync regardless of the job's sync_locations setting.  location/parent_location are permanent Device._attributes (diffsync/models/base.py), so they're diffed for every device on every run; update() applied the LibreNMS-computed location whenever "location" was present in the diffed attrs, with no gate at all. sync_locations only ever gated the separate top-level Location-object sync, never this per-device attribute — so disabling it didn't protect device placement the way its help text implies, and any device manually reparented in Nautobot after onboarding got silently reverted on the next sync.

##  Fix

  One-line guard in NautobotDevice.update():
  -if "location" in attrs:
  +if "location" in attrs and self.adapter.job.sync_locations:

  NautobotDevice.create() is intentionally untouched: a new device always needs some location assigned (Device.location isn't nullable), so create-time assignment must keep working regardless of sync_locations.

##  Before/After

  - Before: onboard into a catch-all Location → manually reparent to the real Location → re-run sync with
  sync_locations=False → device is silently moved back.
  - After: same steps, but the Location is left alone (the diff still shows a cosmetic location mismatch in the job's change summary, but nothing is written).

##  To Do

  - [x] Added change log fragment(s) — changes/1292.fixed
  - [x] Unit, Integration Tests — new tests in test_nautobot_model.py + test_utils.py; full suite green
  (1518 tests)
  - [x] Outline Remaining Work, Constraints from Design — none, self-contained fix

### DIFF
--- a/changes/1292.fixed
+++ b/changes/1292.fixed
@@ -1,0 +1,1 @@
+Fixed LibreNMS device sync overwriting a device's location on every run regardless of the sync_locations setting.
--- a/nautobot_ssot/integrations/librenms/diffsync/models/nautobot.py
+++ b/nautobot_ssot/integrations/librenms/diffsync/models/nautobot.py
@@ -356,7 +356,7 @@ class NautobotDevice(Device):
             device.status = Status.objects.get_or_create(name=attrs["status"])[0]
         if "role" in attrs:
             device.role = ensure_role(role_name=attrs["role"], content_type=ORMDevice)
-        if "location" in attrs:
+        if "location" in attrs and self.adapter.job.sync_locations:
             # Get location data from the device attributes
             location_name = attrs["location"]
             parent_location_name = attrs.get("parent_location")
--- a/nautobot_ssot/tests/librenms/test_nautobot_model.py
+++ b/nautobot_ssot/tests/librenms/test_nautobot_model.py
@@ -1,0 +1,107 @@
+"""Unit tests for the Nautobot-side LibreNMS DiffSync models (NautobotDevice)."""
+
+from unittest.mock import MagicMock
+
+from diffsync import Adapter
+from django.contrib.contenttypes.models import ContentType
+from nautobot.apps.testing import TestCase
+from nautobot.dcim.models import Device as ORMDevice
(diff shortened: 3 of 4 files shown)


jdrew82 (approved):

bile0026 (approved):
LGTM!

PR #1297, opened 2026-07-14
Speed up Infoblox by bulk-fetching records and paging large prefixes

### LINKED ISSUE #1296: Infoblox sync is slow and fails on large data sets.
### Environment
* Python version:  3.12
* Nautobot version:  3.1.6

The Infoblox source adapter has two scaling problems that make sync slow on medium data sets and cause it to fail outright on large ones.

1. N+1 WAPI requests when loading IP addresses. For every IP address, the adapter issues a separate WAPI GET for each associated DNS (A/host/PTR) and fixed-address record (get_a_record_by_ref, get_host_record_by_ref, get_ptr_record_by_ref, get_fixed_address_by_ref). This turns IP loading into thousands of sequential round-trips that scale linearly with the number of IPs.

2. Unpaged fetch for large prefixes. For any prefix larger than 1,000 hosts, the adapter requests every address in a single call (_max_results set to the full prefix size — e.g. 65k for a /16). On large grids this single unbounded request hangs, exceeds the read timeout, or is truncated mid-stream by the Grid.

Impact: On a 2,075-prefix / 3,241-IP test grid, loading takes 6m32s. On the production grid (9,473 prefixes) the job runs 12+ hours and then fails — it never completes.

nopg:
# Closes: #1296 

## What's Changed
Fix Infoblox sync performance: eliminate N+1 API calls and page large-prefix fetches

- Eliminate the N+1 pattern. Bulk-fetch all A, host, PTR, and fixed-address records once up front via a new get_all_records_by_type, keyed by _ref, then enrich each IP with an in-memory dict lookup. Removes the per-IP get_*_by_ref calls. Round-trips during IP loading go from O(number of records) to a fixed handful of paged bulk calls.
- Page large-prefix IP fetches. Replace the single unbounded request for prefixes >1,000 hosts with get_ipaddrs_paged, which walks bounded 1,000-result pages (mirroring get_all_subnets). No single response is large enough to time out or be truncated.
- Removes the now-unused get_*_by_ref client methods and their fixtures; adds get_all_records_by_type coverage.
- New tests for all of the above, reusing and combining the fixtures from the now removed methods.

**Measured result** (2,075 prefixes / 3,241 IPs, same data):

| | develop | this PR | Improvement |
|---|---|---|---|
| Load time | 6m32s | **1m57s** | 3.35× faster (−70%) |

Expected impact on large grids. The two fixes address different bottlenecks. The N+1 removal gives the ~3× improvement measured above and grows with IP count. The paging fix targets the failure mode on large grids: on a much larger grid (9,473 prefixes) the current code runs 12+ hours and fails. With bounded paging the sync is expected to complete in the low-tens-of-minutes range.


garymccann (approved):
LGTM. Nice work.

PR #1300, opened 2026-07-14
Fix Infoblox location EA mapping: case-insensitive, LocationType-safe, and configurable.

### LINKED ISSUE #1299: Infoblox: Sync crashes on LocationType missing Prefix ContentType, and hardcoded to site/facility/location.
### Environment
* Python version:  3.12
* Nautobot version:  3.1.6

### Expected Behavior
If the LocationType in use is missing the Prefix ContentType, the Job should not crash but instead log a warning.

### Observed Behavior
Job fails with:
```
{
    "exc_message": [
        {
            "locations": "Prefixes may not associate to Locations of types ['Region']."
        },
        null,
        null
    ],
    "exc_module": "django.core.exceptions",
    "exc_type": "ValidationError"
}
```

This is further exacerbated by the fact that the Location Infoblox is looking for is based off an Extensibility Attribute that we do not use. (site/facility or location). It would be fantastic if we could tell the SSOT which attribute actually maps to our Nautobot Locations.

### Steps to Reproduce
1. Attempt to sync Infoblox to a Location/LocationType that does not allow Prefix ContentTypes

nopg:
# Closes: #1299

## What's Changed

- Match location EAs case-insensitively (name and value).
- Skip with a warning (instead of crashing) when the matched Location's LocationType can't contain the object.
- Add a `infoblox_location_ext_attr` config option to map a custom-named EA to Location; when set, it replaces the built-in `site/facility/location` names.
- Includes a migration for the new configuration setting, UI/form/template updates, docs, and unit tests.


garymccann (approved):
LGTM. Not the intent of this PR, but please consider removing the model code that is dependent on Nautobot <2.2.0, since the pre-req is now > 3.x

jdrew82 (approved):
LGTM!

PR #1302, opened 2026-07-14
Add grouping and descriptions to Infoblox-managed custom fields

### LINKED ISSUE #1301: Infoblox managed custom fields create too much clutter
### Environment
* Nautobot version:  3.1.6

### Proposed Functionality
Infoblox creates many CustomField objects, and even more if I am using Extensibility Attributes.
This is fine, however when looking at the CustomFields in Nautobot it gets extremely cluttered and I'm unable to determine if they were created manually, by one of many Jobs/Integrations, or the Infoblox SSOT. I would like these 'auto created' CustomFields to more clearly show that they came from the Infoblox SSOT.

### Use Case
Normal Infoblox SSOT usage, just an enhancement request so that I can verify they came from Infoblox easily.

nopg:
# Closes: #1301 

## What's Changed

Adds a new description and grouping to every custom field the integration creates. 'Managed by' for reserved custom fields and 'Auto-created' for custom fields created due to Infoblox extensible attributes.


garymccann (changes requested):
This PR fixes *new* installs of the Infoblox integration, but will leave existing users who are upgrading without this fix. get_or_create will silently miss the already created objects. Is that by design? 

My suggestion would be to perform a migration also. 
See https://github.com/nautobot/nautobot-app-ssot/blob/develop/nautobot_ssot/migrations/0007_replace_dashed_custom_fields.py for an example.

nopg:
> This PR fixes _new_ installs of the Infoblox integration, but will leave existing users who are upgrading without this fix. get_or_create will silently miss the already created objects. Is that by design?
> 
> My suggestion would be to perform a migration also. See https://github.com/nautobot/nautobot-app-ssot/blob/develop/nautobot_ssot/migrations/0007_replace_dashed_custom_fields.py for an example.

It was technically by design as I had seen issues using `update_or_create` in the past for this type of setting. You're example using a migration sounds nice though, I've added it to this PR.

Of note, the integration dynamically creates other CustomFields based on whatever Extensibility Attributes the local Infoblox has, these (existing, not net new) cannot really safely be determined and/or updated, so I just added a comment noting that in the migration file.

nopg:
I could be wrong but I believe the CI errors are the same unrelated issue that my other recent PR's hit but still got merged.

jdrew82 (approved):

PR #1315, opened 2026-07-28
Fix failing unittests

Defiantearth:
## What's Changed
- Fixed Arista CloudVision SSoT failing to load on Python 3.12 without `setuptools` installed.
- Fixed DNA Center sync importing no Devices on Nautobot v3.2.

- Fixed failing Device42 unit test caused by changed Interface ordering in Nautobot v3.2.

## To Do

- [ ] Added change log fragment(s) (for more information see [the documentation](https://docs.nautobot.com/projects/core/en/stable/development/core/#creating-changelog-fragments))
- [ ] Unit, Integration Tests


jtdub (approved):
LGTM!

nautobot_ssot/integrations/dna_center/diffsync/models/nautobot.py
```diff
@@ -480,7 +480,7 @@ def create(cls, adapter, ids, attrs):
         """Create IPAddress in Nautobot from IPAddress object."""
         new_ip = IPAddress(
             address=f"{ids['host']}/{attrs['mask_length']}",
-            namespace=adapter.namespace_map[ids["namespace"]],
+            namespace=Namespace.objects.get(id=adapter.namespace_map[ids["namespace"]]),
```
jdrew82, line 483: Why was this done? We typically want to keep use of maps to enable the best performance during CRUD ops.

nautobot_ssot/tests/device42/unit/test_models_nautobot_ipam.py
```diff
@@ -364,9 +364,7 @@ def test_update_changing_interface(self):
         update_attrs = {"interface": "mgmt0"}
         result = self.mock_addr.update(attrs=update_attrs)
         self.assertIsInstance(result, ipam.NautobotIPAddress)
-        self.addr.refresh_from_db()
-        self.dev2_mgmt.refresh_from_db()
-        self.assertEqual(self.addr.interfaces.first(), self.dev2_mgmt)
+        self.assertEqual(self.addr.interfaces.get(device=self.test_dev2), self.dev2_mgmt)
```
jdrew82, line 367: Also, why is this being done? This seems to just be rewriting the assertion?

jdrew82 (commented):
Few questions.

nautobot_ssot/integrations/dna_center/diffsync/models/nautobot.py
Defiantearth: namespace_map still supplies the PK, but namespace isn't a field on the ORM IPAddress (_meta.get_field("namespace") raises FieldDoesNotExist), so it needs the instance.

nautobot_ssot/tests/device42/unit/test_models_nautobot_ipam.py
glennmatthews: Interface default ordering changed in 3.2, so `.first()` may not necessarily be the correct interface any more?

nautobot_ssot/integrations/dna_center/diffsync/models/nautobot.py
glennmatthews: What about 3.2 made this a new problem?

nautobot_ssot/tests/device42/unit/test_models_nautobot_ipam.py
Defiantearth: It was using `.first()` to get the interface which is not random since core changed the way it is ordered so we are getting the specific interface so it is not random.

nautobot_ssot/integrations/dna_center/diffsync/models/nautobot.py
Defiantearth: @glennmatthews I'm pretty sure it was this PR https://github.com/nautobot/nautobot/pull/9169/changes

Defiantearth: I think it's specifically this line trying to access the pk. So if you pass a pk it will error out since it is a namespace instance. https://github.com/nautobot/nautobot/pull/9169/changes#diff-15c6afc90d7ab9ffc408a2b7c6857caf4c3ebb9bc1213aa5340cbc301f85e7c0R124

glennmatthews: Seems likely. So sounds like we could use a fix in core to recognize that `IPAddress(namespace=<uuid>)` needs to be implicitly converted to `IPAddress(namespace_id=<uuid>)`?

Defiantearth: Yep!

glennmatthews: Would it work as well if instead of doing `namespace=Namespace.objects.get()` you changed it to `namespace_id=adapter.namespace_map[ids["namespace"]]`?

Defiantearth: No it won't since namespace is not a field on the model. status_id works since status is a foreign key on the model.

glennmatthews: Ah okay. I'm fine with this as a short-term workaround for the core issue but let's make sure it's tracked as something to address here again once core fixes it.

glennmatthews (approved):

jdrew82 (approved):

PR #1322, opened 2026-08-05
bump meraki to ^3

### LINKED ISSUE #1321: Meraki is locked on a very old version
### Environment
* Python version:  
* Nautobot version:  
* nautobot-ssot version:  

"meraki" dependency is stuck on v1 while the latest version on pypi is 4.3.1

icovada:
# Closes: #1321

## What's Changed

Meraki updated to v3.

Cannot update to v4 as there's a transient conflict about httpx through slurpit-sdk

## To Do

- [X] Unit, Integration Tests


pyproject.toml
```diff
@@ -28,7 +28,7 @@ include = [
 ]
 
 [tool.poetry.dependencies]
-python = ">=3.10,<3.15"
+python = ">=3.11,<3.15"
```
glennmatthews: Dropping support for 3.10 is probably not desirable yet.

pyproject.toml
```diff
@@ -56,7 +56,7 @@ httpx = { version = ">=0.23.3", optional = true }
 cisco-sdwan = { version = "^1.26", optional = true }
 # used for DNA Center integration
 dnacentersdk = { version = "^2.5.6", optional = true }
-meraki = { version = "^1.37.2,<1.46.0", optional = true }
+meraki = { version = "^3", optional = true }
```
glennmatthews: Any code changes required to move forward here? No breaking changes at all?

pyproject.toml
joewesch: It looks like `meraki` dropped Python 3.10 support as part of this release: https://github.com/meraki/dashboard-api-python/releases/tag/3.0.0. This could be a blocker, we should reassess.

joewesch:
I have reverted the change to drop Python 3.10 and changed `meraki` to a split constraint now so users on Python 3.11+ will get v3. I have also verified there are no breaking changes. CI should test both versions since we test both 3.10 and 3.14 in our unittest matrix.

pyproject.toml
```diff
@@ -56,7 +56,11 @@ httpx = { version = ">=0.23.3", optional = true }
 cisco-sdwan = { version = "^1.26", optional = true }
 # used for DNA Center integration
 dnacentersdk = { version = "^2.5.6", optional = true }
-meraki = { version = "^1.37.2,<1.46.0", optional = true }
+# meraki 3.x requires Python >=3.11. v3.0.0 and v3.0.1 include some regression bugs.
+meraki = [
+    {version = ">=1.37.2,<1.46.0", python = ">=3.10,<3.11", optional = true},
```
glennmatthews, line 61: possible to move this one to v2.x at least? Or was there no v2?

pyproject.toml
icovada: All tests pass

pyproject.toml
joewesch: When `meraki` [started using poetry](https://github.com/meraki/dashboard-api-python/commit/d220b5f1c825dafd8558310d6813fe193b29e3f8) in their environment in v1.56.0, they added their dev dependencies (pytest, build, wheel, setuptools) into their runtime dependencies instead of handling them as dev only; v3 reverts this. I skipped v1.56.0+ (including v2.X) because of this.

There are no meaningful changes beyond v1.46.0 that we use, so I decided to keep the same upper bound for Python 3.10 users.

glennmatthews: Fun! Thanks for the information.

glennmatthews (approved):

jdrew82:
## Notes: `meraki` SDK version differences and what they require

Following up on @glennmatthews' question ("Any code changes required to move forward here? No breaking changes at all?") with the detail behind it. Short answer: there **is** one breaking change, but it lands in **2.0.0**, not 3.0.0 — which is easy to miss when the diff is framed as a v1 → v3 bump.

Method: I pulled the sdists for 1.45.0, 1.56.0, 2.0.3, 2.2.0, 3.0.0, 3.3.0 and 4.4.0, extracted the SDK call surface from `utils/meraki.py`, and bound every call against a live 3.3.0 install.

### Version landscape

| Version | Released | `Requires-Python` | Transport |
| ------- | -------- | ----------------- | --------- |
| 1.45.0 | 2024-04 | `>=3.8` | requests / aiohttp |
| 1.56.0 | 2025-03 | `>=3.10` | requests / aiohttp |
| 2.2.0 | 2026-03 | `>=3.10` | requests / aiohttp |
| 3.3.0 | 2026-07 | `>=3.11` | requests / aiohttp |
| 4.4.0 | 2026-08 | `>=3.11` | **httpx** |

### What is NOT affected

All 14 endpoints the integration calls are present with **byte-identical signatures** in every version from 1.45 through 4.4:

`getOrganizations`, `getOrganizationNetworks`, `getOrganizationDevices`, `getOrganizationUplinksStatuses`, `getOrganizationDevicesUplinksAddressesByDevice`, `getOrganizationDevicesStatuses`, `getOrganizationSwitchPortsBySwitch`, `getDeviceSwitchPortsStatuses`, `getDeviceManagementInterface`, `getDeviceApplianceUplinksSettings`, `getNetworkAppliancePorts`, `getNetworkApplianceVlans`, `getNetworkApplianceSingleLan`, `getNetworkApplianceVlansSettings`

Also unchanged: the `DashboardAPI.__init__` kwargs we pass; `APIError.message` / `.status` / `.reason`; and `USE_ITERATOR_FOR_GET_PAGES` still defaults to `False`, so `total_pages="all"` still returns a list rather than a generator.

The 2.0.0 release note about "changes in operation IDs that are breaking for the library" sounds alarming but only touches `*InventoryOnboardingCloudMonitoring*` and `createOrganizationApplianceDnsLocalProfilesAssignmentsBulkCreate`. We call none of them.

### The actual breaking change (introduced in 2.0.0)

`SessionInputError` was added in 2.0.0. It is **not** a subclass of `APIError` and is **not** exported at the `meraki` top level. Every handler in `utils/meraki.py` is `except meraki.APIError`, so it escapes as an unhandled traceback. Two live trigger paths:

1. **User-agent validation.** `validate_user_agent()` now regex-validates `caller` / `be_geo_id`, including values read from the `MERAKI_PYTHON_SDK_CALLER` and `BE_GEO_ID` environment variables. 1.x accepted any string with no validation. Confirmed against 3.3.0: `BE_GEO_ID=some_partner_id` raises during `DashboardAPI()` construction — inside `connect_dashboard()`'s `try`, sailing straight past the `except meraki.APIError`. Anyone with `BE_GEO_ID` set for partner tracking gets an unhandled traceback on job start.
2. **`total_pages` validation.** `_get_pages_legacy` now raises if `total_pages` is neither an `int` nor the literal string `"all"`. `api_total_pages` is read straight out of `PLUGINS_CONFIG` with no validation.

Under the current split constraint both become **Python-version-dependent**: identical config works on 3.10 and fails on 3.11+. That's a rough failure mode to debug from a job log.

One wrinkle on the fix — `SessionInputError` does not exist in 1.x, so a plain import breaks the 3.10 leg. It needs guarding:

```python
try:
    from meraki.exceptions import SessionInputError
except ImportError:  # meraki < 2.0
    class SessionInputError(Exception):
        """Placeholder so handlers stay uniform across SDK majors."""
```

Note also that `SessionInputError` has no `.status` / `.reason` / `.message`, so the shared logging f-string in the existing handlers would itself raise `AttributeError` while trying to log the error. The 13 duplicated handler blocks need a helper that tolerates both shapes.

### Behavioral differences worth documenting

- `getOrganizationDevices` raised its `perPage` ceiling from **1000 to 5000** in 3.x. `api_page_size: 5000` therefore works on 3.11+ and is rejected on 3.10. Users running mixed Python versions across environments should stay at 1000 or below.
- `getOrganizationSwitchPortsBySwitch` remains capped at 50 per page throughout — unchanged, but worth knowing.
- 3.0.0 fixed the sync `_get_pages_iterator` dropping the last page. We don't hit this (the iterator is off by default), but it's the reason `>=3.0.2` is a sensible floor.

### Suggestion on the split constraint

The real behavioral seam is **v1 → v2**, so the split as written (`1.x` / `3.x`) leaves the two legs differing in exactly the ways that matter, while CI can only meaningfully test one of them.

`meraki` 2.2.0 (March 2026) still declares `Requires-Python: >=3.10`, and I verified all 14 endpoints are intact there. So:

```toml
meraki = [
    {version = ">=2.2.0,<3", python = ">=3.10,<3.11", optional = true},
    {version = ">=3.0.2,<4", python = ">=3.11", optional = true},
]
```

Both legs then share identical validation behavior, and 3.10 users move off a 2024 SDK — which is what #1321 was actually asking for.

### On CI coverage

The 3.14 matrix leg does install 3.3.0, and the unpatched `DashboardClient(...)` constructions in `test_utils_meraki.py` do build a real `DashboardAPI`, so the constructor path is genuinely exercised. Everything past that is `MagicMock` — no test touches `get_pages`, pagination, or any `except meraki.APIError` branch. A green 3.14 run tells us the import works, not that there are no breaking changes.

### On v4

@icovada's reasoning checks out and is worth recording explicitly: `slurpit-sdk==0.9.107` pins `httpx<0.28.0,>=0.27.0` while `meraki` 4.x requires `httpx>=0.28,<1` — genuinely mutually exclusive. Good news is that 4.4.0 keeps all 14 endpoints and preserves the `APIError` attribute surface (`.reason` is remapped from httpx's `reason_phrase`), so once the slurpit pin relaxes, v3 → v4 should be a small step.

### Also worth folding in

Setting `caller` explicitly is a free win — Cisco's [User-Agent guidance](https://developer.cisco.com/meraki/api-v1/user-agents-overview/) treats it as expected practice, and a valid `caller` takes precedence in `validate_user_agent()`, which also inoculates us against the `BE_GEO_ID` trigger above. `"NautobotSSoT/4.6.2 NetworkToCode"` passes the 3.x pattern; note that `nautobot-ssot` is rejected (hyphens fail it), as is `Network To Code` (spaces).

Separately, `connect_dashboard()` logs via `self.logger.log.error(...)` while every other method uses `self.logger.logger.warning(...)`. Since `jobs.py` passes `logger=self` (the Job) and `Job.log` doesn't exist on Nautobot 3.x, the one handler that fires on a connection failure raises `AttributeError` and masks the real error. Pre-existing, but this PR expands what can fail at connection time, so it'd be good to fix here.

I have a patch implementing all of the above (guarded import, widened handlers, `caller`, the logger fix, tests, and docs) — happy to push it to the branch or open it as a follow-up, whichever you'd prefer. Worth noting the code change is safe under 1.x, 2.x and 3.x alike, so it's independently mergeable from the constraint change if you'd rather not move the 3.10 leg in this PR.

joewesch:
@jdrew82 I have fixed the unhandled `SessionInputError` as found as well as set the User Agent to be compliant, thanks.

As for changing the version to include meraki v2, I have already addressed why I have skipped all v2 here: https://github.com/nautobot/nautobot-app-ssot/pull/1322#discussion_r3844128911.

jdrew82 (approved):

PR #1336, opened 2026-08-13
Fix LibreNMS platform and manufacturer resolution, add opt-in platform consolidation

### LINKED ISSUE #1335: LibreNMS: onboarded devices don't match the standard set by device-onboarding, and the sync has related defects
### Environment
* Python version: 3.12.3
* Nautobot version: 3.1.0
* nautobot-ssot version: 4.6.1

### Expected Behavior
A device onboarded from LibreNMS should end up with the same Platform and Manufacturer that
nautobot-app-device-onboarding would give it, so the two apps share objects instead of duplicating
them:

* `Platform.network_driver` should be a netutils driver name (`cisco_ios`), so
  `Platform.network_driver_mappings` populates and Golden Config / the nornir-nautobot dispatcher can
  resolve a driver.
* `Manufacturer` should be the vendor (`Palo Alto`), not the LibreNMS OS string.
* A device that is reclassified should end up fully consistent — Platform *and* the objects that hang
  off it.
* One unsaveable device should be skipped with an error, not abort the run.

### Observed Behavior
Five defects. **(1)** and **(2)** are the substantive ones and share a single cause; **(3)** and
**(4)** are what that causes downstream once the data is corrected; **(5)** is unrelated but too
small to be worth its own issue.

- [ ] 1. `Platform.network_driver` is set to an Ansible collection FQCN
- [ ] 2. `Manufacturer` is named after the device OS (210 of 217 mapped OS values)
- [ ] 3. One device failing validation aborts the entire sync
- [ ] 4. A device that changes Platform keeps its old Platform's SoftwareVersion
- [ ] 5. `KeyError` on `librenms_allow_ip_hostnames` when the setting is absent

---

#### 1. `Platform.network_driver` is set to an Ansible collection FQCN
Platforms are created with the Ansible collection name as the network driver — `cisco.ios.ios` rather
than `cisco_ios`. That is not a netutils driver name, so `network_driver_mappings` comes back empty
and anything dispatching on it (Golden Config, the nornir-nautobot dispatcher) has nothing to work
with.

| LibreNMS `os` | Platform name | `network_driver` written | netmiko mapping |
| `ios` | `cisco.ios.ios` | `cisco.ios.ios` | *(none)* |
| `iosxe` | `cisco.ios.ios` | `cisco.ios.ios` | *(none)* |
| `iosxr` | `cisco.iosxr.iosxr` | `cisco.iosxr.iosxr` | *(none)* |
| `nxos` | `cisco.nxos.nxos` | `cisco.nxos.nxos` | *(none)* |

#### 2. `Manufacturer` is named after the device OS
Devices are filed under a Manufacturer named after their operating system instead of their vendor.
This hits every OS that is not one of the seven the integration maps explicitly: **210 of the 217
values in `os_manufacturer_map`**.

| LibreNMS `os` | Manufacturer created | Expected |
| `panos` | `panos` | `Palo Alto` |
| `asa` | `asa` | `Cisco` |
| `fortios` | `fortios` | `Fortinet` |

The `asa` row is worth calling out: a `Cisco` Manufacturer already exists — the `ios` devices in the
same sync created it correctly — yet the ASAs land under a second, OS-named vendor. Every vendor with
more than one OS in LibreNMS gets fragmented this way, and the resulting DeviceTypes are split across
Manufacturers that should have been one.

Note the inversion between (1) and (2): the Platforms that get the FQCN driver are exactly the ones
that get the *correct* Manufacturer, and vice versa. Both come from the same design — the driver and
the vendor are re-derived from the Platform *name* after the fact, rather than from the LibreNMS `os`
the sync started with. The adapter already resolves the correct vendor from the OS and passes it
along; the device creation path discards it and recomputes.

Neither object is repaired once it exists, so every newly created Platform and Manufacturer is created
wrong and stays that way.

#### 3. One device failing validation aborts the entire sync
A single device that cannot be saved terminates the whole job, so an inventory of several hundred
devices imports nothing because of one bad record. Devices that fail while being *loaded* from
LibreNMS are already handled correctly — logged, collected as failures, and skipped — but a failure on
save is not, and escapes the sync.

The realistic trigger is defect (2). Nautobot requires a Device's Platform manufacturer to match its
DeviceType manufacturer, so an OS-named Manufacturer colliding with a real vendor already created by
device-onboarding fails validation and takes the run down with it.

#### 4. A device that changes Platform keeps its old Platform's SoftwareVersion
`SoftwareVersion` is scoped by Platform, but the sync only re-points a device's software version when
the reported *version string* changes. A device whose platform is corrected while its version stays
the same keeps a SoftwareVersion belonging to a platform it no longer runs.

This is reachable on its own — an `ios` → `iosxe` reclassification in LibreNMS is enough — and it hits
every device at once during any migration that corrects the Platform names from (1).

#### 5. `KeyError` on `librenms_allow_ip_hostnames` when the setting is absent
The integration reads `librenms_allow_ip_hostnames` as a required key, but it is not among the app's
default settings, so an install that enables the integration without defining it raises
`KeyError: 'librenms_allow_ip_hostnames'`.

It is latent rather than immediate: the check is only reached for a device whose hostname is an IP
address, so the job runs fine until it meets one and then dies. That makes it look like a data problem
with that one device rather than a missing default.

`librenms_permitted_values` and `librenms_show_failures` are read safely and do not raise, but they are
also absent from the defaults; adding all three would put the integration's defaults in one
discoverable place.

### Steps to Reproduce
**(1) and (2)**

1. Configure the LibreNMS integration against an instance with at least one device with `os: ios`, one
   with `os: panos` and one with `os: asa`, and run the **LibreNMS SSoT Data Source** job with device
   sync enabled.
2. Open the Platform named `cisco.ios.ios` in Nautobot — its network driver is `cisco.ios.ios` and it
   resolves to no driver mappings, where `cisco_ios` was expected.
3. Open the Manufacturers list — there is a `panos` and an `asa` alongside `Cisco`, where `Palo Alto`
   and `Cisco` were expected.

**(3)**

1. Create a Platform in Nautobot whose name matches one the sync will produce, but with a Manufacturer
different from the one LibreNMS resolves for those devices — what device-onboarding leaves behind in a mixed environment.
2. Run the sync against an inventory containing that device plus a number of unaffected devices.
3. The job fails with a `ValidationError` on the first affected device, and no device later in the
   inventory is created. Expected: that device is logged and skipped, the rest sync normally.

**(4)**

1. Sync a device whose `os` is `ios` and `version` is `16.9.3`. Nautobot creates Platform
   `cisco.ios.ios` and a SoftwareVersion `16.9.3` under it.
2. Change the device's `os` in LibreNMS to `iosxe`, leave `version` unchanged, and re-run the job.
3. The device's platform is updated, but its software version still belongs to the old
   `cisco.ios.ios` platform.

**(5)**

1. Set `"enable_librenms": True` in `nautobot_config.py` and do **not** define
   `librenms_allow_ip_hostnames`.
2. Run the sync against an inventory containing at least one device whose hostname is an IP address.
3. The job fails with `KeyError: 'librenms_allow_ip_hostnames'`. Expected: the setting defaults to
   `False` and the device is reported as failing hostname validation.

FromMun:
# Closes: #1335 

## What's Changed
The LibreNMS integration wrote an Ansible collection FQCN such as `cisco.ios.ios` into
`Platform.network_driver`, where it resolved to no driver mappings and left Golden Config and Nornir
dispatch with nothing, and it named Manufacturers after the device OS (`panos` rather than
`Palo Alto`). Both came from the same design: the driver and the
vendor were re-derived from the Platform *name* after the fact, rather than from the LibreNMS `os` the
sync started with.

This fixes the resolution for newly created objects, adds an opt-in mode that names Platforms after
the network driver so they are shared with nautobot-app-device-onboarding, and ships a job that
migrates the rows an earlier sync already created. **No existing Platform or Manufacturer is modified
automatically** — every repair is opt-in and defaults to a dry run.

### Fixes
- Newly created Platforms get a valid `network_driver` in both platform-naming modes.
- Manufacturers resolve to the vendor instead of the OS string. The adapter already computed the
  correct vendor and passed it along; the device creation path was discarding it.
- A device sync no longer aborts entirely when a single device fails validation — the device is
  skipped with an error logged. This was reachable whenever a Platform another app created collided
  on manufacturer.
- A device that moves to a different Platform no longer keeps a SoftwareVersion belonging to its old
  Platform.
- `librenms_allow_ip_hostnames`, `librenms_show_failures` and `librenms_permitted_values` are now in
  the app's default settings, so an install that does not set them no longer raises `KeyError`.

### New settings
- `librenms_consolidated_platforms` (default `false`) — name Platforms after the network driver
  (`cisco_ios`) instead of the Ansible FQCN (`cisco.ios.ios`), so they are shared with
  device-onboarding rather than duplicated. Consolidated mode resolves platform identity on read and
  understands the legacy shapes, so enabling it adopts rows you already hold rather than forking new
  ones. The only device movement it causes is the IOS/IOS-XE split.
- `librenms_network_driver_map` (default `{}`) — override or suppress the bundled OS-to-driver
  mappings. `{"iosxe": "cisco_ios"}` declines the IOS-XE split; `{"fortios": ""}` suppresses a
  mapping; `{"dnos": "dell_os10"}` resolves one of the deliberately unmapped ambiguous values.

### New job: LibreNMS Platform Consolidation
Migrates rows created before the fix. Opt-in, defaults to a dry run, scoped by default to
LibreNMS-synced platforms. It never invents a driver, never overwrites a `network_driver` that
disagrees, and never touches anything outside its scope. Four independently selectable phases:

- **Repair drivers** (default on) — set the correct driver on FQCN-named rows whose driver is blank or
  duplicates the name. Changes no names and moves no devices, so it is safe in either mode and fixes
  Golden Config immediately.
- **Rename legacy platforms** (default off) — `cisco.ios.ios` → `cisco_ios`, preserving the primary
  key so foreign keys, config context assignments, notes, metadata, relationships and dynamic groups
  survive.
- **Merge duplicates** (default off) — collapse Platforms sharing a driver onto one survivor, moving
  devices, virtual machines, controllers, software versions, config context links, notes and metadata.
- **Repair manufacturers** (default off) — fold OS-named Manufacturers back onto the real vendor.
  Renames in place when the vendor name is free, which moves nothing; merges when it is taken.

The rename and merge phases refuse to run while `librenms_consolidated_platforms` is `false`, because
in legacy mode the sync would simply re-create the row.

The run records its plan as a **Data Sync**, so it appears in SSoT Sync History with the same
paginated diff view the sync jobs produce, renders the plan as a table in the job log, and attaches
CSVs for estates too large to read on screen. It is a plain `Job` rather than a `DataSource` /
`DataTarget` — it remediates Nautobot data in place rather than syncing two systems, so
`get_data_jobs()` keeps it off the SSoT dashboard.

### Also
- Network driver mappings for around forty more LibreNMS OS values. Genuinely ambiguous values
  (`dnos`, `asyncos`, `junose`, `cumulus`, `opnsense`, `vmwareesxi`, `aix` and others) are left
  unmapped rather than guessed, and are resolvable via `librenms_network_driver_map`.
- Documentation covering both modes, the adoption table, the IOS/IOS-XE split, and a six-step
  migration sequence that is independently reversible up to the point of enabling consolidated mode.

## To Do
- [x] Unit, Integration Tests

### Tests
117 tests added across five files: platform consolidation (55), driver resolution utils (25),
Nautobot models (19), Nautobot adapter (12) and LibreNMS adapter (6). Coverage includes the adoption
matrix, refusal paths for every destructive phase, and the primary-key preservation guarantee.

### Remaining work and constraints
- Existing Platforms and Manufacturers are deliberately never modified by a sync. Repairing them is
  the consolidation job's job, and every phase of it is opt-in.
- Ambiguous OS values stay unmapped by design rather than being guessed.
- The merge phase refuses on a software version collision or a manufacturer conflict unless you
  explicitly choose how to resolve it, and never deletes a Platform that still has software versions.
- Dynamic group filters are rewritten during a rename only when you opt in; `ObjectPermission`
  constraints referencing a Platform name cause a refusal rather than a silent break.


jdrew82:
@FromMun Please fix the conflict that's occurring with the test_nautobot_model file.

FromMun:
@jdrew82 Updated

docs/user/integrations/librenms.md
```diff
@@ -21,10 +23,164 @@ PLUGINS_CONFIG = {
...
+
+| | `false` (default) | `true` |
+|---|---|---|
+| New Platform name | `cisco.ios.ios` | `cisco_ios` |
+| `network_driver` on new Platforms | `cisco_ios` | `cisco_ios` |
+| Existing Platforms | never modified | never modified; adopted in place |
+| Identity used for diffing | Platform name | network driver |
+| Devices moved when enabling | none | only the IOS/IOS-XE split |
+
+In both modes a newly created Platform now gets a valid `network_driver`. Previously the Ansible
```
jdrew82: This needs to be rephrased to remove the history. History should be in git and the docs have just the integration details.

jdrew82 (changes requested):
Code looks good but the docs need to be tweaks a bit.

nautobot_ssot/integrations/librenms/constants.py
```diff
@@ -406,3 +406,122 @@
     "juniper_junos": "junos",
     "applogic_procera": "procera",
 }
+
+# LibreNMS `os` | network driver
```
jdrew82, line 415: This should be in netutils.

docs/user/integrations/librenms.md
FromMun: Updated

nautobot_ssot/integrations/librenms/constants.py
FromMun: Updated

nautobot_ssot/integrations/librenms/constants.py
```diff
@@ -406,3 +411,94 @@
...
+#
```
jdrew82:
```suggestion
# TODO: Remove this once PR is merged.
```

jdrew82 (changes requested):

nautobot_ssot/integrations/librenms/constants.py
jdrew82: Add that so we can find it later and I'll approve it.

FromMun: Added

jdrew82 (approved):

PR #1349, opened 2026-08-19
Fixed SNOW Integration to better handle duplicate names

### LINKED ISSUE #1323: ServiceNow SSoT sync silently succeeds when duplicate reference records cause MultipleResults exception
### Environment
* Python version: 
* Nautobot version: 3.2.1
* nautobot-ssot version: 4.6.0

### Expected Behavior
When the sync encounters duplicate records in a ServiceNow reference table (e.g., `cmdb_hardware_product_model`) during a create or update operation, it should surface a meaningful error or warning to the user indicating that the reference field could not be resolved. The sync job should not report overall success when a field write has silently been skipped.

### Observed Behavior
When `get_by_query` is called to resolve a reference field and ServiceNow returns more than one matching record, pysnow raises a `MultipleResults` exception. The exception is caught and `None` is returned. The sync then proceeds without setting the reference field, and the job reports success. The affected field (e.g., `product_model`) is never written to the target record in ServiceNow.

### Steps to Reproduce
1. Ensure duplicate entries exist in a ServiceNow reference table (e.g., `cmdb_hardware_product_model`) with the same name/identifier used for reference resolution.
2. Run the Nautobot SSoT sync targeting ServiceNow with a device that maps to one of the duplicated reference values.
3. Observe that the sync job completes and reports success.
4. Check the target record in ServiceNow and observe that the reference field (e.g., `model_id`) was not updated.

**Additional detail (2026-08-10):**
The references block under device in mappings.yaml resolves product_model using only model_name as the lookup column. The model has 3 identifier values and the adapter already has all of them from the initial sys_id pull. This means even with a table_query filter excluding certain manufacturers from the top-level load, the reference resolution still does a single-column lookup. Entries with the same model_name but different manufacturers collide at this stage regardless of top-level filtering.

joewesch:
# Closes: #1323

## What's Changed

This PR includes many fixes and QOL adjustments for the ServiceNow integration.

- The primary change is that device models are now paired with their manufacturer. This allows for the same model name to exist and be referenced across multiple manufacturers.
- Previously, ambiguous records would fail silently (log to module logger only). Now, they log to the JobResult and report as a failed object in the Sync. This is also now exposed better in the Dry Run so they can be observed before a true sync. It now creates and attaches a txt file to the JobResult that can be reviewed.
- Interfaces are no longer created without a parent device.
- I replaced the `sys_ids_cache` altogether. It had a few flaws and it was not able to properly look up object cross-table, so it caused multiple additional API calls. I collapsed it into the existing `sys_ids` dict instead.
- Moved multiple debug logs behind the debug conditional to cut down on log spam for normal runs.

## To Do

- [N/A] Attached Screenshots, Payload Example
- [x] Unit, Integration Tests

NTC-6416

### DIFF
--- a/changes/1323.fixed
+++ b/changes/1323.fixed
@@ -1,0 +1,1 @@
+Fixed ServiceNow integration to log and report errors for unresolved or ambiguous references, mark affected objects as failed, and resolve device model references by both manufacturer and model name to prevent name collisions.
--- a/docs/user/integrations/servicenow.md
+++ b/docs/user/integrations/servicenow.md
@@ -12,6 +12,20 @@ This integration provides the ability to synchronize basic data from Nautobot in
 
 Once the integration is installed and configured, from the Nautobot SSoT dashboard view (`/plugins/ssot/`), ServiceNow will be shown as a Data Target. You can click the **Sync** button to access a form view from which you can run the Nautobot-to-ServiceNow synchronization Job. Running the job will redirect you to a Nautobot **Job Result** view, from which you can access the **SSoT Sync Details** view to see detailed information about the outcome of the sync Job.
 
+## Reference resolution and duplicate ServiceNow records
+
+Several of the synchronized fields are ServiceNow reference (foreign key) columns: a Device's `model_id` points at a record in `cmdb_hardware_product_model`, its `location` at a record in `cmn_location`, and so on. To write one, the sync has to identify exactly one record in the referenced table.
+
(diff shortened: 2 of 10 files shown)


joewesch:
I'm setting this as Draft for now until we have a chance to test it in a customer environment.

joewesch:
Testing is complete, ready for review.

jdrew82 (approved):

Issue #891, opened 2025-07-09
Import IPv4 Prefixes from IP Fabric


jjeff07:
### Environment
* Nautobot version:  latest
* nautobot-ssot version:  latest

### Proposed Functionality

Depends: #883 

<img width="1468" height="446" alt="Image" src="https://github.com/user-attachments/assets/375d2730-7a0d-42b6-9025-49c085d5b16a" />

Example Network located in multiple sites:

<img width="872" height="194" alt="Image" src="https://github.com/user-attachments/assets/8b5b3ffa-b01c-4afb-a68e-7ada5b86edbf" />

1. Collect IPv4 networks from IP Faabric `Technology > Addressing > Managed IP > IPv4 Managed IP Summary` (Listed by site so networks could be duplicated)
2. Verify no overlapping networks (10.0.0.0/24,10.0.0.0/26) and ignore any network that has conflicts.
3. Import prefix and assign correct locations.

<img width="738" height="598" alt="Image" src="https://github.com/user-attachments/assets/82d1339b-ff42-4bd4-966a-36bc3eb0f2f5" />

### Use Case

Currently IP addresses are imported as-is but the parent prefixes are not created. This is required for mapping VLANs and VRFs to a Prefix.

jdrew82:
### Design decision: gated, not declined

Prefix import goes ahead, gated behind #883 as a per-object-type toggle alongside every other importable object.

The reason this needed settling: where another system is authoritative for IPAM, prefixes should deliberately *not* be created, and the wanted behaviour is instead to surface an address that falls outside every already-governed prefix. That is the opposite of what this issue does, so the two cannot both be unconditional.

Making it a toggle resolves it rather than picking a winner:

- **Off** — no prefixes are created. The authoritative IPAM keeps ownership, and out-of-prefix addresses are reported rather than absorbed.
- **On** — prefixes are created and located as described here, for deployments where nothing else owns them.

Off needs to be a real option, not a degraded one, so the reporting path is a genuine dependency of this decision rather than an unrelated feature.

#892 inherits this gate and needs no separate decision — see the note there.

Issue #889, opened 2025-07-09
Import Interface VRF from IP Fabric


jjeff07:
### Environment
* Nautobot version:  latest
* nautobot-ssot version:  latest

### Proposed Functionality

Depends: #883 #888 

<img width="1185" height="521" alt="Image" src="https://github.com/user-attachments/assets/652f68dd-b2a2-469b-9793-6fbf777c91eb" />

1. Collect Interface VRFs from IP Fabric's `Technology > Routing > VRF > Interfaces` table. 
    1. Do not use Managed IP table as I have found some interfaces do not have IPs but are in a VRF.
2. During Interface Import if the device `sn` and `intName` is found in the VRF Interface table then include the VRF in the import if enabled.

 
<img width="726" height="735" alt="Image" src="https://github.com/user-attachments/assets/81b15cc7-b3c7-487b-81f6-55e016f11293" />

### Use Case

IP Fabric is VRF aware and this data can help customers understand Interfaces configuration.

Issue #886, opened 2025-07-09
Import Route Targets from IP Fabric


jjeff07:
### Environment
* Nautobot version:  latest
* nautobot-ssot version:  latest

### Proposed Functionality

<img width="1183" height="352" alt="Image" src="https://github.com/user-attachments/assets/2b10fccf-3c18-4e72-8b63-b5fad0fdd481" />

1. Capture all Route Targets from IP Fabric's `Technology > MPLS > L3 VPN > VRF Route Targets`
2. This should include columns `RD`, `Import RT`, `Export RT`
3. Add RT to Nautobot Route Targets

### Use Case

Importing RT from IP Fabric allows for accurate import of VRF information.

<img width="954" height="513" alt="Image" src="https://github.com/user-attachments/assets/ba0e001a-11a1-44f3-9516-34c056f8fe64" />

justin-jeffery-ipf:
@darenfulwell Can you please see if this makes sense?  I have never worked with RTs and RDs.  Perhaps there is a better table or something to use?

PR #1352, opened 2026-08-25
Move IP Fabric sync settings to the Job form

### LINKED ISSUE #883: Move IP Fabric Sync Settings to Job Form
### Environment
* Nautobot version:  latest

### Proposed Functionality

Users should be able to have more fine grained controls on the types of objects they would like to synchronize to Nautobot.  Perhaps a user does not want to sync specific items. 

Nice Features:

- Create a "default" template form so why a user loads the job form the correct items are pre-selected.
- Have an administrative way to disable an item globally so it cannot be selected.
- Example: Disable importing Primary IP.  Many customers use the CMDB as the record of truth and IPF does not have a concept of a management IP only the Login IP we use which could be any IP on the device.

Image

Example from Netbox:

Image

### Use Case

As we are looking to extend the capabilities of the integration users need to be able to enable or disable specific items.  Looking at synchronizing VRF, IPv6, etc. could dramatically increase the amount of data.

jdrew82:
# Closes: #883

## What's Changed

The Job form now carries a checkbox per object type, so a run can be narrowed to the data an installation actually wants IP Fabric to own. Alongside the per-run controls, `PLUGINS_CONFIG` can pre-select a type or remove it from the form entirely.

Updated Job form

Every object type is registered once in `nautobot_ssot/integrations/ipfabric/sync_scope.py`, and the form fields, field ordering, administrative deny list and the resolved scope a run works from are all derived from that registration. Adding an object type is a `SyncableObject` plus a gate in both adapters, rather than another keyword argument threaded through the Job and both adapters.

| Object type | Job field | Default | Requires |
|---|---|---|---|
| `locations` | Sync Locations | On | |
| `manufacturers` | Sync Manufacturers | On | |
| `device_types` | Sync Device Types | On | |
| `roles` | Sync Roles | On | |
| `platforms` | Sync Platforms | On | |
| `interfaces` | Sync Interfaces | On | |
| `ip_addresses` | Sync IP Addresses | On | `interfaces` |
| `primary_ip` | Sync Primary IP | On | `ip_addresses` |
| `vlans` | Sync VLANs | On | |
| `cables` | Sync Cables | Off | `interfaces` |

Devices are always synced: every other object type is either a Device or hangs off one, so a run with Devices excluded has nothing left to do.

Every default is unchanged from today's behaviour, so an existing installation that opens the form and clicks through gets exactly the sync it got before.

### The rule that makes a toggle safe

A toggle gates **both** adapters. Reading an object type from Nautobot but not from IP Fabric makes every existing record look absent from the source, which the sync would then delete — so the two sides cannot be allowed to disagree about scope. They are given one `SyncScope` object, held on the shared `DiffSyncModelAdapters` base, rather than a boolean each.

Where the toggle governs an attribute rather than a model, both adapters report the same empty value so it diffs as unchanged and the stored value is left alone.

### Requirements between types

`requires` records the edges, and `SyncScope` closes over them: turning Interfaces off takes IP Addresses with it, and Primary IP with those. A selection dropped this way is named in the Job log with the requirement that was unmet — a ticked box that silently does nothing is the failure mode worth reporting.

### Locations

Locations cannot be gated the way the others are: they are the root of the object tree, so leaving them unloaded would take every Device and VLAN with them. Out of scope a Location is still loaded, but as a tree node rather than as synced data — both adapters report the same placeholder attributes, so it matches on its name and diffs as unchanged, and `SKIP_UNMATCHED_DST` stops one the source does not report being deleted.

Two consequences, both intended:

- A Location this sync did not load is looked up by name rather than created, so a Location another SSoT App owns is found and the Devices at it land there. This matters most with **Sync Tagged Only**: a Location created by another App carries no IP Fabric tag, so this sync never loads it, and the name lookup is what connects the two. A Device whose Location has not arrived yet is attempted and reported by name, so a run made before the owning App catches up reads as work to retry.
- A Nautobot Location that IP Fabric does not report is left alone along with its Devices. A sync holding no opinion about which sites exist cannot read a missing site as evidence that the Devices at it are gone.

### Supporting objects

Manufacturers, Device Types, Roles and Platforms are not part of the object tree — nothing is a child of them, and they exist because a Device needed one. Deselecting them does not stop Devices syncing; it turns a get-or-create into a lookup, so an estate whose hardware catalogue or role taxonomy is owned elsewhere is used rather than added to.

What an absent entry costs depends on whether Nautobot requires it:

| Deselected | A Device needing one Nautobot does not hold |
|---|---|
| `manufacturers` | Skipped, if a Device Type would have had to be created under the missing vendor |
| `device_types` | Skipped — Nautobot requires a Device Type |
| `roles` | Skipped — Nautobot requires a Role |
| `platforms` | Synced without a Platform — Nautobot treats it as optional |

Two details worth review attention:

- Deselecting `manufacturers` restricts *adding* vendors, not using them. A Device Type is still created under a Manufacturer that already exists, and refused only where inventing a vendor would be required.
- Deselecting `roles` also stops the `ipfabric_type` custom field being written. In scope that field records what IP Fabric called the role and is how a Role this integration created is matched again; out of scope, Roles match on that field first and then on their **name**, so a Role another system owns is found without this sync stamping anything on it.

### Settings

```python
PLUGINS_CONFIG = {
    "nautobot_ssot": {
        # Pre-select a checkbox; an operator can still override it for one run.
        "ipfabric_sync_cables": True,
        # Remove object types from the form altogether. Cannot be re-enabled by an
        # operator, by the REST API, or by a scheduled Job saved before it was disabled.
        "ipfabric_disabled_sync_objects": ["locations", "primary_ip"],
    }
}
```

Disabling a type also disables everything that requires it. Every exclusion is reported in the Job log, so a run whose selections were overruled says so rather than quietly doing less than asked.

### Fetching only what a run reads

The toggles gated what got consumed, not what got downloaded. A run with Interfaces deselected still paid for a full paginated fetch of every interface in the fabric — the largest single request the Job makes — and then never read it; the same held for managed IPs and the VLAN summary. Each table is now fetched only when something in scope reads it. Stack membership stays unconditional, being Device data. On the Nautobot side, `primary_ip4`/`primary_ip6` are no longer joined when addresses are out of scope, and the Interface prefetch narrows or drops with the scope.

### One bug fixed along the way

Gating IP addresses by nulling `device_primary_ip` had the opposite of the intended effect: `pseudo_management_interface` reads a null address as "no Interface claims it" and so *fabricates* the pseudo interface. It is now skipped outright when addresses are out of scope. Found by a test written for the new gate.

## To Do

- [x] Added change log fragment(s) — `changes/883.added`, `changes/883.changed`
- [x] Unit, Integration Tests

### Testing

354 tests in `nautobot_ssot/tests/ipfabric/` pass, up from 233 on `develop`. Every line added by this PR is covered. New modules:

- `test_sync_scope.py` — the registry, requirement closure, defaults and the administrative deny list
- `test_location_scope.py` — diffs and syncs a Nautobot estate against a source that disagrees about Locations three ways
- `test_supporting_object_scope.py` — Devices created against a Nautobot that holds some supporting objects and not others, against the real database

### Remaining work and constraints from design

- **VRFs and Prefixes have no toggle** because neither is synced yet (#886–#889, #891). Registering one now would put a checkbox on the form that does nothing whatever its state. Each becomes a one-line `SyncableObject` when its feature lands.
- **Stacks and Virtual Chassis have no toggle.** The types here all have clean both-sides symmetry; stacks do not. Deselecting them on the source side collapses a stack from N member Devices to one, so the Nautobot side would report members the source no longer does and safe-delete would fire on every member. That needs its own design rather than a checkbox.
- **Every attribute added to sync scope stops being enforceable by configuration compliance**, because drift is absorbed and the intended config re-renders to match. That is a property of syncing, not of this PR, but the per-type controls are what make it a deliberate choice rather than an inherited one.
- `#883` is also named as a dependency by a dozen other issues. The per-object-type granularity here is what several of them were waiting on; see the comment on #883 for the object types whose gating was settled as a design input.


jdrew82:
Addressed the NTC review. Two commits.

### Correctness: the pseudo management Interface (`45743385`)

Confirmed as a real bug, reproduced before fixing. The IP Fabric adapter stops fabricating the `pseudo_mgmt` Interface when addresses are out of scope, but the Nautobot adapter had no matching exclusion — so a `pseudo_mgmt` Interface left by an earlier run was reported by one side only, read as absent from the source, and deleted (or safe-deleted to `Deprecated` under the default).

This is exactly the asymmetry the scope design exists to prevent, and the one case where "deselecting IP Addresses leaves address data alone" did not hold: every *real* Interface was safe because both adapters report the same empty address; the synthetic one was not.

Both sides now agree, keyed off a single `PSEUDO_MANAGEMENT_INTERFACE_NAME` in `constants.py` rather than a literal in each adapter. Covered at two levels — the Nautobot adapter no longer loads it, and a diff against a source that does not report it produces no delete. I verified both tests fail with the guard removed; the diff-level one was initially vacuous because Interfaces are nested under Location → Device rather than at the top of `diff.dict()`, so it now walks the tree.

### Standards (`74b8e90b`)

- The new scope log line now passes its value as an argument instead of interpolating an f-string. The surrounding calls are left alone: they are pre-existing and consistent with each other, so converting one of them would create inconsistency without fixing the file.
- The deferred import of the `Location` model, with its `import-outside-toplevel` suppression, turned out to be unnecessary — a test module is not part of any import cycle with the code it imports. Hoisted, suppression removed.

### The `job_scoped_cache` question — checked, no issue

`manufacturer_obj` in the cache key is safe. The key is `(args, tuple(sorted(kwargs.items())))`, and Django's `Model.__hash__` is `hash(self.pk)` with `__eq__` comparing class and pk, so two Python objects for the same Manufacturer hash and compare equal and share a cache entry. The sort never compares model objects either, since it orders by unique kwarg name. Both resolvers only ever pass a saved object or bail on `None`, so the `pk is None` unhashable case cannot arise.

### Not changed

The 217-character `options` f-string log line. It is pre-existing, and the file uses that pattern throughout; restructuring one of roughly fifteen such calls would leave the file less consistent than it is now. Worth doing as a separate pass over the module.

Also noting for the record: the hand-maintained "Requires" column in the setup docs is not generated from `SYNCABLE_OBJECTS`, so a future `requires` change needs updating in both places.

357 tests pass, every added line covered, ruff/markdownlint/docs clean, no new pylint messages.

alhogan (approved):

Issue #1356, opened 2026-09-01
IP Fabric sync never settles an Interface's subnet mask for an address reported on more than one Device


jdrew82:
### Environment

* Nautobot version: 3.1 and 3.2
* nautobot-ssot version: develop

### Expected Behavior

An Interface's `subnet_mask` settles after one sync, and a second sync of unchanged data reports nothing.

### Observed Behavior

For an IP address that IP Fabric reports on more than one Device, `subnet_mask` is reported on every run and the value alternates between runs:

```
interface: GigabitEthernet4
  subnet_mask   IPFabricDiffSync(255.255.255.0)   NautobotDiffSync(255.255.255.128)
```

#### Cause

The two systems describe the subnet at different granularities.

`IPFabricDiffSync.load_data` indexes addressing **by serial number**:

```python
managed_ipv4[ip_address["sn"]].update({ip_address["ip"]: ip_address})
...
self.load_device_interfaces(device_model, interfaces.get(device.sn, []), device_primary_ip, managed_ipv4.get(device.sn, {}))
```

so the subnet reported for an address is whatever the Device it was read from says. Nautobot holds one `mask_length` per IPAddress, and parents an address to the most specific Prefix containing it, so two rows for one host in a Namespace cannot coexist. An address on two Devices reported in two subnets therefore cannot have both masks, and each Interface's update rewrote what the other had written. Which one won depended on the order the writes happened in, so the reported difference alternated.

This is reachable wherever one address is on several Devices: an HA pair sharing a link address, VRRP, anycast, or a cloud fabric where IP Fabric describes each subnet as an Interface.

### Steps to Reproduce

1. Have IP Fabric report one address on two Devices with different `net` values.
2. Sync twice.
3. Both runs report `subnet_mask` for one of the Interfaces, with the value alternating.

Covered by `test_one_address_reported_with_two_subnets_does_not_oscillate` in `nautobot_ssot/tests/ipfabric/test_sync_convergence.py`.

Issue #1354, opened 2026-09-01
Model IP Fabric shared segments as breakout Cables on Nautobot 3.2


jdrew82:
### Environment

* Nautobot version: 3.2 (feature is unavailable on 3.1, which this app also supports)
* nautobot-ssot version: develop

### Proposed Functionality

Model an IP Fabric shared segment as a Nautobot **breakout Cable**, so that every Interface in the segment is recorded rather than only one of them.

IP Fabric describes a shared segment as a link from each Interface in it to the segment. A cloud subnet is the clearest case: the subnet's own Interface is reported on one link per member, for example `NAE:subnet-0688ce44d04efa2f7` to each `eni-*` in that subnet.

Nautobot terminates at most one Cable on an Interface. Breakouts do not change that — `CableToCableTermination.interface` is a `OneToOneField`, and the model states that "each termination is on at most one cable". What 3.2 adds is several terminations *per side of one Cable*, which is the shape a shared segment needs: A side the segment's Interface, B side each member.

#### What this needs

* **A `CableType` per fan-out width.** `CableToCableTermination.clean()` limits `connector` to 1 on each side when `cable_type` is null, so a Cable without a CableType is strictly point to point. A breakout needs a CableType with `a_connectors=1`, `b_connectors=N`, a `total_lanes` evenly divisible by both, and a validated `mapping`. A subnet with 7 members needs a 1x7, one with 23 needs a 1x23. Whether the integration should create these, or require them to be configured, is the main open question.
* **Cable identity derived from the shared end.** Both adapters currently sort a link's two endpoints and use the lower as the A side. `CableType.clean()` rejects `a_connectors > b_connectors`, so the trunk has to be the A side; a segment whose device name sorts after its members' would be invalid. The identifier both adapters agree on would have to come from which end is shared instead of from sort order.
* **A 3.1 fallback.** `nautobot = ">=3.1.0,<4.0.0"`, and 3.1 has no breakouts, so the current behaviour has to remain on that version. `nautobot_ssot/integrations/ipfabric/utilities/cables.py` already branches on `CABLE_TERMINATIONS_ARE_JOINED` for a related reason.

Interfaces are in `BREAKOUT_COMPATIBLE_TERMINATION_TYPES`, so the termination type itself is not a barrier.

#### Worth deciding first

[...]

Issue #1353, opened 2026-08-27
IP Fabric: IP Addresses and Cables are queued for deletion but never deleted


jdrew82:
### Environment
* Python version: 3.12
* Nautobot version: 3.1.0
* nautobot-ssot version: current `develop` (`928c0f08`)

### Expected Behavior

With **Safe Delete Mode** disabled, an IP Fabric sync should delete every Nautobot object it has decided is gone, including the IP Addresses it removes from an Interface and the Cables it removes between two.

### Observed Behavior

IP Addresses and Cables are queued for deletion but never deleted. The sync reports them as deleted and removes them from its own store, so nothing in the job log indicates the rows are still there.

`DiffSyncExtras.safe_delete` queues an object under a key derived from its class name:

```python
# nautobot_ssot/integrations/ipfabric/diffsync/diffsync_models.py:77
self.adapter.objects_to_delete[f"_{nautobot_object.__class__.__name__.lower()}"].append(nautobot_object)
```

`NautobotDiffSync.sync_complete` then drains only four of the keys that produces:

```python
# nautobot_ssot/integrations/ipfabric/diffsync/adapter_nautobot.py
for grouping in ("_vlan", "_interface", "_device", "_location"):
```

Mapping every `safe_delete` call site to the key it generates:

| `safe_delete` call site | key generated | drained by `sync_complete` |
|---|---|---|
| `Location.delete` | `_location` | yes |
| `Device.delete` | `_device` | yes |
| `Interface.delete` (the Interface) | `_interface` | yes |
| `Vlan.delete` | `_vlan` | yes |
| `Interface.delete` (its IP Addresses) | `_ipaddress` | **no** |
| `Cable.delete` | `_cable` | **no** |

So `objects_to_delete["_ipaddress"]` and `objects_to_delete["_cable"]` accumulate for the whole run and are never read.

Two consequences beyond the rows surviving:

1. `safe_delete` calls `super().delete()` on the DiffSync model in the same branch, so the object is removed from the DiffSync store and counted as deleted. The discrepancy is silent.
2. Because the lists are never emptied, they hold every queued IP Address and Cable model instance until the adapter is garbage collected.

Safe Delete Mode is enabled by default, and in that mode nothing is queued at all, so only runs with it disabled are affected.

### Steps to Reproduce

1. Create a Device tagged `SSoT Synced from IPFabric` with an Interface that has an IP Address assigned, where IP Fabric does not report that Interface.
[...]

PR #1357, opened 2026-09-01
Optimize the IP Fabric sync for bulk updates

### LINKED ISSUE #882: Optimize IP Fabric Sync for Bulk Updates
### Environment
* Nautobot version:  latest

### Proposed Functionality

Review how IP Fabric imports, edits, and deletes objects in Nautobot and look for optimizations for bulk changes.  Consideration needs to be taken into account for many parent-child relationships.  

### Use Case

Currently when performing synchronization the job can go over the max time limit due to the amount of data being sync'ed.  The majority of IPF customers have 1k-5k devices which could result in 100k+ interfaces, IP addresses etc.  As we are looking to extend the capabilities this amount will increase significantly when considering adding VRF, IPv6, VLAN, etc.

### LINKED ISSUE #1355: IP Fabric sync loads untagged Devices but will not write to them, so their diffs never settle
### Environment

* Nautobot version: 3.1 and 3.2
* nautobot-ssot version: develop

### Expected Behavior

With **Sync Tagged Only** deselected, the IP Fabric sync writes to every Device the Nautobot adapter loaded, and a second sync of unchanged data reports nothing to do.

### Observed Behavior

Interface changes on a Device that does not carry the `SSoT Synced from IPFabric` Tag are reported on every run and never applied. The job completes without errors, because the write path reports the miss as a warning:

```
Unable to retrieve a Device named <name>, so unable to update its interface named <name>
```

The diff shows the same attributes indefinitely. Seen in the field as `subnet_mask` never settling on cloud Devices (AWS VPCs, Azure VNets) and on firewalls that another process had created:

```
interface: GigabitEthernet4
  subnet_mask   IPFabricDiffSync(255.255.255.128)   NautobotDiffSync(255.255.255.0)
```

#### Cause

The two sides disagree about which Devices are in scope:

* `NautobotDiffSync.filtered_devices()` filters on the Tag **only when** `sync_ipfabric_tagged_only` is set, so with the option off it loads untagged Devices as well.
* `Interface.create`, `Interface.update`, `Interface.delete` and `Cable.resolve_interfaces` resolved the Device through a lookup that **always** required the Tag.

So a run with the option off loads Devices it then refuses to write to. It is the same invariant #883 records for the object type toggles: a condition that gates loading has to gate writing identically, or the objects it excludes from one side look permanently changed from the other.

The default for **Sync Tagged Only** is on, which is why this is not seen in a default run.

### Steps to Reproduce

1. Sync with **Sync Tagged Only** deselected, so that untagged Devices are loaded.
2. Remove the `SSoT Synced from IPFabric` Tag from a Device, or use one another process created.
3. Change something IP Fabric reports about one of its Interfaces.
4. Sync twice. The change is reported both times and never applied.

Covered by `nautobot_ssot/tests/ipfabric/test_sync_convergence.py`, which requires a second sync of unchanged data to report nothing.

jdrew82:
# Closes: #882

Also closes #1355 and #1356, two non-convergence bugs found while validating this work against a live estate. Both are pre-existing on `develop` rather than introduced here, and both are described below.

### Bulk Updates (#882)

The issue asks for the work an IP Fabric sync does per object to come down, because a large estate takes the job over its time limit. This addresses it in two layers.

**Per-object cost, on by default.** Nothing about the sync's guarantees changes here; it simply does less work for the same result.

| | Before | After |
|---|---|---|
| Loading 20 Locations | 103 queries | 8 |
| Deleting 100 objects | 2400 queries | 25 |
| Creating an addressed Interface | 33.2 queries/object | 19.2 |
| Re-syncing an Interface | 28.1 queries/object | 24.2 |

Devices and VLANs are read in one query each rather than one per Location; creating an Interface, Location or VLAN takes one validated save rather than two; a create or delete records one change log entry rather than one per write; the safe delete tag is resolved once per model rather than once per object; and deletions go in batches.

The change log work uses `deferred_change_logging_for_bulk_operation()` **per object**, not per run. Nautobot keys deferred changes per object, so a per-object scope saves everything a run-wide one would without holding every changed instance in memory or keeping one transaction open for the length of the sync. Measured at -22% on creates and -23% on safe deletes.

**Bulk Write Mode, opt in.** A new job option writes Locations, Devices, Interfaces, IP Addresses and VLANs in batches. Measured at **8.80s to 0.46s for 200 addressed Interface creates, a 19x improvement.**

It is off by default because it gives up per-object `clean()`, signals and a change log entry each. Database constraints still apply, a refused batch is retried an object at a time so the offending row is named, and the trade is documented along with two differences to expect. Cables are deliberately excluded: creating one drives cable-path construction through signals a batched insert does not fire.

An intermediate design was measured and rejected: batching only the change log across objects, while still saving each object individually, saves 0.6%. `validated_save()` costs 11.1 q/object, `save()` 7.0, `bulk_create` ~0. The saving is in the writes, not the change log, which is why the mode batches the inserts themselves.

### Devices the sync loaded but would not write to (#1355)

`filtered_devices()` filters on the `SSoT Synced from IPFabric` tag only when **Sync Tagged Only** is set, so with that option off it loads untagged Devices. Every Interface and Cable operation, meanwhile, resolved its Device through a lookup that always required the tag.

So a run with the option off loaded Devices it would not write to. The miss was reported as a warning, not an error, so the job completed cleanly while the same attributes were reported on every run and never applied. Seen in the field as an Interface's subnet mask never settling on cloud Devices and on firewalls another process had created.

The lookup now takes the condition the load applies. This is the invariant #883 records for the object type toggles, in a place it had not been applied: a condition that gates loading has to gate writing identically, or what it excludes from one side looks permanently changed from the other.

### An Interface's subnet mask alternating between runs (#1356)

IP Fabric indexes addressing by serial number, so the subnet reported for an address is whatever the Device it was read from says. Nautobot holds one `mask_length` per IP Address and parents an address to the most specific Prefix containing it, so two records for one host in a Namespace cannot coexist.

An address on two Devices reported in two subnets therefore could not carry both masks, and each Interface's update rewrote what the other had written. Which one won depended on write order, so the difference was reported every run with the value alternating.

The mask is now resolved once per address before any Interface is loaded, taking the narrowest report: narrowest because it agrees with the address's parent, and a fixed rule rather than the order IP Fabric answered in, which is not guaranteed. Addresses this was needed for are logged, since an Interface then carries a mask it was not reported with.

Reachable wherever one address sits on several Devices: an HA pair sharing a link address, VRRP, anycast, or a cloud fabric where each subnet is an Interface.

### Other fixes

Eleven `.fixed` fragments under #882. The ones worth calling out:

- **A Device queued with a primary IP Address the same run had yet to write.** PostgreSQL defers foreign key checks to `COMMIT`, which belongs to whichever model operation triggered the flush, so a broken reference ended the job somewhere unrelated to its cause. Each batch now checks its own foreign keys inside its own savepoint, where the existing retry reports them. That contains the class of mistake, not just the instance.
- **Cables failing in bulk mode**, because a Cable reads its Interfaces back from the database while the mode had only queued them.
- **A Cable recorded for only one of the links on an Interface IP Fabric reports several times**, such as a cloud subnet's, and a different one each run. Links are now reduced to at most one per Interface in sorted order, so a re-sync keeps the same one. Representing such a segment properly needs a breakout Cable; filed as #1354.
- **Object caches surviving a failed run** into the next run on the same worker, including instances from rolled back transactions.
- Several IP Address resolution fixes: a missing parent Prefix in bulk mode, an address whose only containing Prefix is longer than the reported mask, an address reported on more than one Interface, and an address Nautobot holds under a different mask.

## Testing

**A convergence harness.** `test_sync_convergence.py` syncs the fixtures into an empty Nautobot, loads Nautobot again, diffs against the same source, and requires the second diff to be empty. When it fails it names the attribute. Seven scenarios, per-object and in bulk mode.

This is the check that found #1355 and #1356, after three rounds of fixing the wrong thing from a one-line symptom. A sync that reports the same change every run is as much a defect as one that fails, and neither adapter's own tests can catch it because it comes from the two describing one object differently.

**Both supported Nautobot versions.** `nautobot = ">=3.1.0,<4.0.0"`, and the integration branches on `CABLE_TERMINATIONS_ARE_JOINED` for 3.2's Cable layout. Verifying on 3.2 found two tests that had only ever passed on 3.1: `_get_closest_parent()` gained a required argument, and migration `0092` dropped `dcim_interface_device_name_unique`. Nothing shipped called the former; the latter is a real difference in bulk mode on 3.2 and is now documented. Full suite green on 3.1.0 and 3.2.4 from a fresh database.

**Cost tests.** `test_sync_query_counts.py` asserts the query counts above, so a regression in them fails rather than going unnoticed.

## To Do

- [x] Unit, Integration Tests
- [x] Documentation Updates

## Remaining work and constraints

**Not addressed here, deliberately:**

- **#1353** — IP Addresses and Cables are queued for deletion but never deleted. Found while reading `objects_to_delete`; a behaviour change beyond this issue's scope.
- **#1354** — modelling a shared segment as a breakout Cable on Nautobot 3.2. Needs a `CableType` per fan-out width and Cable identity derived from the shared end rather than sort order, and it is worth deciding first whether a breakout is the right representation for a multi-access segment.
- **An IPv6-only Interface crashes the load.** `cidr_to_netmask` accepts 0-32, so an Interface whose only addresses are IPv6 raises `ValueError` out of `load_interfaces`. IPv4 sorts first, so a dual-stack Interface is safe. What the sync should report for such an Interface is a product decision, and IP Fabric's side reads `managed_ip_ipv4`, so it can never report a match. Not filed yet.
- **Every synced IP Address takes a write per run**, because `create_ip` re-stamps the `system_of_record` and last-synchronized fields whether or not anything changed. Real cost on a large estate, pre-existing, and out of scope here.
- **Two pre-existing pylint warnings** in `cables.py` flag the 3.1 branch of `cabled_interfaces` as deprecated in 3.2. That branch is required while the floor is 3.1.

**Constraints worth a reviewer's attention:**

- **Bulk mode assumes the Global Namespace.** Every address lookup resolves through it. A deployment using several Namespaces is not covered.
- **The single-save collapse for Locations and VLANs rests on the suite passing**, unlike Interfaces (measured directly) and Devices (caught by a #883 test that the collapse initially broke — `Device.create` therefore keeps its `get_or_create` outside bulk mode, because `get_or_create` inserts without `full_clean()` and the Platform scope control deliberately allows a Platform whose Manufacturer differs from the DeviceType's, which `Device.clean()` rejects).
- **A flush inside a Cable create shares that operation's transaction.** An exception escaping the create would roll back up to `PENDING_WRITE_HIGH_WATER` rows. Every database call in it is guarded, and the per-batch constraint check means a deferred violation is now caught inside the batch rather than at that commit.


jdrew82:
## Deferred work, filed

Everything found during this change and its review that was deliberately not fixed here now has an issue, so nothing depends on this PR's history to be remembered.

| Issue | What | Why not here |
|---|---|---|
| #1362 | Remaining per-object query and memory waste — six items, largest first: an unchanged IP Address still written every run (~11 queries each), a Cable written twice on create, `resolve_ip`'s two SELECTs per address, ~4 queries per VLAN, `objects_to_delete` holding ORM instances, and a needless SELECT on a fresh Interface | Each either changes behaviour, sits outside this diff, or touches the address lookup that caused three production bugs during this work |
| #1364 | Structural follow-ups — retiring the 15 hand-applied `deferred_change_logging` decorators and their AST test via `__init_subclass__`; replacing the `pending` argument threading with a writer seam; `Cable` knowing about batching; `get_syncable_device` mirroring only the Tag half of the adapter's scope; four smaller items | The first needs verifying against pydantic's metaclass; the second is a refactor across `nbutils` and five models |
| #1360 | An IPv6-only Interface makes the Nautobot adapter load raise `ValueError` (`cidr_to_netmask` accepts 0–32) — reproduced | What to report for such an Interface is a product decision, and interacts with #1344 |
| #1361 | `safe_delete_mode` is class state, so two runs in one worker share it | A correctness bug in its own right; this PR fixed the other two instances of that pattern (`objects_to_delete`, `bulk_write_mode`) and stops there |
| #1363 | The sync assumes the Global Namespace for every address and Prefix | Needs a design decision about where the Namespace mapping comes from, since IP Fabric has no such concept |
| #1353 | IP Addresses and Cables queued for deletion but never deleted | Behaviour change beyond this issue |
| #1354 | Modelling shared segments as breakout Cables on Nautobot 3.2 | Needs a `CableType` per fan-out width and Cable identity derived from the shared end |

Also recorded: a comment on #1340 noting that this PR makes that fallback's consequence wider by writing the fabricated `255.255.255.255` onto an address it reuses. That is not new — `Interface.update` already did it — but the two write paths now agree, and #1340 is the only place it can be fixed properly.

The three constraints in the PR description above (the untested single-save collapse for Locations and VLANs, the flush sharing a Cable create's transaction, and bulk mode's Namespace assumption, now #1363) stay there as things for a reviewer to weigh rather than separate issues.

nautobot_ssot/integrations/ipfabric/bulk_writes.py
```diff
@@ -0,0 +1,240 @@
...
+        for instance in objects:
+            # `bulk_create` marks everything it hands to the database as saved, and a constraint
+            # PostgreSQL defers to `COMMIT` fails after that, so the batch being retried here was
+            # rolled back while its objects still look written. Left that way, `save()` would issue
+            # an `UPDATE` matching no rows, and a `clean()` that reads its own row back would raise
+            # `DoesNotExist` instead of a validation error.
+            instance._state.adding = True  # pylint: disable=protected-access
+            try:
+                with transaction.atomic():
+                    instance.validated_save()
```
alhogan, line 196: seems like a pretty heavy trade off for potentially 1 bad record. Should the default batch size be a little smaller to help mitigate the impact? Might be worth mocking up a performance test to see.

alhogan, line 196: Probably worth testing this failure condition with a device that has a device type with an interface template. You could potentially get duplicate interfaces since validated_save() would create them and the ones being ingested from IPFabric would also be created even if they are duplicates.

nautobot_ssot/integrations/ipfabric/utilities/nbutils.py
```diff
@@ -34,68 +45,117 @@
...
+    if is_new and pending is not None:
+        return queue_new_object(pending, location_obj, key=location_name)
+
+    # tag_object performs validated_save(), which is the only save a new Location takes.
+    try:
+        tag_object(nautobot_object=location_obj, custom_field=LAST_SYNCHRONIZED_CF_NAME)
+    except (DjangoBaseDBError, ValidationError):
+        if logger:
+            logger.warning(
+                f"Unable to perform a validated_save() on Location {location_name} with an ID of {location_obj.id}"
```
alhogan, line 156: wouldn't you want to `return None` in this case since the location obj failed to create?

nautobot_ssot/integrations/ipfabric/bulk_writes.py
jdrew82: If that's a concern we could toggle the SkipAutoComponentCreation toggle that I added [here](https://github.com/nautobot/nautobot/pull/9027).

jdrew82: In my experience 1k records tends to be a reasonable batch size for these operations and the individual handling in those rare cases they happen don't actually impact import time significantly.

nautobot_ssot/integrations/ipfabric/utilities/nbutils.py
jdrew82: Isn't that implied already and doesn't require the explicit return statement?

nautobot_ssot/integrations/ipfabric/bulk_writes.py
jdrew82: Per our discussion this is already addressed.

nautobot_ssot/integrations/ipfabric/utilities/nbutils.py
jdrew82: Scratch my last comment, you're right, and I had this backwards. The except falls through to return location_obj at the end of the function, so None isn't implied.

It's also worse than it was before this PR. On develop the Location came from Location.objects.get_or_create(), so it was already persisted by the time tag_object ran and returning it after a failed stamp was harmless. This PR constructs it unsaved and lets tag_object's validated_save() be its only save. So now that path returns a Location with no row, and the Devices built against it take an FK to nothing.

Fixing it, but conditionally: return None only when is_new, since for an existing Location a failed re-stamp shouldn't throw the Location away.

nautobot_ssot/integrations/ipfabric/bulk_writes.py
jdrew82: Reopening this one. I checked the mechanism and it is reachable, so "already addressed" was premature.

_insert_one_at_a_time sets _state.adding = True, present_in_database is not _state.adding, and Device.save() calls create_components() when it sees a new Device. So the fallback path creates template components while the bulk path doesn't, which is the asymmetry you're pointing at. Devices are written before Interfaces, so a Device batch that falls back gets its template Interfaces and then IP Fabric's Interfaces for that Device go in on top. On 3.1 the (device, name) database constraint turns that into a rejected IP Fabric interface plus its dropped address rows; on 3.2 the uniqueness check is Python-only and a batched insert skips it, so you get the duplicates you described.

SkipAutoComponentCreation is the right tool but it's 3.2+, and we support 3.1, so it wants a capability guard like CABLE_TERMINATIONS_ARE_JOINED rather than a plain import. Adding a regression test with a Device Type that has an interface template, on the fallback path specifically.

jdrew82: Fair point, and it's the same root as the interface-template thread: the fallback's blast radius is exactly the batch size, in wall-clock and in how many Devices take the per-object path they weren't meant to.

Rather than shrinking the default I'd rather bisect: split a refused batch in half, retry each half in bulk, recurse, and only fall back to validated_save() at a batch of one. One bad row in 1000 then costs about twenty bulk attempts and one validated save instead of a thousand, and the offending row is still named. Happy to do that here or as a follow-up, it's contained either way. batch_size is already a constructor argument, so exposing it as a setting is the cheap version if you'd prefer that for now.

jdrew82: I've gone ahead and ensured that Job form setting is passed so it takes affect when sync_data() is called.

nautobot_ssot/integrations/ipfabric/utilities/nbutils.py
jdrew82: Addressed.

alhogan (approved):

Issue #1343, opened 2026-08-18
Adapter fabricates a pseudo_mgmt Interface that does not exist on the device


jdrew82:
### Environment
* Python version:  3.12
* Nautobot version:  3.2.1
* nautobot-ssot version:  4.6.1

### Expected Behavior

Every Interface in Nautobot corresponds to an interface that exists on the device.

### Observed Behavior

**A synthetic interface is fabricated.** When a device's primary IP matches none of its discovered interfaces, `adapter_ipfabric.py::pseudo_management_interface` invents an interface named `pseudo_mgmt` to hold the address:

```python
return {
    "hostname": hostname,
    "intName": "pseudo_mgmt",
    "dscr": "pseudo interface for NAT IP address",
    ...
}
```

This object does not exist on the device. It is written to Nautobot as an ordinary Interface, indistinguishable from a real one except by name. Any consumer that treats Nautobot interfaces as real — configuration rendering, interface counts, cabling, capacity reporting — is misled by it. The original motivation (somewhere to attach a NAT management address) is legitimate, but the mechanism makes a false statement about the device.

**Virtual interface coverage is unverified.** Virtual interfaces are synced only if IP Fabric returns them in the interface inventory. `convert_media_type` has name-based fallbacks mapping VLAN, loopback, tunnel and VXLAN names to a virtual type, which implies they do appear, but this has not been confirmed against a current IP Fabric release. If SVIs, loopbacks, tunnels or subinterfaces are absent from the inventory, they are silently absent from Nautobot, and any configuration rendered from Nautobot data omits them.

### Steps to Reproduce
1. Sync a device whose primary IP is a NAT address not configured on any discovered interface.
2. Observe an Interface named `pseudo_mgmt` in Nautobot that does not exist on the device.
3. Separately, compare the SVIs, loopbacks, tunnels and subinterfaces present on a device against those synced into Nautobot.

Suggested direction: two parts. Mark the fabricated interface so consumers can exclude it (a dedicated type, a tag, or a naming contract), or find a way to hold the address that does not invent an interface. And confirm which virtual interface classes a current IP Fabric release returns, so coverage is known rather than assumed.

PR #1369, opened 2026-09-04
Check what IP Fabric reported before writing an address or creating a supporting object

### LINKED ISSUE #1340: Interface address prefix length silently falls back to a host mask
### Environment
* Python version:  3.12
* Nautobot version:  3.2.1
* nautobot-ssot version:  4.6.1

### Expected Behavior

An interface address synced from IP Fabric carries the prefix length actually configured on the device, or the sync reports that it could not determine one.

### Observed Behavior

When an interface's address is not found in the managed IPv4 map, `adapter_ipfabric.py` falls back to a host mask:

```python
if ip_address in managed_ipv4 and managed_ipv4[ip_address].get("net"):
    subnet_mask = str(ipaddress.ip_interface(managed_ipv4[ip_address]["net"]).netmask)
else:
    subnet_mask = "255.255.255.255"
```

The address is then written to Nautobot with a `/32`. Nothing distinguishes an interface genuinely configured as a host route from one whose real mask could not be resolved, and no warning is logged.

The consequence is silent bad data rather than a visible failure. A `/32` resolves to the wrong parent prefix or to none, so the address lands in the wrong place in IPAM, and any template rendering a subnet mask from Nautobot renders the wrong value.

### Steps to Reproduce
1. Sync a device from IP Fabric that has an interface address IP Fabric does not return in `technology.addressing.managed_ip_ipv4` (filtered to `type eq primary`).
2. Inspect the resulting IP address in Nautobot.
3. The mask length is 32 regardless of the mask configured on the device, and the job log records nothing.

Suggested direction: treat an unresolvable mask as a failure for that address, log it, and skip writing it rather than substituting a host mask. If a fallback is wanted for compatibility, it should be opt-in and logged at warning level on every use.

### LINKED ISSUE #1346: Add a strict mode that matches supporting objects instead of creating them
### Environment
* Nautobot version:  3.2.1
* nautobot-ssot version:  4.6.1

### Proposed Functionality

Related: #883

The integration creates supporting objects on demand rather than requiring them to exist. From `diffsync_models.py`:

| Object | Mechanism |
|---|---|
| Location | `get_or_create_location_object` |
| Device Type | `get_or_create_device_type_object` |
| Manufacturer | created alongside Device Type |
| Platform | `get_or_create_platform_object` |
| Role | `get_or_create_device_role_object` |
| Status | `get_or_create_status_object` |
| Virtual Chassis | `get_or_create_virtual_chassis_object` |

That is the right default for bootstrapping an empty Nautobot. It is the wrong behaviour where another system is authoritative for those objects, because a typo or an unexpected value in IP Fabric silently mints a new Location, Role or Device Type instead of failing visibly. The created objects then look indistinguishable from curated ones.

Proposed: an opt-in strict mode that matches these objects rather than creating them, failing the affected record with a clear log line and continuing.

This must be additive and default to today's behaviour, or it breaks every existing user who relies on auto-creation. #883 is the natural home for the control, ideally per object type rather than a single global switch.

Two related notes:

- Setting `ipfabric_sync_ipf_dev_type_to_role` to `False` does not make the integration role-neutral. The adapter passes a null role, but device creation then applies `attrs.get("role") or DEFAULT_DEVICE_ROLE` and creates or assigns a role named `Network Device`. The flag stops the IP Fabric device type leaking into the role; every synced device still receives a written role. Worth documenting even if the behaviour stays.
- Virtual Chassis is absent from the integration's documented data mappings but is created on demand, so it is easy to miss when reasoning about what the sync writes.

### Use Case

As an operator whose Locations, Roles and Device Types are governed by a design or CMDB system, I want the IP Fabric sync to reference those objects and fail loudly when one is missing, rather than create a near-duplicate. Today a single unexpected site name in IP Fabric produces a new Location that then has to be found and merged by hand.

jdrew82:
# Closes: #1340

Closes #1346

Partially addresses #1343 — deliberately not a closing reference; see [What this does and does not do for #1343](#what-this-does-and-does-not-do-for-1343).

> [!IMPORTANT]
> Stacked on #1357 — this PR targets `ipfabric-bulk-sync-optimization`, not `develop`, and is a draft
> until that merges. It rides along one commit from that stack (`bump meraki to ^3`) which drops
> itself on rebase. Review the three commits above it.

### Interface addresses no longer fall back to `/32` (#1340)

An address has no mask of its own in IP Fabric — the mask comes from the managed address table, and
that table does not cover every address. The sync previously wrote `255.255.255.255` whenever it
found nothing there, which parents the address to the wrong Prefix and leaves nothing to distinguish
it from an address genuinely configured as a host route.

Such an address is now reported as absent by **both** adapters, so the mask Nautobot already holds is
left alone and an address Nautobot does not hold is not created. Reporting it on one side only would
leave it diffed on every run without ever being applied — or, worse, read as absent from the source
and get deleted. The job logs how many Interfaces this applied to, with the addresses themselves at
debug level.

Two related fixes in the same area:

- An `Interface.update()` where the address changed but its mask did not fell back to `/32`, taking
  an existing `/24` address to a host mask. It now keeps the mask already recorded.
- An unparseable or non-IPv4 subnet ended the whole job with a `ValueError` *while it was still
  reading addresses*, losing every address that was fine. Such a row is now logged and passed over.

A NAT management address is deliberately unaffected: it belongs to no interface, so no subnet is
reported for it and a host mask is the whole of it — the value rather than a fallback.

### Strict Objects (#1346)

A new job field naming the object types whose reported value is **checked rather than taken on
trust**. It layers on top of the existing sync scope: the scope decides which types a run covers,
strictness asks, of the types it does cover, whether IP Fabric's word can be trusted. Nothing here
brings a type into scope, and selecting a type that is out of scope is reported rather than left
looking like the reason nothing was written.

Left unselected, the sync fills any gap itself — a site name becomes a Location, a model becomes a
Device Type. That is what makes bootstrapping an empty Nautobot work, and it stays the default.
Where another system owns the data it means a typo mints a near duplicate indistinguishable from a
curated record.

Selectable types, and what a failed check costs:

| Selected | When the reported value does not check out |
| -------- | ------------------------------------------ |
| Locations | The site is reported and the Devices at it are skipped. |
| Manufacturers | The vendor is reported, and no Device Type is filed under it. |
| Device Types | The model is reported and the Device is skipped, since Nautobot requires one. |
| Roles | The role is reported and the Device is skipped, since Nautobot requires one. |
| Platforms | The platform is reported and the Device is synced without one, as Platform is optional. |
| Statuses | The status is reported and the record that needed it is skipped. |
| Virtual Chassis | The stack is reported and its membership is left unrecorded. |
| Interfaces | No `pseudo_mgmt` placeholder is invented for a management address reported against no Interface, so that address goes unsynced. |
| IP Addresses | The address is reported and left as Nautobot holds it. |

Only **IP Addresses** is selected by default, so no existing sync changes behaviour on upgrade.
Whether a *name* may be trusted depends on who owns the object, which only the operator knows; a
mask the source never reported is not data whoever owns IPAM. Each default can be set for the whole
instance with an `ipfabric_strict_<type>` setting.

Also adds Virtual Chassis to the job's data mappings — stack members were already synced to one, but
nothing in the job said so.

### What this does and does not do for #1343

#1343 asks for two things. This PR does the first and nothing for the second, which is why it
references the issue rather than closing it.

**Done — the fabricated `pseudo_mgmt` Interface.** The issue suggests either marking it so consumers
can exclude it, or finding a way to hold the address that does not invent an Interface. Selecting
**Interfaces** under Strict Objects takes the second route: the placeholder is not invented at all,
and the management address goes with it, since it had no Interface of its own to sit on. Two
caveats a reviewer should weigh:

- It is **opt-in**, so by default the placeholder is still fabricated. That is deliberate — turning
  it off by default would drop a management address every existing install currently syncs — but it
  means the false statement about the device is now *controllable* rather than *gone*.
- A Device reached only through NAT ends up with **no primary IP**, which needs the same exemption
  the DHCP case does. Not addressed here.

**Not done — virtual interface coverage.** Whether a current IP Fabric release returns SVIs,
loopbacks, tunnels and subinterfaces in the interface inventory is still unverified, so whether
they are silently absent from Nautobot is still unknown. Nothing in this PR touches that.

### Notes for review

- **Both controls gate reading and writing identically.** `may_create()` and
  `carries_pseudo_management_interface()` are each defined once on the shared adapter base and read
  by both adapters. That symmetry is load-bearing: a condition that gates what one adapter loads
  without gating what the other writes makes excluded objects look permanently changed, and
  withdrawing an object from one side of the diff reads as absent from the source and deletes it.
  Regression tests assert the diff itself, not just the load.
- **Strictness stops new placeholder Interfaces rather than removing existing ones**, for the same
  reason. The job names every one it finds so they can be dealt with deliberately.
- **Safe Delete Mode's own statuses are outside this control** — they are the integration's
  vocabulary rather than anything IP Fabric reported, and refusing to create one would leave a record
  neither deleted nor marked, the one outcome Safe Delete Mode exists to prevent.
- **`import ipaddress` guard placement**: `reported_prefix_length()` returns `None` rather than
  raising, because one bad row must not end a load.

## To Do

- [x] Added change log fragment(s) — `1340.fixed`, `1346.added`, `1346.changed`
- [x] Unit, Integration Tests — 509 ipfabric tests passing; ruff, pylint and markdownlint clean
- [x] Documentation Updates — new **Strict Objects** and **Placeholder Interfaces** sections, plus a
      consolidated **Subnet Masks** section, in `docs/user/integrations/ipfabric.md`

## Remaining work

- **Prefixes are not in the strict registry.** `create_parent_prefix` still auto-creates Prefixes in
  the Global namespace, which is the one remaining gap of this class. Deferred to a later sprint by
  decision, not oversight.
- **Leftover `pseudo_mgmt` placeholders are reported, not removed.** Whether they should instead be
  safe-deleted is an open question.
- **The second half of #1343** — confirming which virtual interface classes IP Fabric actually
  returns — and the `primary_ip4` exemption for NAT-reached Devices both remain open under that
  issue.


nautobot_ssot/integrations/ipfabric/utilities/nbutils.py
```diff
@@ -557,6 +566,19 @@ def get_or_create_status_object(  # pylint: disable=too-many-arguments
     return None
 
 
+def get_virtual_chassis_object(name: str, logger=None) -> Optional[VirtualChassis]:
```
alhogan, line 570: Is `@job_scoped_cache` intentionally left off here, or missing? The other match-only lookups and `get_or_create_virtual_chassis_object` below all have it.

alhogan:
https://github.com/nautobot/nautobot-app-ssot/blob/d8685d74325a8b8aae8fc4d44e951cff252ab41a/nautobot_ssot/integrations/ipfabric/diffsync/diffsync_models.py#L371

I think this condition wants to be `adapter.may_create("roles")` rather than `adapter.scope.roles`, to match what the comment above it describes.

With Roles selected under Strict Objects, `get_device_role_object` falls back to a name match, so a Role owned by another system (with no `ipfabric_type` cf) is found by name. That makes `cf.get("ipfabric_type") != role_name` true, and since the Role is still in scope, role is updated with the CF marked, which I believe is what is intended to be avoided.

If so, then it should probably be:
```python
if adapter.may_create("roles") and device_role_object.cf.get("ipfabric_type") != role_name:
```

nautobot_ssot/integrations/ipfabric/utilities/nbutils.py
jdrew82: Missing, not intentional — good catch. Added in 45b81b3f.

I checked the rest of the module while I was there: every other `get_*_object` carries it, and the only uncached helpers left are `get_tagged_interface` and `create_interface`, which each document why (one so callers that mutate an Interface see their own writes, the other because it is unhashable). This one had no such reason.

jdrew82:
Thanks @alhogan — both addressed in 45b81b3f.

**`may_create("roles")`**: you were right, and it is a real bug rather than a style point. I wrote the test before touching the code to be sure, and it failed exactly as you described:

```
AssertionError: 'Core Router' is not None
```

Under Strict Objects with Roles selected, `get_device_role_object` falls back to the name match, finds a Role another system owns, sees no `ipfabric_type`, and stamps one — claiming a record strictness exists to leave alone. The comment above the line already described the intended behaviour; the condition did not implement it. Now `adapter.may_create("roles")`, which is the scope and the strictness together, and `test_a_role_another_system_owns_is_not_stamped_when_strict` pins it.

**`@job_scoped_cache`**: missing rather than intentional, replied on the thread.

One thing I tried and backed out: a companion test asserting the non-strict path still stamps a name-matched Role. It does not — without strictness, `get_or_create_device_role_object` matches on the custom field, misses, and attempts to create a second Role of that name. That is separate pre-existing behaviour rather than anything this PR changes, and the default path is already covered by `test_nothing_strict_creates_what_is_missing`, so I dropped the test rather than leave a misleading one. Happy to open an issue for it if you think the duplicate-name attempt is worth chasing.

Kept as its own commit rather than folded into the originals, so the delta since your review is readable. 539 tests pass, pylint clean.

alhogan (approved):

Issue #1362, opened 2026-09-02
IP Fabric: remaining per-object query and memory waste after the bulk sync work


jdrew82:
### Environment

* Nautobot version: 3.1 and 3.2
* nautobot-ssot version: develop

### Proposed Functionality

#882 brought the per-object cost of an IP Fabric sync down and added an opt-in Bulk Write Mode. A review of that change found further per-object waste that was left alone at the time, either because the fix would change behaviour or because it sat outside the change. Costs below are for the estate #882 targets: 1k-5k devices, 100k+ Interfaces and IP Addresses.

Listed roughly by value.

#### 1. An existing IP Address is written on every run even when nothing about it changed

`create_ip` calls `tag_object` for an address Nautobot already holds, which runs a full `validated_save()` purely to re-stamp `system_of_record` and the last-synchronised date. `IPAddress.save()` calls `clean()` itself, and `clean()` issues three queries of its own, so this is roughly 11 queries per address per run — about 1.1M on a re-sync of 100k addresses that changes nothing.

The same applies to the Interfaces and other objects `tag_object` re-stamps. Worth deciding whether the date stamp needs rewriting when no field changed; if it does, `queryset.update()` on the custom-field data would avoid the validation and the change log entry.

#### 2. A newly created Cable is written twice

`utilities/cables.py:create_cable` does `cable.validated_save()` and then `tag_object(cable)`, which saves again to carry the tag and stamp. About 11 extra queries per Cable created. The same fix already applied to Interfaces and IP Addresses in #882 works here: `stamp_synced` before the one save, then `cable.tags.add(synced_tag_for(cable))`. Left out of #1357 because `cables.py` is otherwise untouched by it.

#### 3. `resolve_ip` costs two SELECTs per address where one would do

```python
existing = IPAddress.objects.filter(address=address).first()
...
existing = IPAddress.objects.filter(host=address.split("/")[0], parent__namespace=get_global_namespace()).first()
```

`filter(address=...)` is Nautobot's shorthand for host plus mask length, so both are keyed on the same host; the second widens the mask and narrows the namespace. On a first sync both miss and the pair runs for every address, in both write paths — about 100k wasted queries.

[...]

PR #1374, opened 2026-09-10
Sync every IP Address IP Fabric reports for an Interface

### LINKED ISSUE #1344: Sync secondary, IPv6 and FHRP interface addresses from IP Fabric
### Environment
* Nautobot version:  3.2.1
* nautobot-ssot version:  4.6.1

### Proposed Functionality

Related: #890 #891

Today the integration syncs at most one IPv4 address per interface. Two things enforce that:

```python
ip_filter = {"type": ["eq", "primary"]}
for ip_address in self.client.technology.addressing.managed_ip_ipv4.all(columns=ip_columns, filters=ip_filter):
```

and the Interface DiffSync model carrying singular fields:

```
_attributes = (..., "ip_address", "subnet_mask", "ip_is_primary", ...)
```

So secondary addresses, FHRP virtual addresses and IPv6 are all absent from Nautobot. There is a note in the existing code acknowledging the limitation ("Interfaces only support synchronizing 1 IP Address at the moment").

Proposed:

1. Drop the `type eq primary` filter and load all managed IPv4 addresses per interface, retaining which one is primary.
2. Add the equivalent IPv6 source so v6 addresses are synced alongside v4.
3. Include FHRP virtual addresses, which are configured on the device and render into configuration.
4. Change the Interface model from singular `ip_address` / `subnet_mask` to a collection, so multiple addresses per interface diff correctly.

Point 4 is the substantial part: it changes the shape of the Interface DiffSync model and its create, update and delete paths, so it is not a purely additive change. Worth sequencing against #882, which touches the same load path.

### Use Case

As an operator using Golden Config, I want the addresses my templates render to come from Nautobot. Any stanza that renders a secondary address, an HSRP or VRRP virtual address, or an IPv6 address cannot be driven from Nautobot data today, so those parts of the configuration cannot be generated or compared. Dual-stack estates are effectively unmodelled.

### LINKED ISSUE #1353: IP Fabric: IP Addresses and Cables are queued for deletion but never deleted

### LINKED ISSUE #1360: IP Fabric: an IPv6-only Interface makes the Nautobot adapter load raise ValueError
### Environment

* Nautobot version: 3.1 and 3.2
* nautobot-ssot version: develop

### Observed Behavior

An Interface whose only IP Addresses are IPv6 makes the Nautobot adapter's load raise, ending the job:

```
ValueError: Parameter must be an integer between 0 and 32.
```

Reproduced directly against a Device with one Interface carrying `2001:db8::1/64`:

```python
adapter.load_interfaces(device, mock.MagicMock())
# ValueError: Parameter must be an integer between 0 and 32.
```

### Cause

`NautobotDiffSync.load_interfaces` reports the mask of the first address on the Interface:

```python
ip_address_obj = ip_addresses[0]
subnet_mask = cidr_to_netmask(ip_address_obj.mask_length)
```

`netutils.ip.cidr_to_netmask` accepts 0-32 and raises above that (netutils has a separate `cidr_to_netmaskv6`). `IPAddress` orders by `("ip_version", "host", "mask_length")`, so IPv4 sorts first and a dual-stack Interface is safe. An Interface with only IPv6 addresses puts an IPv6 address at `[0]` and the conversion raises. Nothing catches it, so the whole load fails rather than the one Interface being skipped.

Reachable on any Interface whose addresses came from another source, since this integration reads `managed_ip_ipv4` and so never creates IPv6 itself.

### Expected Behavior

The load completes. What it should report for such an Interface is the open question, and it interacts with #1344 (syncing IPv6 addresses from IP Fabric):

* Reporting `None` for `ip_address` and `subnet_mask` matches what the IP Fabric side reports for an Interface it has no IPv4 for, so the two agree and the diff settles — but it means Nautobot's IPv6 address is invisible to the sync, and a future `ip_address`-clearing update path could remove it.
* Preferring the first IPv4 address, and reporting `None` when there is none, keeps a dual-stack Interface behaving as it does today.
* Reporting the IPv6 mask properly requires the source side to report IPv6 too, i.e. #1344.

Whichever is chosen, the conversion should not be able to raise out of a load.

jdrew82:
# Closes: #1344

Closes #1360

Closes #1353

Partially addresses #1362 — items 5 and 6 only; see [What this does for #1362](#what-this-does-for-1362).

> [!IMPORTANT]
> Stacked on #1369 — this PR targets `ipfabric-unresolvable-subnet-mask`, not `develop`, and is a
> draft until that and #1357 beneath it merge. Review the two commits above that branch's tip.

### Every address IP Fabric reports for an Interface is synced (#1344)

An Interface carried one address. IP Fabric reports several, and the ones that were dropped are the
ones an operator most wants rendered from Nautobot: a secondary address, an IPv6 address alongside
IPv4, and an HSRP or VRRP virtual address.

Three tables are now read — `technology.addressing.managed_ip_ipv4`, `managed_ip_ipv6` and
`technology.fhrp.group_members` — and **an address is its own DiffSync model**, a child of the
Interface, rather than three fields on it. That is the substance of the change: as a model each
address diffs on its own, so an Interface gaining a third address reports that one address rather
than its whole set, one whose mask is corrected reports the mask, and one IP Fabric stops reporting
is removed without touching the rest. The alternative — a list attribute on the Interface — would
report any change to any address as the whole Interface changing, and name none of them in the log.

An address is identified by **host within its Interface**, not by address. Nautobot makes an address
unique within its parent Prefix, and the mask is the attribute IP Fabric can change for an address
it keeps reporting, so keying on the mask as well would report a corrected mask as one address
replacing another — a delete and a create for what is one row.

What a prefix length buys, beyond IPv6 being expressible at all: `create_ip` takes a length rather
than a dotted netmask throughout, so one call serves either version.

### `primary_ip4` and `primary_ip6` are marked, not resolved (#1344)

A primary address is now an attribute of an address the sync already read from the Interface
carrying it — `is_primary` — rather than a separate resolution of its own. The prefix length is
whatever that Interface reported; nothing about being primary resolves a mask.

Consequences worth reviewing:

- **A dual-stack Device carries a primary of each version.** IP Fabric reports `loginIpv4` and
  `loginIpv6` as separate columns, and both are read; a release exposing only the older single
  `loginIp` marks that one. Previously whichever was read first won and the other was never set.
- **An address that stops being logged in on is unmarked**, so a primary that moves to another
  address, or a Device IP Fabric reports no login address for at all, no longer leaves Nautobot
  pointing at the old one.
- **The demotion is keyed on identity, not order.** Whichever address becomes primary instead is a
  different model and DiffSync does not order the two, so clearing the field outright could undo a
  promotion already applied. `clear_primary` clears only where the Device still points at *this*
  address: applied after the promotion it is a no-op, applied before it the promotion writes the
  field again. Both orders are tested.
- `is_primary` is dispatched on `"is_primary" in attrs`, not on its truth — an address that stops
  being primary reports `False`, which is a value to act on rather than an attribute left unset.

**Sync Primary IP** governs the marking alone; deselecting it still syncs the addresses.

### An IPv6-only Interface no longer ends the Nautobot adapter's load (#1360)

`ValueError`, which took the whole job with it. The adapter reported a mask by converting the prefix
length to a dotted netmask, which is undefined above 32. Addresses sort by IP version, so a
dual-stack Interface put IPv4 first and was unaffected while an IPv6-only one raised — which is why
this had gone unnoticed.

Fixed structurally rather than guarded: a prefix length is what the sync records on both sides, so
there is no conversion left to raise. The same correction reaches `create_parent_prefix`, which
built its covering prefix with a hardcoded `/32`.

### Every queued deletion is drained (#1353)

With Safe Delete Mode disabled, each object is queued for removal under a grouping taken from its
class name, and `sync_complete` drained **four** of them. IP Addresses and Cables were dropped from
the sync's own records and reported as deleted while remaining in the database.

Every queued grouping is now drained, children before whatever they hang off, and a grouping the
order does not name is removed last and **reported** rather than accumulating unread — so the next
model to reach this queue fails loudly instead of silently.

### What this does for #1362

Two of that issue's six items fall out of this work, so it is referenced rather than closed:

- **Item 6 — a newly created Interface is asked what addresses it has.** Gone. `Interface.create`
  no longer calls `ip_addresses.set([])`; an address is its own model, so the Interface has no
  address relation to reconcile. About 100k queries on a first sync of the estate #882 targets.
- **Item 5 — the retained Interface graph.** Partially. `get_device_interfaces_by_name` no longer
  prefetches `ip_addresses__interfaces`, so a retained Interface no longer drags its addresses and
  their sibling Interfaces with it. Two joins nothing reads, since removing an address now goes
  through `InterfaceAddress`, which resolves its Interface afresh.

  The instances themselves are still appended rather than `(type, pk)`, so that half stands. Worth
  correcting the issue on one point, though: `safe_delete` appends **only** when Safe Delete Mode is
  off, so the list does not accumulate in the default configuration as the issue states.

Items 1-4 are untouched.

## Notes for review

- **`find_interface` uses the uncached lookup deliberately.** The by-Device lookup is cached, so it
  would see neither an Interface this run created nor an address written onto one it did. Using it
  here made a `pseudo_mgmt` address report `MISSING in NautobotDiffSync` on every run.
- **The subnet ladder gained a rung, and it is capacity-gated.** A record with no subnet of its own
  may take the subnet of a sibling address on its Interface that contains it — that is how an FHRP
  virtual address resolves. Only records marked `_inherits_subnet` may take it; applying it to every
  record would silently weaken #1340's guarantee that an address with no reported subnet is withheld
  rather than guessed at. Such records are sorted last so the subnets they read are already resolved.
- **Removal only where no other Interface holds the address.** One Nautobot IP Address can be
  assigned to several Interfaces, and deleting it for one would take it from all. An address on
  another Interface is unassigned from this one instead; otherwise it goes through Safe Delete Mode
  as any other object does.
- **`assign_primary` uses `validated_save()`**, matching the house convention (417 uses to 15 bare
  saves, seven other integrations doing exactly this) — with one consequence: it validates the whole
  Device, so a Device this sync deliberately writes despite `Device.clean()` (a Platform naming a
  different Manufacturer to its DeviceType) is refused here for a reason unrelated to the address.
  It is logged, and the Device keeps whatever primary it had. Documented on the method.
- **The FHRP virtual-address column is looked for under `vip`, `virtualIp` and `virtualIP`.** IP
  Fabric reports a table's columns from the appliance rather than declaring them offline, so this is
  unverified against a live release; a group member carrying none of the three is reported in the job
  log rather than silently syncing nothing.

## To Do

- [x] Added change log fragment(s) — `1344.added`, `1360.fixed`, `1353.fixed`
- [x] Unit, Integration Tests — 585 ipfabric tests passing, 100% of the 263 new statements covered;
      ruff, pylint (9.97) and markdownlint clean
- [x] Documentation Updates — new **Interface Addresses** and **FHRP virtual addresses** sections,
      and a reworked **Subnet Masks** section, in `docs/user/integrations/ipfabric.md`

## Remaining work

- **A demotion cannot express "no longer primary" without an address to hang it on.** Because
  `is_primary` lives on the address, a Device whose last login address disappears from IP Fabric
  entirely is handled by the delete path rather than by an attribute change. Moving the fact to the
  Device as `primary_ip4_host`/`primary_ip6_host` would express it directly; deferred, as it changes
  the Device's identity surface.
- **The dual-NAT placeholder carries only the first unreported address.** A Device reached through
  more than one NAT address still gets one `pseudo_mgmt` address.
- **A NAT-reached Device still has no primary IP** under strict **Interfaces**, and the FHRP column
  name still wants confirming against a live appliance. Both carried over from #1369.
- **#1362 items 1-4** remain open, as does the `(type, pk)` half of item 5.


alhogan (approved):
I like the changes, especially the cidr & subnet_mask notation conversion to a consistent mask_length. Good way to simplify things. Some of the doc strings are a little long winded, but make sense.

Issue #1363, opened 2026-09-02
IP Fabric sync assumes the Global Namespace for every address and Prefix


jdrew82:
### Environment

* Nautobot version: 3.1 and 3.2
* nautobot-ssot version: develop

### Observed Behavior

Every IP address the IP Fabric sync resolves is assumed to be in the Global Namespace. A deployment that separates addressing into several Namespaces is not covered, and the failure is quiet rather than loud.

`utilities/nbutils.py`:

```python
@job_scoped_cache
def get_global_namespace() -> Namespace:
    """Return the Global Namespace, which every Prefix this integration creates belongs to."""
    return get_default_namespace()
```

Used by `create_parent_prefix` (every Prefix the sync creates goes in it) and by `resolve_ip` (an existing address is only matched in it). `_cleaned_ip` builds `IPAddress(address=..., status=...)` with no Namespace, so Nautobot's default applies there too.

Consequences in a multi-Namespace deployment:

* An address Nautobot holds in a non-default Namespace is not found, so the sync tries to create a second one in Global.
* A Prefix the sync creates always lands in Global, whatever Namespace the surrounding addressing uses.
* Two devices with the same address in different Namespaces — the case Namespaces exist for — collapse onto one row, since the match is on host within Global only.

### Expected Behavior

The Namespace an address belongs to is decided per address, or at least configurable for the run, rather than assumed.

There is a design question first: IP Fabric's addressing table has no Namespace concept, so something has to supply the mapping. Candidates are a job option naming the Namespace for the run, a mapping from IP Fabric site or device to Namespace, or per-VRF derivation if IP Fabric's VRF data can carry it.

Related: #1356 fixed one address reported in two subnets by choosing one mask; a Namespace-aware sync would be able to hold both instead. #1347 (reporting addresses outside every Prefix) touches the same resolution path.

Noted as a constraint on #1357 rather than fixed there.

Issue #888, opened 2025-07-09
Import Device VRF from IP Fabric


jjeff07:
### Environment
* Nautobot version:  latest
* nautobot-ssot version:  latest

### Proposed Functionality

Depends #883 #887 

<img width="1072" height="597" alt="Image" src="https://github.com/user-attachments/assets/4e88e7b4-f572-4468-95ee-d9fb92ab6a1b" />

1. Collect Device VRFs from IP Fabric's `Technology > Routing > VRF > Detail` table
2. During Nautobot import of devices if the device's `sn` is found in the VRF Detail table then include all VRFs configured on the device in the import if enabled.

### Use Case

IP Fabric is VRF aware and this data can help customers understand Interfaces and Devices along routing information.  This is required for importing VRF on Interfaces.

PR #1378, opened 2026-09-16
Sync VRFs, Route Targets and their Device and Interface assignments from IP Fabric

### LINKED ISSUE #887: Import VRFs from IP Fabric
### Environment
* Nautobot version:  latest

### Proposed Functionality

Depends on #886

Image

1. Use IP Fabric's `Technology > Routing > VRF > Summary` table to get unique VRF names in the Network.
2. Use `Technology > MPLS > L3 VPN > VRF Route Targets` table to collect:
    1. RD
    2. Import RT
    3. Export RT
3. Ensure that every unique VRF has the same RD and RT configurations.  The `VRF Route Targets` lists VRFs per device and misconfigurations could lead to conflicting information.
4. Import into Nautobot's default Global Namespace
    1. If VRF has RD and RT without conflicts, then import with this information.
    2. If VRF has conflicting RD and RT import without this information and possibly note this in a custom field?
    3. If VRF does not have RD or RT import as is.

Image

### Use Case

IP Fabric is VRF aware and this data can help customers understand Interfaces and Devices along routing information.

jdrew82:
# Closes: #887, #886, #888, #889

## What's Changed

Adds VRF support to the IP Fabric integration, as four new object types. All are opt-in and off by default, so no existing sync changes on upgrade.

| Job option | Object type | Issue |
| --- | --- | --- |
| Sync VRFs | `ipam.VRF`, in the Global Namespace | #887 |
| Sync Route Targets | `ipam.RouteTarget`, and each VRF's import/export lists | #886 |
| Sync Device VRFs | `ipam.VRFDeviceAssignment` | #888 |
| Sync Interface VRFs | `Interface.vrf` | #889 |

Read from `tables/vrf/detail`, `tables/mpls/l3-vpn/vrf-targets` and `tables/vrf/interfaces`. The VRF summary table is not read: it names the same VRFs but carries no route distinguisher, so the detail table has to be read regardless and answers both questions.

### Reconciling what each device reports

IP Fabric reports a route distinguisher per device, and route targets per device and per address family, while Nautobot holds one of each per VRF. Both are reconciled before anything is written:

- Where every device carrying a VRF reports the same value, that value is the VRF's.
- Where they disagree the network is misconfigured. Nothing is recorded, and the disagreement is written to an **IPFabric VRF Conflict** custom field — a VRF silently carrying one device's route distinguisher would be worse than one carrying none, since nothing would say which device it came from.
- The two are reconciled independently, so a VRF whose devices agree on its route distinguisher but not its route targets still records the route distinguisher.
- Address families are combined rather than reconciled against each other, since Nautobot holds one set of targets per VRF: a device importing one target for IPv4 and another for IPv6 imports both.

### Write ordering

Route Targets are written before the VRFs that name them, and deleted after. VRF device assignments come after both the VRF and the Device. Interface VRFs come last of all, because `Interface.clean()` refuses a VRF that is not assigned to the Interface's Device — and those assignments cannot exist until every Device is written, which is after its Interfaces are. That is why the Interface VRF is a model of its own rather than an attribute of the Interface, and why it requires **Sync Device VRFs**.

### Guards on what the sync may touch

- A VRF name the Global Namespace already holds twice is left alone entirely. Nautobot constrains a VRF to a unique route distinguisher within a Namespace but not to a unique name, and IP Fabric reports nothing that tells the two apart.
- A Route Target's name is unique across Nautobot, so one another process created is adopted and marked as synced rather than duplicated. That is what lets a first run converge against an existing estate.
- Only the Route Targets carrying this integration's Tag are loaded back. A Route Target has no Location, Device or Namespace to bound a load by, so loading all of them would delete every Route Target another system owns as soon as IP Fabric stopped reporting it.
- With a Site Filter, VRFs and Route Targets are created and updated but never deleted, since a filtered run sees only one site's. The Device and Interface assignments need no such protection — both hang off a Device, and Devices are narrowed on both sides at once.
- Only Devices and Interfaces the run covers are assigned to, on both sides.

### Bulk Write Mode

VRFs and device assignments are written in batches; Route Targets are not, because a network has a handful of them and a VRF's join rows need theirs to exist already. Every write path resolves a VRF, Device or Interface from the pending queue before the database, since in bulk mode any of them may be queued rather than written.

### Safe Delete Mode

A VRF is moved to `Deprecated` and tagged, as other object types are. A Route Target has no status field, so it can only be tagged. A VRF device assignment has neither, so Safe Delete Mode leaves it in place — documented, since it means a run reports a deletion it did not make. Taking an Interface out of a VRF is an attribute change rather than a deletion, so Safe Delete Mode does not apply to it.

## To Do

- [x] Added change log fragment(s) — `changes/886.added`, `887.added`, `888.added`, `889.added`
- [x] Unit, Integration Tests — 105 new tests in `nautobot_ssot/tests/ipfabric/test_vrfs.py`; the IP Fabric suite is 575 tests and passes. Includes convergence tests asserting a second sync of unchanged data reports nothing, in both per-object and bulk write modes.
- [x] Documentation Updates — `docs/user/integrations/ipfabric.md`
- [x] Outline Remaining Work, Constraints from Design — below

## Remaining work and constraints

- **Namespace.** Every VRF is created in the Global Namespace, matching where this integration already puts Prefixes and IP Addresses. `VRF.namespace` is part of Nautobot's uniqueness constraint, so this work had to settle it; it takes the same answer as the rest of the integration, through the single `get_global_namespace()` seam, which leaves #1363 one place to change rather than several.
- **Three toggles** are needed to get an Interface into a VRF (Sync VRFs, Sync Device VRFs, Sync Interface VRFs). Each is independent because there are real deployments where another system owns that layer, but it is worth a look during review as a usability question.
- **Verified against a live IP Fabric instance**, not only the fixtures: VRFs appear on both Devices and Interfaces as expected.


bile0026 (approved):
LGTM!

